PR #1, opened 2016-10-15
fix coverage job

dirk-thomas:
The coverage jobs hasn't run any coverage tests since 3.5 months. The last successful build was http://ci.ros2.org/view/nightly/job/nightly_linux_coverage/9/

This patch addresses several problems:
* check return value of commands and fail if non-zero (since they were actually failing)
* pass the --coverage argument to the right script (to fix the wrong invocation)
* fix filtering only .gcov files which point to files outside the workspace (fix logic as well as syntax)

See http://ci.ros2.org/view/nightly/job/nightly_linux_coverage/161/cobertura/

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -196,7 +196,7 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: ament build')
     # Now run ament build
-    job.run([
+    ret_build = job.run([
         '"%s"' % job.python, '-u', '"%s"' % ament_py, 'build', '--build-tests',
         '--build-space', '"%s"' % args.buildspace,
         '--install-space', '"%s"' % args.installspace,
@@ -211,7 +211,10 @@ def build_and_test(args, job):
             gcov_flags + ' " -DCMAKE_C_FLAGS="${CMAKE_C_FLAGS} ' +
             gcov_flags + '" -- ']
             if coverage else []), shell=True)
+    info("ament.py build returned: '{0}'".format(ret_build))
     print('# END SUBSECTION')
+    if ret_build:
+        return ret_build
 
     print('# BEGIN SUBSECTION: ament test')
     # Run tests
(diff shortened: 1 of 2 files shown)


ros2_batch_job/__main__.py
```diff
@@ -155,6 +155,7 @@ def process_coverage(args, job):
         [job.python, '-u', ament_py, 'list_packages', args.sourcespace])
     for line in output.decode().splitlines():
         package_name, package_path = line.split(' ', 1)
+        print(package_name)
```
wjwwood: Is this still necessary?

ros2_batch_job/__main__.py
```diff
@@ -163,17 +164,12 @@ def process_coverage(args, job):
...
-        subprocess.run(['gcov', '--preserve-paths'] + gcda_files, check=True)
-        # Remove coverage files that did not originate in the workspace
-        for cov_file in (f for f in os.listdir(package_build_path)
-                         if f.endswith('.gcov')):
-            cov_path = cov_file.replace('#', '/')
-            # TODO Better path filter here
-            if not cov_path.beginswith(os.path.abspath(args.workspace)):
-                os.remove(os.path.join(root, cov_file))
+        cmd = ['gcov', '--preserve-paths', '--relative-only', '--source-prefix', os.path.abspath('.')] + gcda_files
+        print(cmd)
```
wjwwood: Same here, might be useful, but also might be accidentally left over.

ros2_batch_job/__main__.py
```diff
@@ -182,9 +178,14 @@ def process_coverage(args, job):
...
-        subprocess.run(
-            ['gcovr', '-e', '/usr', '--xml', '--output=' + outfile, '-g'],
-            check=True)
+        cmd = [
+            'gcovr',
+            '--object-directory=' + package_build_path,
+            '-k',
+            '--xml', '--output=' + outfile,
+            '-g']
+        print(cmd)
```
wjwwood: Same.

dirk-thomas:
I would propose to keep the print statements since it helps when looking at the console output. Otherwise it is completely opaque what happens.

wjwwood:
That's fine by me, so long as it is intentional.

PR #3, opened 2016-10-18
Install flake8 and flake8-import-order

dhood:
connects to ament/ament_lint#53
needed by ament/ament_lint#63

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -35,7 +35,7 @@ RUN if test ${PLATFORM} = x86; then curl --silent https://packagecloud.io/gpg.ke
 RUN apt-get update && apt-get install -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
-RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-mock python3-nose python3-pep8 uncrustify
+RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-flake8 python3-mock python3-nose python3-pep8 uncrustify
 
 # Install and self update pip/setuptools to the latest version.
 RUN apt-get update && apt-get install -y python3-pip
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -383,6 +383,8 @@ def run(args, build_function, blacklisted_package_names=None):
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
         job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U'] + pip_dependencies, shell=True)
+        # Show what pip has
+        job.run(['"%s"' % job.python, '-m', 'pip', 'freeze'], shell=True)
         print('# END SUBSECTION')
 
         # Skip git operations on arm because git doesn't work in qemu. Assume
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -45,15 +45,15 @@ sys.stdout = UnbufferedIO(sys.stdout)
 sys.stderr = UnbufferedIO(sys.stderr)
 
 pip_dependencies = [
+    'EmPy',
+    'coverage',
+    'flake8',
+    'flake8-import-order',
+    'mock',
     'nose',
     'pep8',
     'pydocstyle',
     'pyflakes',
-    'flake8',
-    'flake8-import-order',
-    'mock',
-    'coverage',
-    'EmPy',
     'vcstool',
 ]
 
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -50,6 +50,7 @@ pip_dependencies = [
     'pydocstyle',
     'pyflakes',
     'flake8',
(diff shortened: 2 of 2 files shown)


ros2_batch_job/__main__.py
```diff
@@ -48,7 +48,8 @@
     'pep8',
     'pydocstyle',
     'pyflakes',
-    'flake8',
+    'flake8==2.6.2',
```
dirk-thomas: What is the reason to pin the version to 2.x instead of using the latest release - currently 3.0.4?

dhood: version 3 was a complete re-write and doesn't have a stable API yet http://flake8.pycqa.org/en/latest/user/python-api.html
There's a shim to the 2.x API but I wasn't able to see how to overload the reporter as is needed [here](https://github.com/ament/ament_lint/pull/63/files#diff-6f9f51ecc5d47adabeb8b56a439b27deR207)
I can watch the releases to see when an API gets added

dirk-thomas: I would propose to still use the latest release. If they really change the API we will notice immediately and can update our usage.

dhood: it's not straightforward to use the new release in https://github.com/ament/ament_lint/pull/63 (there is no 3.x API, and the shim to the 2.x API doesn't expose what's needed).

I appreciate pinning to an old version isn't _ideal_ (especially since one of the selling points of flake8 is its active development). if using 2.x is a _blocker_, I'll put https://github.com/ament/ament_lint/pull/63 on the backburner until there is an API in 3.x since it might take me a bit to get 3.x to work

dirk-thomas: I don't think it is that difficult: see dhood/ament_lint#1

linux_docker_resources/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN if test ${PLATFORM} = x86; then curl --silent https://packagecloud.io/gpg.ke
 RUN apt-get update && apt-get install -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
-RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-mock python3-nose python3-pep8 uncrustify
+RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git python3-coverage python3-mock python3-nose uncrustify
```
dirk-thomas: I don't think the existing linter dependencies should be removed. Every package should be able to choose what linters it would like to use and I don't see a reason to delete the linter packages we already have.

dhood: they are being installed through pip in [this script](https://github.com/ros2/ci/blob/140bd1e0e150800754b9f2483e53a9f7a52921ef/ros2_batch_job/__main__.py#L49). installing them here is what is causing version conflicts I believe (see http://ci.ros2.org/view/All/job/ci_linux/1802/testReport/junit/(root)/projectroot/flake8/).

wjwwood: I think the idea was to install what ever you could from apt, but everything must be installed in the Python script so that OS X and Windows get the Python dependencies.

I'm agnostic as to whether or not we use the apt version of the packages, though I suppose that might be more like a "typical" user's setup.

dirk-thomas: The reason why we recommend **not** to use PIP packages in ROS is that they overlay the Debian packages and are not being updated by the "native" package manager run on the system (aka `apt`).

dhood: I can see the motivation for using apt instead of pip. I have just had apt suggest installing `flake8` despite already having flake8 installed through pip.
If we want to stick with packages installed through apt, in the case of flake8 we'll need to either pin the [batch job script to use v 2.x](https://github.com/ros2/ci/pull/3/commits/9dfd52c9b4f4883a35ef9131db1b99b2362942a0#diff-a7a7037d3453364950fd80f5f91ed7eeR48) again, or support 2.x and 3.x simultaneously. (apt will install 2.5.4 while pip will install 3.0.4)

dhood:
Ready for review again please. to summarise the conversation:

We were discussing installing flake8 through pip or apt. [apt installs flake8 v2 on xenial](https://launchpad.net/ubuntu/+source/python-flake8) and pip installs v3. Because of this, this PR relies on flake8 being installed only through pip.

the import order flake8 plugin is not available through apt.

dhood:
since https://github.com/ament/ament_lint/pull/63 can work with flake8 v2 or v3 now (see https://github.com/ament/ament_lint/pull/63#issuecomment-272061385), I will add flake8 installation through apt back into this PR.

I haven't done it yet so that it's clear what commits are included in this PR when running CI jobs for https://github.com/ament/ament_lint/pull/63.

ros2_batch_job/__main__.py
```diff
@@ -45,14 +45,15 @@
 sys.stderr = UnbufferedIO(sys.stderr)
 
 pip_dependencies = [
+    'EmPy',
```
mikaelarguedas: should dependencies be listed in alphabetical order regardless of the case? (it seems that's how flake8 import order interprets alphabetical order according to the fixes in the attached PRs e.g. https://github.com/ros2/rosidl_dds/pull/40/files)

dhood: yep you're right that that's how the 'google-style' for flake8-import-order does it. I think it might be more common to sort considering case, so I'll leave it like this unless there are objections

PR #4, opened 2016-10-25
rewrite shebang line

dirk-thomas:
Fixes ros2/ros2#141. See http://ci.ros2.org/job/packaging_linux/437/

Even without using a venv embedding a system specific path to the Python interpreter is sub optimal. Rewriting the shebang line to the standard using `/usr/bin/env` seems to be the most flexible approach.

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -79,6 +79,23 @@ def build_and_package(args, job):
         ])
         print('# END SUBSECTION')
 
+    # Only on Linux and OSX Python scripts have a shebang line
+    if args.os in ['linux', 'osx']:
+        print('# BEGIN SUBSECTION: rewrite shebang lines')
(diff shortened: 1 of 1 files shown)


dhood:
As I understand it: [This rclpy tutorial](https://github.com/ros2/ros2/wiki/Python-Programming#run-the-examples) can be updated now to just refer to `talker_py` instead of 
```
python3 `which talker_py`
```

It probably should have also been corrected in the [linux binary install instructions](https://github.com/ros2/ros2/wiki/Linux-Install-Binary) but it has always referred to simply `talker_py`.

I'll update the page.

dirk-thomas:
Maybe we should keep it as-is for a while since that is still necessary for the latest release which users might be using.

dhood:
Excellent point - I'll update the linux binary install instructions to match the rclpy tutorial and we can address this later

PR #7, opened 2016-10-27
support arm builds again

dirk-thomas:
* ARM builds need to happen on Xenial since we don't support Trusty anymore.
* In order to not pass `--src-mounted --workspace-path` as part of the greedy `--ament-test-arg` they need to be prefixed with `--`.
* Since there is currently no OpenSplice Debian package for Xenial / ARM we disable the installation of it (this means enabling the `use_opensplice` flag for ARM jobs will fail the build).
* Since there are no ARM binaries of Connext we also change the default value of the parameter to `false`.

--

* http://ci.ros2.org/job/ci_linux-aarch64/18/
* http://ci.ros2.org/job/ci_linux-armhf/35/

### DIFF
--- a/job_templates/ci_job.xml.template
+++ b/job_templates/ci_job.xml.template
@@ -141,10 +141,10 @@ if [ "$CI_ENABLE_C_COVERAGE" = "true" ]; then
   export CI_ARGS="$CI_ARGS --coverage"
 fi
 if [ -n "${CI_AMENT_BUILD_ARGS+x}" ]; then
-  export CI_ARGS="$CI_ARGS --ament-build-args $CI_AMENT_BUILD_ARGS"
+  export CI_ARGS="$CI_ARGS --ament-build-args $CI_AMENT_BUILD_ARGS --"
 fi
 if [ -n "${CI_AMENT_TEST_ARGS+x}" ]; then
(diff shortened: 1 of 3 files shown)


job_templates/ci_job.xml.template
```diff
@@ -177,7 +177,7 @@ tmpd=`mktemp -d`
 mkdir $tmpd/src
 curl -sk https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos -o $tmpd/ros2.repos
 vcs import $tmpd/src --input $tmpd/ros2.repos
-export CI_ARGS="$CI_ARGS --src-mounted --workspace-path $tmpd"
+export CI_ARGS="$CI_ARGS -- --src-mounted --workspace-path $tmpd"
```
dhood: This logic could end up duplicated in another part of the code and then we might one day end up with `--ament-test-args -- -- --src-mounted --workspace-path` (which current makes ament invocations fail). I know it's not an issue for now, but I ticketed it at https://github.com/ament/ament_tools/issues/111

dirk-thomas:
Good catch. Added the trailing delimiter to the windows args. See e.g. https://github.com/ros2/ci/pull/7/files#diff-1cba2a8b70d9923b4b888bca2911fd9aR245

PR #8, opened 2016-10-28
install poco

dirk-thomas:
On the other platforms we could install it or just rely on the fallback of [poco_vendor](https://github.com/ros2/poco_vendor) to on demand fetch and build the package from source. The drawback of compiling from source is the additional time spent in every build (e.g. 77s on Windows). So if anyone would volunteer to install `poco` on all the nodes that would be great :wink:

http://ci.ros2.org/job/ci_linux/1840/

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -73,6 +73,9 @@ RUN apt-get update && apt-get install -y libboost-chrono-dev libboost-date-time-
 # Install OpenCV.
 RUN apt-get update && apt-get install -y libopencv-dev
 
+# Install build dependencies for class loader.
+RUN apt-get update && apt-get install -y libpoco-dev libpocofoundation9v5 libpocofoundation9v5-dbg
+
 # Install Git-LFS.
 RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install -y git-lfs; fi
 


dirk-thomas:
I also update [poco_vendor](https://github.com/ros2/poco_vendor/commit/e5b6017d97f1ba735b638b3925182d5eaa0e5afb) to use the latest released version as well as suppressed the compiler warnings (https://github.com/ros2/poco_vendor/commit/2d4f67f2eea29d9cb8f9901cb26ae7cdc220debe).

On OS X we should be able to install it via homebrew (?). On Windows I would argue to wait until there is / we made a Chocolatey package.

mikaelarguedas:
should be merged along https://github.com/ros2/ros2/pull/133 that updates the ros2.repos file accordingly

+1 for using poco_vendor until we look into creating our own chocolatey packages

dirk-thomas:
ros2/ros2#133 has to "wait" until it can build on all platform without warnings (including Windows :worried:).

dirk-thomas:
Ready for review (together with ros2/ros2#133 and ros/class_loader#41).

dirk-thomas:
Waiting for review.

dirk-thomas:
Also install poco in packaging Docker: cb64de9e6f16e214d49f9846af39ca4cc06c6087

PR #10, opened 2016-11-08
end --only-packages with --

dirk-thomas:
Follow up of ament/ament_tools#114.

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -68,7 +68,7 @@ def build_and_package(args, job):
             job.python, '-u', ament_py, 'build',
             '--build-space', '"%s"' % args.buildspace,
             '--install-space', '"%s"' % args.installspace,
-            '--only', 'ros1_bridge',
+            '--only-packages', 'ros1_bridge', '--',
             '"%s"' % args.sourcespace,
         ] + (['--isolated'] if args.isolated else []) +
             (


dhood:
I think it's better to move the sourcespace to the front of the argument list since the `--` delimiter will not work if any other optionals are added.

dhood:
since this has been merged we'll just have to keep this in mind in the future.

dirk-thomas:
I don't think a trailing `--` makes the command fail. Some of my local bash aliases always do that and work fine.

dhood:
If we want to package in isolation, [with how it's currently setup in line 53](https://github.com/ros2/ci/blob/4265d10e8dc79f9ba8068baa5fbf045a2611da08/ros2_batch_job/packaging.py#L53), I expect the job to fail:

```
$ ./src/ament/ament_tools/scripts/ament.py build --only-packages ament_tools -- ./src --isolated
usage: ament.py [-h]
                [build | build_pkg | list_packages | package_name | package_version | test | test_pkg | test_results | uninstall | uninstall_pkg]
                ...
ament.py: error: unrecognized arguments: --isolated
```

that's why I recommend moving the positional arguments to the front e.g. 
```
$ ./src/ament/ament_tools/scripts/ament.py build ./src --only-packages ament_tools --isolated
```

dirk-thomas:
Sounds good to me: #11

PR #11, opened 2016-11-08
move positional args to front

dirk-thomas:
Follow up of #10.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -218,10 +218,11 @@ def build_and_test(args, job):
     print('# BEGIN SUBSECTION: ament build')
     # Now run ament build
     ret_build = job.run([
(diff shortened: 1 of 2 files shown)


ros2_batch_job/__main__.py
```diff
@@ -218,10 +218,11 @@ def build_and_test(args, job):
     print('# BEGIN SUBSECTION: ament build')
     # Now run ament build
     ret_build = job.run([
-        '"%s"' % job.python, '-u', '"%s"' % ament_py, 'build', '--build-tests',
+        '"%s"' % job.python, '-u', '"%s"' % ament_py, 'build',
+        '"%s"' % args.sourcespace
```
dhood: copy-paste has caused a syntax error here, at L245 and at L49 of packaging.py :smile:

dirk-thomas: Fixed. Same below.

PR #13, opened 2016-12-02
add description for CI_ROS2_REPOS_URL

dirk-thomas:
Besides that the PR refactors some stuff to avoid duplication:
* rename the files ending in `.template` to `.em` (because that is the language they use, then https://atom.io/packages/language-empy will also highlight them nicely :wink:)
* clone the existing template (so git knows its the same file)
* remove the common parameters from two location and keep only those in the new common template

In order to deploy this #9 needs to be merged since otherwise the already deployed changes of that PR will be reverted.


job_templates/snippet/property_parameter-definition_common.xml.em
```diff
@@ -0,0 +1,37 @@
...
+        <hudson.model.StringParameterDefinition>
+          <name>CI_SCRIPTS_BRANCH</name>
+          <description>Branch of ros2/ci repository from which to get the ci scripts.</description>
+          <defaultValue>@ci_scripts_default_branch</defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_ROS2_REPOS_URL</name>
+          <description>Custom .repos file to use instead of the default (https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos).
+For example, copy the content of the .repos file to a GitHub Gist, modify it to your needs, and then pass the raw URL here.</description>
+          <defaultValue></defaultValue>
```
dirk-thomas: Done: 95d442da4b353194cfc383f4f6c01ebcb680298e

Issue #12, opened 2016-11-29
create a patch to disable all unused messages on CI


mikaelarguedas:
All message packages have been reenabled in https://github.com/ros2/common_interfaces/pull/23
To save time on the CI ran to test pull requests it would be great to disable all the unused messages to avoid building/testing them (c.f. https://github.com/ros2/ros2/issues/289#issuecomment-262416090)

tfoote:
Blacklisting was done in #14, but it's unintentially removing the packages from the packaging jobs too. @dirk-thomas is looking into this.

tfoote:
fixed in #14 and #23

PR #14, opened 2016-12-03
blacklist unused msg pkgs

mikaelarguedas:
implementation of the quick fix suggested by @dhood in https://github.com/ros2/ros2/issues/289#issuecomment-262418140 to reduce CI time 
This is a workaround until https://github.com/ros2/ci/issues/12 is implemented.
The messages will still be built by the packaging jobs given that they don't run this script but [packaging.py](https://github.com/ros2/ci/blob/master/ros2_batch_job/packaging.py)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -431,7 +431,14 @@ def run(args, build_function):
                 exit_on_error=False)
             print('# END SUBSECTION')
 
-        blacklisted_package_names = []
+        blacklisted_package_names = [
+            'actionlib_msgs',
+            'shape__msgs',
+            'std_srvs',
+            'stereo_msgs',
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -431,7 +431,14 @@ def run(args, build_function):
                 exit_on_error=False)
             print('# END SUBSECTION')
 
-        blacklisted_package_names = []
+        blacklisted_package_names = [
+            'actionlib_msgs',
+            'shape__msgs',
```
mikaelarguedas: yep, thanks for catching it https://github.com/ros2/ci/pull/14/commits/452b1935e84c246532a54ac75d7924fe1dd5d9ec

PR #16, opened 2016-12-08
Update submodule for new connext license

dhood:


dhood:
I think if we can configure our submodule in a way that the checkout does 
git submodule update --init --recursive **--remote** linux_docker_resources/rticonnextdds-license

at the [beginning of the linux docker jobs](http://ci.ros2.org/job/nightly_linux_debug/292/consoleFull) then this step won't be necessary anymore. I haven't seen how to do that myself

dirk-thomas:
Lets test this on CI first and get it merged so that the jobs work again. It can be improved later if desired.

dhood:
yep I've started [Build Status](http://ci.ros2.org/job/ci_linux/2004/)

dhood:
thanks for keeping an eye on it. I'll relaunch the failed jobs

dhood:
actually, that's proabbly not a good idea since people will want to use them for CI. I won't relaunch the nightlies!

PR #15, opened 2016-12-03
not add delimiters if already there

mikaelarguedas:
c.f. https://github.com/ros2/ci/pull/9#issuecomment-264612625

PR #18, opened 2016-12-11
End delimiter

mikaelarguedas:
fix syntax now that #15 has been merged. Will prevent breakage if we append any other argument to the command line

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -53,7 +53,7 @@ def build_and_package(args, job):
     if args.isolated:
         cmd.append('--isolated')
     if args.cmake_build_type:
-        cmd += ['--cmake-args', '-DCMAKE_BUILD_TYPE=' + args.cmake_build_type + ' -- ']
+        cmd += ['--cmake-args', '-DCMAKE_BUILD_TYPE=' + args.cmake_build_type + ' --']
     job.run(cmd)
 
     if ros1_bridge_ignore_marker:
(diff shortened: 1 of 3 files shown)


ros2_batch_job/packaging.py
```diff
@@ -73,10 +73,10 @@ def build_and_package(args, job):
             '--only-packages', 'ros1_bridge',
         ] + (['--isolated'] if args.isolated else []) +
             (
-                ['--cmake-args', '-DCMAKE_BUILD_TYPE=' + args.cmake_build_type]
+                ['--cmake-args', '-DCMAKE_BUILD_TYPE=' + args.cmake_build_type + ' -- ']
                 if args.cmake_build_type else []
             ) + [
-            '--make-flags', '-j1'
+            '--make-flags', '-j1', ' -- '
```
dirk-thomas: The added string should neither have a leading nor trailing space.

All other changes shouldn't have a trailing space.

mikaelarguedas: If I understand well that means that we should never add trailing whites spaces along end delimiters right? If I did it is because I copied what was done in https://github.com/ros2/ci/blob/master/ros2_batch_job/__main__.py#L230-L235, I guess I should remove those too then?

dirk-thomas: Yes, all these strings are items in a list which is later joined with spaces. So none of the items need spaces around them.

mikaelarguedas: ok thanks for the explanation. I'll modify other occurrences too.

mikaelarguedas:
This doesn't bring anything for the beta, I'll put it back in ready and we can merge it after the release

mikaelarguedas:
ready for review

mikaelarguedas:
I can merge it but won't be able to push the changes to the farm from here. Can anybody do it ? otherwise I'll merge/deploy this early next week

dirk-thomas:
Deployed.

mikaelarguedas:
thanks!

PR #22, opened 2016-12-19
test bridge by default

dirk-thomas:
[Build Status](http://ci.ros2.org/job/packaging_linux/517/)


wjwwood:
Is this for CI or just packaging?

dirk-thomas:
We only build (and with this change test) the bridge in the packaging job on Linux.

PR #25, opened 2017-02-14
Packaging test args

dhood:
This is to avoid flaky tests when testing the bridge (e.g. http://ci.ros2.org/view/packaging/job/packaging_linux/580/#showFailuresLink)

I have tested this on http://ci.ros2.org/view/All/job/test_packaging_linux

### DIFF
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -38,6 +38,11 @@
           <description>By setting this to True, the tests for the ros1_bridge will be run.</description>
           <defaultValue>@(test_bridge_default)</defaultValue>
         </hudson.model.BooleanParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_AMENT_TEST_ARGS</name>
+          <description>Additional arguments passed to 'ament test' if testing the bridge.</description>
+          <defaultValue>@(ament_test_args_default)</defaultValue>
+        </hudson.model.StringParameterDefinition>
       </parameterDefinitions>
     </hudson.model.ParametersDefinitionProperty>
   </properties>
@@ -93,7 +98,8 @@ ci_branch: ${build.buildVariableResolver.resolve('CI_SCRIPTS_BRANCH')}, <br/>
 repos_url: ${build.buildVariableResolver.resolve('CI_ROS2_REPOS_URL')}, <br/>
 used_rmw_impl: ${build.buildVariableResolver.resolve('CI_USED_RMW_IMPL')}, <br/>
 cmake_build_type: ${build.buildVariableResolver.resolve('CI_CMAKE_BUILD_TYPE')}, <br/>
-test_bridge: ${build.buildVariableResolver.resolve('CI_TEST_BRIDGE')}\
+test_bridge: ${build.buildVariableResolver.resolve('CI_TEST_BRIDGE')}, <br/>
+ament_test_args: ${build.buildVariableResolver.resolve('CI_AMENT_TEST_ARGS')}\
 """);]]>
         </command>
       </scriptSource>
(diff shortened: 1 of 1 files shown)


dhood:
to clarify: this avoids flaky tests because [the default ament-test-args include `--retest-until-pass 5`](https://github.com/ros2/ci/blob/5727cf1fe1b5b8c796028ef9ccc6b171305b29e7/create_jenkins_job.py#L80)

job_templates/packaging_job.xml.em
```diff
@@ -134,7 +134,7 @@ if [ -n "${CI_AMENT_TEST_ARGS+x}" ]; then
   if [[ "$CI_AMENT_TEST_ARGS" != *-- ]]; then
     CI_AMENT_TEST_ARGS="$CI_AMENT_TEST_ARGS --"
   fi
-  export CI_ARGS="$CI_ARGS --ament-test-args $CI_AMENT_TEST_ARGS"
+  export CI_ARGS="$CI_ARGS --ament-test-args $CI_AMENT_TEST_ARGS --"
```
mikaelarguedas: Can you explain why this was necessary ? I would have expected line134 to take care of adding the delimiters ? I think this will break if you pass test argument that already have the `--` delimiter.
Maybe the condition is too strict and should check if the AMENT_TEST_ARGS ends-with `--` or `-- `?

dirk-thomas: Good point this should probably match the same code in the non-packaging jobs: https://github.com/ros2/ci/blob/eace1b0f2600222a26c75fbf919593d6cf06ee3f/job_templates/ci_job.xml.em#L149

dirk-thomas: http://ci.ros2.org/job/packaging_osx/511/

dhood: l134 adds the delimiters beforehand, not afterwards.
Without the delimiter added at the end, it fails with:
```
"/home/rosbuild/ci_scripts/venv/bin/python" -u ""./src/ament/ament_tools/scripts/ament.py"" test "src" --build-space "build" --install-space "install" --only-packages ros1_bridge --skip-build --skip-install --retest-until-pass 5 --ros1-path /opt/ros/kinetic
02:31:45 usage: ament.py [-h]
02:31:45                 [build | build_pkg | list_packages | package_name | package_version | test | test_pkg | test_results | uninstall | uninstall_pkg]
02:31:45                 ...
02:31:45 ament.py: error: unrecognized arguments: --ros1-path /opt/ros/kinetic
```

I believe because `--ros1-path` is consumed as part of the test args group [here](https://github.com/ros2/ci/blob/c1e34ffc70da8b0ead7ee94d03eae52225f3a265/ros2_batch_job/__main__.py#L150)

dhood: I was incorrect in saying that l134 adds the delimter beforehand. 
I'll investigate why it's not appending `--` as expected at L134 (as seen in http://ci.ros2.org/job/test_packaging_linux/3/console)
the difference to the non-packing jobs is that the `--ros1-path` arg gets passed afterwards.

dhood: If you look closely at that console output you see:
```
13:04:28 + export CI_ARGS=--do-venv --force-ansi-color --fastrtps --test-bridge --cmake-build-type RelWithDebInfo
13:04:28 + [ -n x ]
13:04:28 + [[ --retest-until-pass 5 != *-- ]]
13:04:28 /tmp/hudson3577370792795485862.sh: 24: /tmp/hudson3577370792795485862.sh: [[: not found
13:04:28 + export CI_ARGS=--do-venv --force-ansi-color --fastrtps --test-bridge --cmake-build-type RelWithDebInfo --ament-test-args --retest-until-pass 5
13:04:28 + export CI_ARGS=--do-venv --force-ansi-color --fastrtps --test-bridge --cmake-build-type RelWithDebInfo --ament-test-args --retest-until-pass 5 --ros1-path /opt/ros/kinetic
```

It was never appending `--` because of an error with the `[[` syntax (it's using dash).

I've opened https://github.com/ros2/ci/pull/26

PR #28, opened 2017-02-18
Increase retest-until-pass for nightlies, packaging and ci jobs

dhood:
**Goal**: make "build became unstable" emails worth paying attention to
**Pros**: simple fix for the "short term" (until we fix all of the flaky tests)
**Cons**: if there are failing non-flaky tests, they will fail 10 times instead of 5. This will make CI jobs take longer if they have failing tests. But, when there are only flaky tests that are failing (as should be the case for nightlies), the time increase is not much (provided that the rate of failure is low, which is the case for our flaky service tests).

Personally I think that the value of not having to filter through failing tests in CI jobs justifies the extended length of CI jobs that introduce test failures.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -77,7 +77,7 @@ def main(argv=None):
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
         'ament_build_args_default': '--parallel',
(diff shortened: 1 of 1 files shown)


dhood:
Deployed to ci.ros2.org

PR #29, opened 2017-02-22
send emails to ros2_build_farm mailing list

mikaelarguedas:
With this PR, unsuccessfull jobs will now send emails to `ros2_build_farm` instead of `ros`

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -140,7 +140,7 @@ def main(argv=None):
         # all following jobs are triggered nightly with email notification
         job_data['time_trigger_spec'] = '30 7 * * *'
         # for now, skip emailing about Windows failures
-        job_data['mailer_recipients'] = 'ros@osrfoundation.org'
+        job_data['mailer_recipients'] = 'ros2-buildfarm@googlegroups.com'
 
         # configure packaging job
         # no aarch64 packaging yet


dirk-thomas:
How can individuals subscribe / unsubscribe to this list on their own?

mikaelarguedas:
I planned on inviting everybody on the ROS team at first and then allow anybody with an email under `osrfoundation.org` domain to unsubscribe/resubscribe. I'm looking into list configuration right now. I'll put this PR in review once configuration of the mailing list is ready.
Any comment on the strategy to manage members and subscription is welcome

dirk-thomas:
I would like to be able to subscribe and unsubscribe without the need for another person. Afaik that is not the case for ros@osrf. It is possible for the Google mailing lists we use for the ROS 1 farm though.

tfoote:
ros@osrf is a google group too: https://groups.google.com/a/osrfoundation.org/forum/#!forum/ros

mikaelarguedas:
the current configuration should allow you to join without the need for another  person. If any of you is willing to give it a try, I'd be happy to adjust settings if it doesn't behave the way we want it to

nuclearsandwich:
Stumbled upon this while working on other PRs. The "conflict" is trivial, the nightly job start time was changed in https://github.com/ros2/ci/commit/6dd1a5047d1b0ba5c66fa09bd47351d9f2603e53

A quick edit would make this mergeable again if the change is desired.

mikaelarguedas:
Sounds good, thanks for pointing it out!
I'll resolve once the mailing list works as expected and this is ready for merging

mikaelarguedas:
Rebased and updated

dirk-thomas:
The patch looks good.

Can you please configure the new mailing list in the same way as the existing ROS1 buildfarm mailing list.

mikaelarguedas:
hmm I thought that's what I did. Could you point me to the discrepancies so that I can adapt the settings ? otherwise as an owner you should be able to make the appropriate changes directly if you prefer

dirk-thomas:
I couldn't even see the settings as a member. Now being an owner I updated the posting access to "Not allowed" for all everyone except the buildfarm email. I also changed the delivery for the buildfarm email to "No email" (since that inbox is not checked by anyone). Currently the delivery option is set to "All email" for everyone else. Should it stay on and anybody who doesn't want to get the emails change hisdelivery setting?

mikaelarguedas:
>I couldn't even see the settings as a member. Now being an owner I updated the posting access to "Not allowed" for all everyone except the buildfarm email. 

Yeah I kept the posting before for testing purposes, thanks for putting it back to "Not Allowed".

> I also changed the delivery for the buildfarm email to "No email" (since that inbox is not checked by anyone). 

:+1: 

> Currently the delivery option is set to "All email" for everyone else. Should it stay on and anybody who doesn't want to get the emails change hisdelivery setting?

I figured we will let people know at buildcop selection time this afternoon how to change it and anybody will then change their own setting according to their needs.

Thanks for looking into it

mikaelarguedas:
Merging this. For now all emails will go to the same mailing list. I'll deploy the changes tonight before the nightlies

dirk-thomas:
The global email address configured in Jenkins which is e.g. used for the node watcher needs to be updated too.

mikaelarguedas:
Yup I'll update it at the same time this change is deployed

mikaelarguedas:
deployed along #41 
Every node + Jenkins Sys admin addresses updated accordingly

PR #30, opened 2017-03-01
make the aarch64 ci job run nightly

gerkey:
This change has been deployed to the farm. Let's wait until tomorrow to see a successful overnight run before merging the change.

Here is the diff that was deployed (all other jobs unchanged):
~~~
Updating job 'ci_linux-aarch64'
    <<<
    --- remote config
    +++ new config
    @@ -145 +145,5 @@
    -  <triggers />
    +  <triggers>
    +    <hudson.triggers.TimerTrigger>
    +      <spec>0 11 * * *</spec>
    +    </hudson.triggers.TimerTrigger>
    +  </triggers>
    @@ -432,0 +437,5 @@
    +    <hudson.tasks.Mailer plugin="mailer@1.19">
    +      <recipients>ros@osrfoundation.org</recipients>
    +      <dontNotifyEveryUnstableBuild>true</dontNotifyEveryUnstableBuild>
    +      <sendToIndividuals>false</sendToIndividuals>
    +    </hudson.tasks.Mailer>
    >>>
~~~

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -137,7 +137,7 @@ def main(argv=None):
             continue
 
         # all following jobs are triggered nightly with email notification
-        job_data['time_trigger_spec'] = '0 11 * * *'
+        job_data['time_trigger_spec'] = '30 7 * * *'
         # for now, skip emailing about Windows failures
         job_data['mailer_recipients'] = 'ros@osrfoundation.org'
 
@@ -192,7 +192,7 @@ def main(argv=None):
             job_name = 'nightly_' + os_name + '_repeated'
             if os_name == 'win':
                 job_name = job_name[0:-5]
-            job_data['time_trigger_spec'] = '0 12 * * *'
+            job_data['time_trigger_spec'] = '30 7 * * *'
             job_data['cmake_build_type'] = 'None'
             job_data['ament_test_args_default'] = '--retest-until-fail 20 --ctest-args -LE linter --'
             job_config = expand_template('ci_job.xml.em', job_data)
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -126,9 +126,9 @@ def main(argv=None):
         job_name = 'ci_' + os_name
         job_data['cmake_build_type'] = 'None'
         # For the aarch64 job:
-	#  * do it nightly (because it takes a very long time to run); and
-	#  * don't email on test failures (because there are many, likely
-	#    related to qemu)
+        #  * do it nightly (because it takes a very long time to run); and
+        #  * don't email on test failures (because there are many, likely
+        #    related to qemu)
         if os_name == 'linux-aarch64':
             job_data['time_trigger_spec'] = '0 11 * * *'
             job_data['mailer_recipients'] = 'ros@osrfoundation.org'
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -125,19 +125,11 @@ def main(argv=None):
         # configure manual triggered job
         job_name = 'ci_' + os_name
         job_data['cmake_build_type'] = 'None'
-        # For the aarch64 job:
-        #  * do it nightly (because it takes a very long time to run); and
-        #  * don't email on test failures (because there are many, likely
-        #    related to qemu)
-        if os_name == 'linux-aarch64':
-            job_data['time_trigger_spec'] = '0 11 * * *'
-            job_data['mailer_recipients'] = 'ros@osrfoundation.org'
-            job_data['dont_notify_every_unstable_build'] = 'true'
         job_config = expand_template('ci_job.xml.em', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
-        # skip packaging and special nightly jobs on ARM for now
-        if os_name == 'linux-armhf' or os_name == 'linux-aarch64':
+        # skip packaging and nightly jobs on ARM32 for now
+        if os_name == 'linux-armhf':
             continue
 
         # all following jobs are triggered nightly with email notification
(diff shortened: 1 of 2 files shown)


create_jenkins_job.py
```diff
@@ -124,10 +125,18 @@ def main(argv=None):
         # configure manual triggered job
         job_name = 'ci_' + os_name
         job_data['cmake_build_type'] = 'None'
+        # For the aarch64 job:
+        #  * do it nightly (because it takes a very long time to run); and
+        #  * don't email on test failures (because there are many, likely
+        #    related to qemu)
+        if os_name == 'linux-aarch64':
+            job_data['time_trigger_spec'] = '0 11 * * *'
```
dirk-thomas: Does this trigger the job `ci_linux-aarch64` every night? The nightly triggered build should be a separate job with the `nightly_` prefix while the `ci_` job is used for manually triggering only.

gerkey: I considered this and decided that it would be more consistent to keep the `ci_` prefix because this job is configured like the other `ci_` jobs, as opposed to being configured like the other `nightly_` jobs (each of which does something special, like coverage, non-None build type, or coverage). As it is now, it'll run nightly, and also be available for manual triggering.

Do you feel strongly that this naming scheme is unclear? If so, what would you suggest for the new nightly job? Would `nightly_linux-aarch64`, with no suffix, make sense?

dirk-thomas: The reason why we have separate jobs is that the history of the nightly jobs allows us to see how the results change over time. When a single job is used for both this advantage is lost.

`nightly_linux-aarch64` sounds good to me if it uses the `None` CMake build type.

gerkey: In 3ffdb1f I reverted the changes to `ci_linux-aarch64` and create a new nightly job. I decided for it to be a `Debug` build (the exact build type doesn't matter that much; the goal here is to check for successful compilation):

http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/

gerkey:
Now that there's a new nightly job (instead of reusing the existing `ci_` job), we'll need to wait for two overnight runs: the first run should generate an Unstable email, while the second run should not send email.

gerkey:
I forgot to disable the linter tests for the ARM jobs, which we want to do to keep the job time sort of manageable. I did that in 5386159, which resulted in the following diff on deployment:
~~~
Updating job 'ci_linux-aarch64'
    <<<
    --- remote config
    +++ new config
    @@ -107 +107 @@
    -          <defaultValue>--retest-until-pass 10</defaultValue>
    +          <defaultValue>--retest-until-pass 10 --ctest-args -LE linter --</defaultValue>
    >>>
The Jenkins master does not require a crumb
    <<<
    --- remote config
    +++ new config
    @@ -107 +107 @@
    -          <defaultValue>--retest-until-pass 10</defaultValue>
    +          <defaultValue>--retest-until-pass 10 --ctest-args -LE linter --</defaultValue>
    >>>
Updating job 'ci_linux-armhf'
    <<<
    --- remote config
    +++ new config
    @@ -107 +107 @@
    -          <defaultValue>--retest-until-pass 10</defaultValue>
    +          <defaultValue>--retest-until-pass 10 --ctest-args -LE linter --</defaultValue>
    >>>
~~~

gerkey:
First build worked as expected, giving an Unstable result and generating an email:

http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1/

The next build should also give an Unstable result, but not generate an email.

create_jenkins_job.py
```diff
@@ -163,22 +177,26 @@ def main(argv=None):
...
-        job_data['cmake_build_type'] = 'None'
-        job_data['ament_test_args_default'] = '--retest-until-fail 20 --ctest-args -LE linter --'
-        job_config = expand_template('ci_job.xml.em', job_data)
-        configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
+        # no aarch64 repeated job yet
+        if os_name != 'linux-aarch64':
+            job_name = 'nightly_' + os_name + '_repeated'
+            if os_name == 'win':
+                job_name = job_name[0:-5]
+            job_data['time_trigger_spec'] = '0 12 * * *'
```
dirk-thomas: This time is likely going to collide with the periodic reboot of the router. So it might be better to wait with this until the conflict is resolved.

dirk-thomas: Actually doesn't this mean "noon" which would be in the middle of the day and occupy a node for the remaining day?

dhood: think it's UTC

dirk-thomas: @dhood Oh, of course. Thanks for clarifying.

gerkey:
In e19612f I moved the nightly start times earlier, to 730 UTC (1130pm PT), to allow more time for them to finish over night. That change has been deployed.

gerkey:
The second build ran as expected, remaining Unstable and not generating email:

http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/2/

create_jenkins_job.py
```diff
@@ -150,10 +157,17 @@ def main(argv=None):
         if os_name == 'win':
             job_name = job_name[0:-2]
         job_data['cmake_build_type'] = 'Debug'
+	# For the aarch64 job, don't email on test failures (because there are
```
dirk-thomas: Please do not use tabs in Python code: fixed in c1180319d10cf166bfad16400b51f07bf7f6d8e4

gerkey: Whoops, don't know how that snuck in there. Thanks for fixing it!

PR #33, opened 2017-03-08
remove duplicate Dockerfile for packaging

dirk-thomas:
See the temporary jobs starting with `test_dirk_`: http://ci.ros2.org/view/All/


ros2_batch_job/__main__.py
```diff
@@ -61,47 +62,51 @@
...
+    else:
+        return run(args, build_and_test_and_package)
+
+
+def get_args(sysargv=None):
     parser = argparse.ArgumentParser(
         description="Builds the ROS2 repositories as a single batch job")
     parser.add_argument(
+        '--packaging', default=False, action='store_true',
+        help='Build and test the ros1_bridge as well as package the install space')
```
dhood: Since testing of the ros1_bridge is decided by [another argument](https://github.com/ros2/ci/pull/33/files#diff-dcc9b88c6bfa21d6151296e463494a16R129) I would suggest removing it from this comment

dirk-thomas: In my opinion it is good to have separate options for building the bridge and packaging the software. While we currently only build the bridge when packaging that doesn't have to be the case. Therefore I think it would be good to keep both options.

dhood: yes I agree with you; I was just speaking about the help mentioning that it will test the bridge, when this isn't always the case

dirk-thomas: Oh sorry, you only meant removing it from the help message. I didn't get it.

How does this look: https://github.com/ros2/ci/pull/33/commits/a5ff6c4e14dbe70d829676f5aeb533a15ff3bab2#diff-dcc9b88c6bfa21d6151296e463494a16R84

dhood: :+1:

Issue #32, opened 2017-03-08
Turtlebot CI job


dhood:
Here's what I have in mind for creating a CI job for the turtlebot demo. I am looking to get the direction approved before speaking with @nuclearsandwich about how to implement it.

As I imagine it, this will be a manually-triggered CI job while we're developing the demo and then can switch to a nightly once it's stable.

1. Add the supplementary repos file as a jenkins parameter
    1. Modify [the ci_job.xml template](https://github.com/ros2/ci/blob/6dd1a5047d1b0ba5c66fa09bd47351d9f2603e53/job_templates/ci_job.xml.em) to include this parameter if something like `supplementary_repos_file` is set when evaluating the template.
1. Add the ci_linux-turtlebot_demo job to the farm
    1. Add it to the [create_jenkins_job.py script](https://github.com/ros2/ci/blob/6dd1a5047d1b0ba5c66fa09bd47351d9f2603e53/create_jenkins_job.py)
    1. Set the default supplementary repos file to https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos or something (using a similar strategy to https://github.com/ros2/ci/commit/38f57f3f9b661d79563ce5c35955513037409413#diff-10b89ee3f1ac5a693c0ff698ad83c16dR1)
1. Pull the supplementary repos file into the source space
    1. Put a for-loop around [this section that calls vcs](https://github.com/ros2/ci/blob/6dd1a5047d1b0ba5c66fa09bd47351d9f2603e53/ros2_batch_job/__main__.py#L393..L411)
1. For turtlebot demo dependencies available as binaries, only add them to the dockerfile if necessary
    1. Inject the installation of the dependencies into the dockerfile using sed as is done [to modify the FROM command for arm jobs](https://github.com/ros2/ci/blob/6dd1a5047d1b0ba5c66fa09bd47351d9f2603e53/job_templates/ci_job.xml.em#L180) -- OR --
    1. Use the [`create_dockerfile` method from ros_buildfarm](https://github.com/ros-infrastructure/ros_buildfarm/blob/e142cedf50ad9ffa31444c55c0a34f95e520965a/ros_buildfarm/templates/__init__.py#L150) to generate the Dockerfile from a template, only including turtlebot dependency snippet if appropriate
1. Ask turtlebot demo developers to target PRs at the turtlebot2_demo.repos file, and PRs for additional binary dependencies at the turtlebot demo extra dependencies list on ros2/ci

[...]

PR #34, opened 2017-03-11
Add tinyxml and Eigen3 as dependencies

Karsten1987:
This installs `libtinyxml-dev` and `libeigen3-dev` as a dependency on the linux docker.
For mac, we have to install these packages via homebrew, for windows, we are currently in the process of writing choco packages for it.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -79,7 +79,7 @@ RUN apt-get update && apt-get install -y libopencv-dev
 RUN apt-get update && apt-get install -y libpoco-dev libpocofoundation9v5 libpocofoundation9v5-dbg
 
 # Install dependencies for robot_model and robot_state_publisher
-RUN apt-get update && apt-get install -y liburdfdom-dev liburdfdom-headers-dev libeigen3-dev
+RUN apt-get update && apt-get install -y libeigen3-dev
 
 # Install Git-LFS.
 RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install -y git-lfs; fi
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -79,7 +79,7 @@ RUN apt-get update && apt-get install -y libopencv-dev
 RUN apt-get update && apt-get install -y libpoco-dev libpocofoundation9v5 libpocofoundation9v5-dbg
 
 # Install dependencies for robot_model and robot_state_publisher
-RUN apt-get update && apt-get install -y libeigen3-dev
+RUN apt-get update && apt-get install -y libtinyxml-dev libeigen3-dev
 
 # Install Git-LFS.
 RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install -y git-lfs; fi
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -78,6 +78,9 @@ RUN apt-get update && apt-get install -y libopencv-dev
 # Install build dependencies for class loader.
 RUN apt-get update && apt-get install -y libpoco-dev libpocofoundation9v5 libpocofoundation9v5-dbg
 
+# Install orocos-kdl for robot_model and robot_state_publisher
+RUN apt-get update && apt-get install -y liborocos-kdl-dev
+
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Should this wait for https://github.com/ros2/ci/issues/32 and be added to the "only install the dependencies if appropriate" part ?

Karsten1987:
windows: [Build Status](http://ci.ros2.org/job/ci_linux/2332/)
linux: [Build Status](http://ci.ros2.org/job/ci_windows/2435/)

As discussed offline, I think these changes should go into the `ros2.repos` as I consider the `robot_state_publisher` a critical package which should be shipped to all platforms.

dirk-thomas:
Is there a reason not to use TinyXML2 (`libtinyxml2-dev`)? Newer versions of FastRTPS use version 2 so it would be good if we use it here too.

Karsten1987:
That would require a change in URDFdom / URDFdom_headers / orocos_kdl / robot_model. For now, we only have minor changes to these packages.

dirk-thomas:
I am not sure if both versions can be used at the same time which would be necessary as soon as we update to the latest FastRTPS version again.

wjwwood:
TinyXML and TinyXML2 are completely separate. They can be used at the same time. We already have this case in ROS 1, where urdf uses tinyxml(1) and other stuff uses tinyxml2. On my machine, both are installed:

```
william@farl ~
% brew list tinyxml
/usr/local/Cellar/tinyxml/2.6.2/include/tinyxml.h
/usr/local/Cellar/tinyxml/2.6.2/lib/libtinyxml.dylib
/usr/local/Cellar/tinyxml/2.6.2/lib/pkgconfig/tinyxml.pc

william@farl ~
% brew list tinyxml2
/usr/local/Cellar/tinyxml2/4.0.1/include/tinyxml2.h
/usr/local/Cellar/tinyxml2/4.0.1/lib/libtinyxml2.4.0.1.dylib
/usr/local/Cellar/tinyxml2/4.0.1/lib/pkgconfig/tinyxml2.pc
/usr/local/Cellar/tinyxml2/4.0.1/lib/ (2 other files)
```

I believe it is the same on Linux.

Karsten1987:
So it compiles on all platforms, can anybody give feedback on that?
https://github.com/ros2/demos/pull/118

wjwwood:
> I believe it is the same on Linux.

Can someone verify this that has a Xenial system handy?

Also, don't forget to update the installation (binary and source) guide on the wiki. Also for macOS and manually install these on the macOS Jenkins nodes.

Karsten1987:
didn't check for symbol collision, but so far:

```
karsten@karsten-mint ~ $ ll /usr/include/tinyxml*
-rw-r--r-- 1 root root 62897 Jul 30  2015 /usr/include/tinyxml2.h
-rw-r--r-- 1 root root 64887 Aug  5  2015 /usr/include/tinyxml.h

karsten@karsten-mint ~ $ ll /usr/lib/x86_64-linux-gnu/libtinyxml*
-rw-r--r-- 1 root root 139980 Jul 30  2015 /usr/lib/x86_64-linux-gnu/libtinyxml2.a
lrwxrwxrwx 1 root root     16 Jul 30  2015 /usr/lib/x86_64-linux-gnu/libtinyxml2.so -> libtinyxml2.so.2
lrwxrwxrwx 1 root root     20 Jul 30  2015 /usr/lib/x86_64-linux-gnu/libtinyxml2.so.2 -> libtinyxml2.so.2.2.0
-rw-r--r-- 1 root root  76296 Jul 30  2015 /usr/lib/x86_64-linux-gnu/libtinyxml2.so.2.2.0
-rw-r--r-- 1 root root 163194 Aug  5  2015 /usr/lib/x86_64-linux-gnu/libtinyxml.a
lrwxrwxrwx 1 root root     19 Aug  5  2015 /usr/lib/x86_64-linux-gnu/libtinyxml.so -> libtinyxml.so.2.6.2
-rw-r--r-- 1 root root  88768 Aug  5  2015 /usr/lib/x86_64-linux-gnu/libtinyxml.so.2.6.2
```

mikaelarguedas:
Yep, I know that several ROS1 packages depend on tinyxml2 while several low level packages in ROS depend on tinyxml. They are different packages and can coexist without problem

PR #35, opened 2017-03-14
Add turtlebot-demo jenkins job.

nuclearsandwich:
First crack at implementing the plan described in https://github.com/ros2/ci/issues/32#issuecomment-285113012

This adds a new job to the Jenkins config by adding turtlebot-demo as an additional os_config. I looked for another place like `job_configs` to do this since the turtlebot isn't a separate os per se but it looks like `os_configs` is where this happens.

The job script will build a docker image with the `INSTALL_TURTLEBOT2_DEMO_DEPS` variable set for turtlebot-demo jobs.

Then dockerfile has an added directive to install the demo dependencies when that variable exists.

Still to do

- [ ] verify what other system package dependencies should be installed
- [x] create a repos file (at https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos?)
- [x] run this in a non-dryrun capacity to test the job

cc @dhood and @tfoote for feedback and review.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -216,14 +216,14 @@ def main(argv=None):
     job_config = expand_template('ci_launcher_job.xml.em', job_data)
     configure_job(jenkins, 'ci_launcher', job_config, **jenkins_kwargs)
 
-
     # Run the turtlebot job on Linux only for now.
     os_name = 'linux'
     turtlebot_job_data = dict(data)
     turtlebot_job_data['os_name'] = os_name
     turtlebot_job_data.update(os_configs[os_name])
     turtlebot_job_data['turtlebot_demo'] = True
-    turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
+    # Will use a turtlebot2_demo repos file when one is available.
+    # turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
     turtlebot_job_data['cmake_build_type'] = 'None'
     job_config = expand_template('ci_job.xml.em', turtlebot_job_data)
     configure_job(jenkins, 'ci_turtlebot-demo', job_config, **jenkins_kwargs)
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -80,6 +80,7 @@ def main(argv=None):
         'ament_test_args_default': '--retest-until-pass 10',
         'enable_c_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
+        'turtlebot_demo': False,
     }
 
     jenkins = connect(args.jenkins_url)
@@ -110,11 +111,6 @@ def main(argv=None):
             'shell_type': 'Shell',
             'use_connext_default': 'false',
         },
-        'turtlebot-demo': {
-            'label_expression': 'linux',
-            'shell_type': 'Shell',
-            'default_repos_url': DEFAULT_REPOS_URL,
-        },
     }
 
     jenkins_kwargs = {}
@@ -139,9 +135,6 @@ def main(argv=None):
 
         # skip packaging and nightly jobs on ARM32 for now
         if os_name == 'linux-armhf':
-            continue
-        # skip packaging and nightly jobs for turtlebot CI
-        if os_name == 'turtlebot-demo':
             continue
 
         # all following jobs are triggered nightly with email notification
@@ -222,6 +215,19 @@ def main(argv=None):
     job_data['cmake_build_type'] = 'None'
     job_config = expand_template('ci_launcher_job.xml.em', job_data)
     configure_job(jenkins, 'ci_launcher', job_config, **jenkins_kwargs)
+
+
+    # Run the turtlebot job on Linux only for now.
+    os_name = 'linux'
+    turtlebot_job_data = dict(data)
+    turtlebot_job_data['os_name'] = os_name
+    turtlebot_job_data.update(os_configs[os_name])
+    turtlebot_job_data['turtlebot_demo'] = True
+    turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
+    turtlebot_job_data['cmake_build_type'] = 'None'
+    job_config = expand_template('ci_job.xml.em', turtlebot_job_data)
+    configure_job(jenkins, 'ci_turtlebot-demo', job_config, **jenkins_kwargs)
+
 
 
 if __name__ == '__main__':
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -110,6 +110,11 @@ def main(argv=None):
             'shell_type': 'Shell',
             'use_connext_default': 'false',
         },
+        'turtlebot-demo': {
+            'label_expression': 'linux',
+            'shell_type': 'Shell',
+            'default_repos_url': DEFAULT_REPOS_URL,
+        },
     }
 
     jenkins_kwargs = {}
@@ -134,6 +139,9 @@ def main(argv=None):
 
         # skip packaging and nightly jobs on ARM32 for now
         if os_name == 'linux-armhf':
+            continue
+        # skip packaging and nightly jobs for turtlebot CI
+        if os_name == 'turtlebot-demo':
             continue
 
         # all following jobs are triggered nightly with email notification
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -222,8 +222,8 @@ def main(argv=None):
     turtlebot_job_data['os_name'] = os_name
     turtlebot_job_data.update(os_configs[os_name])
     turtlebot_job_data['turtlebot_demo'] = True
-    # Will use a turtlebot2_demo repos file when one is available.
-    # turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
+    # Use a turtlebot2_demo-specific repos file by default.
+    turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
     turtlebot_job_data['cmake_build_type'] = 'None'
     job_config = expand_template('ci_job.xml.em', turtlebot_job_data)
     configure_job(jenkins, 'ci_turtlebot-demo', job_config, **jenkins_kwargs)
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -187,7 +187,7 @@ docker build --build-arg PLATFORM=arm -t ros2_batch_ci_aarch64 linux_docker_reso
 docker build --build-arg INSTALL_TURTLEBOT2_DEMO_DEPS=true -t ros2_batch_ci_turtlebot_demo linux_docker_resources
 @[else]@
 docker build -t ros2_batch_ci linux_docker_resources
-@[endif]@
+@[end if]@
 @[else]@
 @{ assert 'Unknown os_name: ' + os_name }@
 @[end if]@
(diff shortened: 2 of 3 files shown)


nuclearsandwich:
> I think this will work, though I do feel it might be abusing the os_configs.

I am in complete agreement, I looked (admittedly so far only within this script) for a larger global job_configs structure and didn't find it. My impetus for wanting to use an existing declarative structure rather than copy the loop contents separately was to avoid creating two separate invocations that need to be maintained.

It seems a bit ambitions but I suppose that the common job creation elements could be factored into functions to avoid strict duplication.

nuclearsandwich:
> create a new class of job which is orthogonal to which os it is run on for this. Either way, thanks for looking into it!

I've moved the job outside of the os_configs reliance but it's not yet on par with the other jobs as I'm not sure if we should be creating turtlebot jobs for all platforms. If we do want to do that right now, then we'll need to identify and install dependencies on the Windows and Mac Jenkins nodes. Additionally, we'll need to inject build-args into the docker build invocation a little more intelligently as right now the turtlebot job overrides the `os_name` and uses the standard linux invocation with one additional build-arg.

job_templates/ci_job.xml.em
```diff
@@ -176,7 +176,9 @@ echo "# BEGIN SECTION: docker info"
 docker info
 echo "# END SECTION"
 echo "# BEGIN SECTION: Build Dockerfile"
-@[if os_name == 'linux-armhf']@
+@[if turtlebot_demo]@
```
dhood: Rather than add another branch here, you can use the existing `os_name=='linux'` invocation and just specify the `INSTALL_TURTLEBOT2_DEMO_DEPS` arg with something like: 

```
@[elif os_name == 'linux']@
docker build @(--build-arg INSTALL_TURTLEBOT2_DEMO_DEPS='yes' if turtlebot_demo) -t ros2_batch_ci_turtlebot_demo linux_docker_resources
@[else]@

File: linux_docker_resources/Dockerfile [hunk 2]
```diff
@@ -38,6 +38,9 @@ RUN apt-get update && apt-get install -y build-essential ccache cmake pkg-config
 # Install build and test dependencies of ROS 2 packages.
 RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-flake8 python3-mock python3-nose python3-pep8 uncrustify
 
+# Install build dependencies for turtlebot demo.
+RUN if test -n ${INSTALL_TURTLEBOT2_DEMO_DEPS}; apt-get update && apt-get install -y libeigen3-dev libtinyxml-dev
```
dhood: I think you'll need to add this to the top of the dockerfile as with https://github.com/ros2/ci/commit/a5ff6c4e14dbe70d829676f5aeb533a15ff3bab2#diff-a7a7037d3453364950fd80f5f91ed7eeR3

dhood: I think you will also want to move this to after [this line](https://github.com/ros2/ci/commit/a5ff6c4e14dbe70d829676f5aeb533a15ff3bab2#diff-a7a7037d3453364950fd80f5f91ed7eeR87) that prints a different string each day, so that (1) the docker images only diverge once necessary and (2) the docker image gets invalidated daily. this ensures that "fresh" dependencies get installed.

Otherwise, if the _text_ in the dockerfile of what dependencies to install does not change, it will not get re-built even if there are newer versions of dependencies that ought to be installed. For system dependencies it might not matter much but we will probably end up installing ROS packages, which get released regularly. (@dirk-thomas please chime in if this doesn't match what you explained to me the other day!)

wjwwood: bump (the email outage likely did not send the previous comment's notification)

nuclearsandwich:
> You can feel free to test non-dryrun when you want, as this should not affect the existing jobs, just add a new one.

The primary thing holding this back is the lack of a repos file that is turtlebot specific. I suppose I could set it up with the default repos file and make sure that it builds the standard stuff. I'll do that.

nuclearsandwich:
Testing with the default repos file.

job_templates/ci_job.xml.em
```diff
@@ -183,7 +183,11 @@ docker build --build-arg PLATFORM=arm -t ros2_batch_ci_armhf linux_docker_resour
 sed -i 's+^FROM.*$+FROM osrf/ubuntu_arm64:xenial+' linux_docker_resources/Dockerfile
 docker build --build-arg PLATFORM=arm -t ros2_batch_ci_aarch64 linux_docker_resources
 @[elif os_name == 'linux']@
+@[if turtlebot_demo]@
+docker build --build-arg INSTALL_TURTLEBOT2_DEMO_DEPS=true -t ros2_batch_ci_turtlebot_demo linux_docker_resources
+@[else]@
 docker build -t ros2_batch_ci linux_docker_resources
+@[end if]@
```
dirk-thomas: You might want to use to spaces if indentation to make it more readable:

```
@[  if turtlebot_demo]@
docker build --build-arg INSTALL_TURTLEBOT2_DEMO_DEPS=true -t ros2_batch_ci_turtlebot_demo linux_docker_resources
@[  else]@
docker build -t ros2_batch_ci linux_docker_resources
@[  end if]@
```

nuclearsandwich: I definitely can. Is that kind of indentation a common pattern in empy templates/

dirk-thomas: We try to use it in our templates (if we don't forget :wink:).

nuclearsandwich: https://github.com/ros2/ci/pull/35/commits/62916e6ae1f0198218fa1993110dd11bfa515c05

create_jenkins_job.py
```diff
@@ -216,14 +216,14 @@ def main(argv=None):
...
-
     # Run the turtlebot job on Linux only for now.
     os_name = 'linux'
     turtlebot_job_data = dict(data)
     turtlebot_job_data['os_name'] = os_name
     turtlebot_job_data.update(os_configs[os_name])
     turtlebot_job_data['turtlebot_demo'] = True
-    turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
+    # Will use a turtlebot2_demo repos file when one is available.
+    # turtlebot_job_data['default_repos_url'] = 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos'
```
dhood: This is totally fine, but just for your understanding of how the things inter-relate: even with the `default_repos_url` set to a url that doesn't exist yet, it is just the default. The url can be specified to be something other than the default using the `CI_ROS2_REPOS_URL` on jenkins when doing "Build with Parameters"

job_templates/ci_job.xml.em
```diff
@@ -176,16 +176,18 @@ echo "# BEGIN SECTION: docker info"
...
-@[elif os_name == 'linux-armhf']@
+@[if os_name == 'linux-armhf']@
 sed -i 's+^FROM.*$+FROM osrf/ubuntu_armhf:xenial+' linux_docker_resources/Dockerfile
 docker build --build-arg PLATFORM=arm -t ros2_batch_ci_armhf linux_docker_resources
 @[elif os_name == 'linux-aarch64']@
 sed -i 's+^FROM.*$+FROM osrf/ubuntu_arm64:xenial+' linux_docker_resources/Dockerfile
 docker build --build-arg PLATFORM=arm -t ros2_batch_ci_aarch64 linux_docker_resources
 @[elif os_name == 'linux']@
+@[if turtlebot_demo]@
+docker build --build-arg INSTALL_TURTLEBOT2_DEMO_DEPS=true -t ros2_batch_ci_turtlebot_demo linux_docker_resources
```
dhood: > I started using an inline postfix conditional as in the suggestion but
I'm not sure if we want separate docker image names for the turtlebot
build and the regular linux CI build. It seems like we do, so I made
this a full nested conditional.

:+1: I didn't think about the name so I agree that this is better as you have it now

Thanks for the other changes in this commit too

create_jenkins_job.py
nuclearsandwich: > it is just the default. The url can be specified to be something other than the default using the CI_ROS2_REPOS_URL on jenkins when doing "Build with Parameters"

Understood, I just didn't want it to be broken by default.

nuclearsandwich:
Sidebar: My editor is crying to me about trailing whitespace on [two](https://github.com/ros2/ci/blob/baa2befd19e633d98957ac3cd455e366014efea8/job_templates/ci_job.xml.em#L146) [lines](https://github.com/ros2/ci/blob/baa2befd19e633d98957ac3cd455e366014efea8/job_templates/ci_job.xml.em#L157) in the job template. Is that whitespace significant or should I just clean that up?

dhood:
not significant, feel free to remove it

nuclearsandwich:
After a couple of errors were ironed out we have a successful build of the default repos file http://ci.ros2.org/job/ci_turtlebot-demo/3/console

nuclearsandwich:
From https://github.com/ros2/turtlebot2_demo/#install-some-dependencies

The dependencies are `ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi ros-kinetic-common-msgs ros-kinetic-astra-camera libusb-1.0.0-dev libudev-dev`

Does this mean that tinyxml and eigen aren't dependencies for this demo, or that they're pulled in transitively by this dependency set?

mikaelarguedas:
These were the dependencies of the previous demo, they will likely be other ones for this demo (I think that at least tinyxml will be necessary to get the `robot_state_publisher` that @Karsten1987 ported recently to work).

I think it's fine to go with a small set of deps for now and let people add the corresponding dependencies to the dockerfile at the same time they add their packages to the turtlebot repo file. 

Regarding this set of dependencies: I didn't chase down the dependency list but very likely these ros packages will pull in most of the dependencies needed for this demo. There is some overlap too: for example ros-kinetic-astra-camera will pull in `libusb-1.0.0-dev` and `libudev-dev` automatically

dhood:
> After a couple of errors were ironed out we have a successful build of the default repos file http://ci.ros2.org/job/ci_turtlebot-demo/3/console

Cool, nice work! Let's revert https://github.com/ros2/ci/pull/35/commits/12545d45254649038bb8821f7a87288e6b607b7a, push the standard ros2.repos file to https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos as a placeholder and merge this. Then we can iterate on the repos file and the dependencies to be installed afterwards.

nuclearsandwich:
> Let's revert 12545d4, push the standard ros2.repos file to https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos as a placeholder and merge this. Then we can iterate on the repos file and the dependencies to be installed afterwards.

https://github.com/ros2/ci/pull/35/commits/9b4ef463175c0728a0f456b76dbd065937ab42b4 sets the default repos url and dry-ran acceptably.

Do  you want to make the repos file push @dhood or shall I PR it?

dhood:
> 9b4ef46 sets the default repos url and dry-ran acceptably.

You will have to re-run `create_jenkins_job.py` to push the new config to the job. (The default listed in the description of the `CI_ROS2_REPOS_URL` parameter should update accordingly)

> Do you want to make the repos file push @dhood or shall I PR it?

happy for you to thanks and then turtlebot demo developers can iterate on it

nuclearsandwich:
http://ci.ros2.org/job/ci_turtlebot-demo/7/ passed once I remembered to set the ci scripts branch. Eliminating that problem by merging this PR now.

dhood:
we usually delete branches once merged if you could please do so when you're done with this branch. Thanks for working on it! :tada:

dhood:
Ah also we generally prefer squash-and-merge -- my fault for not mentioning it (it is the default on most repos but not all)

dirk-thomas:
Can we please squash the already committed commits on the master branch? Otherwise the verbose history increases future effort in reviewing / tracing past changes significantly.

nuclearsandwich:
> Can we please squash the already committed commits on the master branch? Otherwise the verbose history increases future effort in reviewing / tracing past changes significantly.

Before I force push to master, can I confirm that this is what you'd like me to do?

dirk-thomas:
@nuclearsandwich Yes, force push to master to clean up the history. You can also checkout master locally and squash the commits with `git rebase -i HEAD~<N>`.

nuclearsandwich:
Master has been force pushed. Anyone who pulled the previous history of master will need to create a new clone or forcibly update their local clone with

```
git checkout master
git fetch origin master
git reset --hard origin/master
```

PR #38, opened 2017-03-23
disable package parallelism on arm build

gerkey:
This change has been deployed to the farm so that we can get overnight test results. The config diff was:
~~~
    <<<
    --- remote config
    +++ new config
    @@ -102 +102 @@
    -          <defaultValue>--parallel</defaultValue>
    +          <defaultValue />
    >>>
~~~

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -162,13 +162,11 @@ def main(argv=None):
         # many, likely related to qemu). Also disable linter tests (because
         # this is already a very slow job and because we lint plenty on other
         # jobs). Also don't build packages in parallel because we've seen hung
-        # builds that seem to be caused by parallelism. Also don't rerun tests,
-        # as we're currently not paying attention to test failures.
+        # builds that seem to be caused by parallelism.
         if os_name == 'linux-aarch64':
             job_data['dont_notify_every_unstable_build'] = 'true'
-            job_data['ament_test_args_default'] = '--retest-until-pass 10 --ctest-args -LE linter --'
+            job_data['ament_test_args_default'] = '--ctest-args -LE linter --'
             job_data['ament_build_args_default'] = ''
-            job_data['ament_test_args_default'] = ''
         job_config = expand_template('ci_job.xml.em', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -162,11 +162,13 @@ def main(argv=None):
         # many, likely related to qemu). Also disable linter tests (because
         # this is already a very slow job and because we lint plenty on other
         # jobs). Also don't build packages in parallel because we've seen hung
-        # builds that seem to be caused by parallelism.
+        # builds that seem to be caused by parallelism. Also don't rerun tests,
+        # as we're currently not paying attention to test failures.
         if os_name == 'linux-aarch64':
             job_data['dont_notify_every_unstable_build'] = 'true'
             job_data['ament_test_args_default'] = '--retest-until-pass 10 --ctest-args -LE linter --'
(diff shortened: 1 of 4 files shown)


gerkey:
Thanks, @wjwwood. I'm going to wait until we get some results before merging.

wjwwood:
Bump, were there any results from this? It looks like the debug job is still using parallel, and still getting hung up:

http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/28/console

dhood:
that's my fault that it's using parallel again: I overwrote it. I'll change it back manually. Sorry for stalling the investigation!

wjwwood:
There could have been results from previous jobs, so I don't think you messed anything up. I was just curious what the conclusion was.

gerkey:
I made another change, with the goal of reducing build times: remove the `--retest-until-pass` arg. I deployed the change, which produced this diff:
~~~
    <<<
    --- remote config
    +++ new config
    @@ -107 +107 @@
    -          <defaultValue>--retest-until-pass 10 --ctest-args -LE linter --</defaultValue>
    +          <defaultValue>--ctest-args -LE linter --</defaultValue>
    @@ -437 +437 @@
    -    <hudson.tasks.Mailer plugin="mailer@1.20">
    +    <hudson.tasks.Mailer plugin="mailer@1.19">
    >>>
~~~
Let's get another overnight result to check the effect.

create_jenkins_job.py
```diff
@@ -162,11 +162,13 @@ def main(argv=None):
...
         # this is already a very slow job and because we lint plenty on other
         # jobs). Also don't build packages in parallel because we've seen hung
-        # builds that seem to be caused by parallelism.
+        # builds that seem to be caused by parallelism. Also don't rerun tests,
+        # as we're currently not paying attention to test failures.
         if os_name == 'linux-aarch64':
             job_data['dont_notify_every_unstable_build'] = 'true'
             job_data['ament_test_args_default'] = '--retest-until-pass 10 --ctest-args -LE linter --'
             job_data['ament_build_args_default'] = ''
+            job_data['ament_test_args_default'] = ''
```
mikaelarguedas: Do we want to run tests at all if we dont check the results ? that would be another use case to add the "run tests" checkbox to jenkins and disable it by default

If you want to run them once anyway, maybe we should keep `--ctest-args -LE linter --` to avoid running the linters event on the first round of testing

gerkey: I went back to running the tests but excluding the linter tests (check the latest commits on this branch). I suppose that we could just disable tests, but I feel that running them once is a good middle ground between no testing and full testing, so I'd rather keep doing that if we can afford the CPU time for it.

gerkey:
The new args seem to work fine, but we still got a hung build. It had run for 10 hours, but stopped producing output 12 minutes into the build:

http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/30/consoleFull

It could be lots of things, including the CMake-related system call hangs that we've seen before under qemu.

I'm going to merge this change because I think that it's going in the right direction of improving performance of the ARM build, but we'll need to find an alternative mechanism for doing this build, likely outside qemu. In the meantime, we'll need to keep an eye on the build and kill the hung ones.

dirk-thomas:
I would rather try to avoid that the build farmer has to monitor / kill this job every morning. Can we just add a timeout to the job and automatically kill it after X hours?

gerkey:
Oh, a timeout is a great idea. Should I install the build-timeout plugin? https://wiki.jenkins-ci.org/display/JENKINS/Build-timeout+Plugin

dirk-thomas:
:+1: we are using it on the ROS 1 farm successfully.

gerkey:
See #43 for the timeout.

PR #44, opened 2017-03-29
install new FastRTPS dependencies: asio, tinyxml2

dirk-thomas:
Necessary since eProsima/Fast-RTPS#26. This allows us to switch back to use the current master branch of FastRTPS (at least on Linux).

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -70,7 +70,7 @@ ADD rticonnextdds-src/librticonnextdds52-dev_5.2.3-1_amd64.deb /tmp/librticonnex
 ADD rticonnextdds-src/rticonnextdds-tools_5.2.3-1_amd64.deb /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
 
 # Install the eProsima dependencies.
-RUN apt-get update && apt-get install -y libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev valgrind
+RUN apt-get update && apt-get install -y libasio-dev libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev valgrind
 
 # Install OpenCV.
 RUN apt-get update && apt-get install -y libopencv-dev
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -70,7 +70,7 @@ ADD rticonnextdds-src/librticonnextdds52-dev_5.2.3-1_amd64.deb /tmp/librticonnex
 ADD rticonnextdds-src/rticonnextdds-tools_5.2.3-1_amd64.deb /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
 
 # Install the eProsima dependencies.
-RUN apt-get update && apt-get install -y libasio-dev libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev valgrind
+RUN apt-get update && apt-get install -y libasio-dev libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libtinyxml2-dev valgrind
 
 # Install OpenCV.
 RUN apt-get update && apt-get install -y libopencv-dev


wjwwood:
Any idea about the story for Windows? macOS has these in brew:

```
% brew info asio
asio: stable 1.10.8 (bottled), devel 1.11.0, HEAD
Cross-platform C++ Library for asynchronous programming
https://think-async.com/Asio
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/asio.rb
==> Dependencies
Build: autoconf ✔, automake ✔
Required: openssl ✔
Optional: boost ✔
==> Options
--c++11
	Build using C++11 mode
--with-boost
	Build with boost support
--with-boost-coroutine
	Use Boost.Coroutine to implement stackful coroutines
--devel
	Install development version 1.11.0
--HEAD
	Install HEAD version

% brew info tinyxml2
tinyxml2: stable 4.0.1 (bottled), HEAD
Improved tinyxml (in memory efficiency and size)
http://grinninglizard.com/tinyxml2
/usr/local/Cellar/tinyxml2/4.0.1 (7 files, 137KB) *
  Poured from bottle on 2016-08-25 at 16:23:08
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/tinyxml2.rb
==> Dependencies
Build: cmake ✔
```

On Windows, will we need to have a `_vendor` package (like we do for `poco_vendor`) to satisfy Windows or does Fast-RTPS still embed or download and build missing deps?

mikaelarguedas:
If you clone their repo they embed these dependencies in the thirdparty directory and build them when required. I think a vendor package for those would make sense for the time being and having a chocolatey package for them in the long term would be better.

linux_docker_resources/Dockerfile
```diff
@@ -70,7 +70,7 @@ ADD rticonnextdds-src/librticonnextdds52-dev_5.2.3-1_amd64.deb /tmp/librticonnex
 ADD rticonnextdds-src/rticonnextdds-tools_5.2.3-1_amd64.deb /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
 
 # Install the eProsima dependencies.
-RUN apt-get update && apt-get install -y libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev valgrind
+RUN apt-get update && apt-get install -y libasio-dev libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libtinyxml2-dev valgrind
```
mikaelarguedas: removing all the boost dependencies should also be part of this PR now that they are not required by FastRTPS anymore

dirk-thomas: Sine we can't update to the master branch yet I didn't remove the boost dependencies yet. That will happen in a separate PR when we actually update the repos file.

mikaelarguedas: I see, :+1: then

dirk-thomas:
I have already installed both packages on all OS X nodes (dosa, mini1, mini2).

I also updated the instructions for [Linux](https://github.com/ros2/ros2/wiki/Linux-Development-Setup/_compare/929f26cb6223f4397a153b3d414453c3b0c7846b...093bc7af9df297cc7c5c50727d5e45e5ef5c5db6) and [OS X](https://github.com/ros2/ros2/wiki/OSX-Development-Setup/_compare/ebf174d2f2e45c7899de4ffc8e9f312a855724d8...07f894458e5a23101f19533f4afe9968dfa4e29c).

For Windows I am still think about the following options, e.g.:

* Manually update the submodules.
* Pass `THIRDPARTY` to the CMake invocation of FastRTPS to make it update the submodules.
* Create a Chocolatey package for `asio` and `tinyxml2` (probably the best for the long term future).
* Create a vendor package but the the problem for that is that it won't be considered as a dependency of the `FastRTPS` package. So we can't ensure that it is built before in the same workspace.

Karsten1987:
creating a chocolatey package for tinyxml2 should be rather straight
forward.

dirk-thomas:
@gerkey When updating the Linux instructions I noticed that you duplicated the block about installing the dependencies for Debian a while ago: https://github.com/ros2/ros2/wiki/Linux-Development-Setup/_compare/873858432a84f199ebac86dadaa8cce13fac8846...24ab40621654331e9e9cb3fdfb2c2fea18f6b383 I think it would be better to reference the existing instructions from Ubuntu and only describe the differences (if there are any). That would safe us the effort of maintaining both blocks and avoid the risk of them getting out of sync.

mikaelarguedas:
Yeah I'vs been reformatting the list dependencies on Debian a [few days ago](https://github.com/ros2/ros2/wiki/Linux-Development-Setup/_compare/4fd6f28f4a4b5d85fda63f053f6f36f5ca5ecee4...929f26cb6223f4397a153b3d414453c3b0c7846b) to make it more obvious when they can be collapsed. I think we can merge the instructions now that the dependencies seems identical. The only difference was where to pull vcs from. Looks like we have python3-vcstool built for all platforms in the ros1 repos so I think we're good for de-duplication now.
Only question is where should turtlebot2 specific deps land, not sure they belong on the "core install page"

nuclearsandwich:
> creating a chocolatey package for tinyxml2 should be rather straight forward.

Does this mean you're volunteering? 😈

PR #47, opened 2017-04-04
New aarch64 builds

gerkey:
This PR gets things in place to do aarch64 builds on packet.net's native servers. My goal here is to try treating aarch64 as a first-class citizen, with all the usual builds.

Because we're not using qemu, we can treat those machines like our existing x86 servers, and so I was able to remove a bunch of special logic. Along the way I decided to remove armhf; we can add that back later if there's demand.

FYI, the use of `(apt-get update || true)` in the Dockerfile is motivated by the fact that the upstream docker image we're now using (https://hub.docker.com/r/aarch64/ubuntu/) has a bad entry in its apt sources list that produces a 404, which I'm ignoring. That feels like a pretty harmless thing to do in general for the Linux builds.

This change has been deployed, with the following diff:
~~~
Connecting to Jenkins 'http://ci.ros2.org'
Connected to Jenkins version '2.46.1'
Updating job 'ci_linux'
    <<<
    --- remote config
    +++ new config
    @@ -249 +248,0 @@
    -export EXTRA_MOUNT=""
    @@ -251 +250 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
The Jenkins master does not require a crumb
Skipped 'packaging_linux' because the config is the same
Updating job 'nightly_linux_debug'
    <<<
    --- remote config
    +++ new config
    @@ -253 +252,0 @@
    -export EXTRA_MOUNT=""
    @@ -255 +254 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
Updating job 'nightly_linux_coverage'
    <<<
    --- remote config
    +++ new config
    @@ -253 +252,0 @@
    -export EXTRA_MOUNT=""
    @@ -255 +254 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
Updating job 'nightly_linux_release'
    <<<
    --- remote config
    +++ new config
    @@ -253 +252,0 @@
    -export EXTRA_MOUNT=""
    @@ -255 +254 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
Updating job 'nightly_linux_repeated'
    <<<
    --- remote config
    +++ new config
    @@ -253 +252,0 @@
    -export EXTRA_MOUNT=""
    @@ -255 +254 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
Updating job 'ci_linux-aarch64'
    <<<
    --- remote config
    +++ new config
    @@ -107 +107 @@
    -          <defaultValue>--retest-until-pass 10 --ctest-args -LE linter --</defaultValue>
    +          <defaultValue>--retest-until-pass 10</defaultValue>
    @@ -246 +246 @@
    -#sed -i 's+^FROM.*$+FROM osrf/ubuntu_arm64:xenial+' linux_docker_resources/Dockerfile
    +sed -i 's+^FROM.*$+FROM aarch64/ubuntu:xenial+' linux_docker_resources/Dockerfile
    @@ -250,7 +249,0 @@
    -export EXTRA_MOUNT=""
    -#tmpd=`mktemp -d`
    -#mkdir $tmpd/src
    -#curl -sk https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos -o $tmpd/ros2.repos
    -#vcs import $tmpd/src --input $tmpd/ros2.repos
    -#export CI_ARGS="$CI_ARGS --src-mounted --workspace-path $tmpd"
    -#export EXTRA_MOUNT="-v $tmpd/src:/home/rosbuild/ci_scripts/ws/src"
    @@ -258,2 +251 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    -rm -rf $tmpd
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
Creating job 'packaging_linux-aarch64'
    <<<
    --- remote config
    +++ new config
    @@ -102 +102 @@
    -          <defaultValue />
    +          <defaultValue>--parallel</defaultValue>
    @@ -107 +107 @@
    -          <defaultValue>--ctest-args -LE linter --</defaultValue>
    +          <defaultValue>--retest-until-pass 10</defaultValue>
    @@ -140 +140 @@
    -  <assignedNode>linux</assignedNode>
    +  <assignedNode>linux_aarch64</assignedNode>
    @@ -250 +250 @@
    -sed -i 's+^FROM.*$+FROM osrf/ubuntu_arm64:xenial+' linux_docker_resources/Dockerfile
    +sed -i 's+^FROM.*$+FROM aarch64/ubuntu:xenial+' linux_docker_resources/Dockerfile
    @@ -254,7 +253,0 @@
    -export EXTRA_MOUNT=""
    -tmpd=`mktemp -d`
    -mkdir $tmpd/src
    -curl -sk https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos -o $tmpd/ros2.repos
    -vcs import $tmpd/src --input $tmpd/ros2.repos
    -export CI_ARGS="$CI_ARGS --src-mounted --workspace-path $tmpd"
    -export EXTRA_MOUNT="-v $tmpd/src:/home/rosbuild/ci_scripts/ws/src"
    @@ -262,2 +255 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    -rm -rf $tmpd
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    @@ -439 +431 @@
    -      <dontNotifyEveryUnstableBuild>true</dontNotifyEveryUnstableBuild>
    +      <dontNotifyEveryUnstableBuild>false</dontNotifyEveryUnstableBuild>
    @@ -444,8 +435,0 @@
    -    <hudson.plugins.build__timeout.BuildTimeoutWrapper plugin="build-timeout@1.18">
    -      <strategy class="hudson.plugins.build_timeout.impl.AbsoluteTimeOutStrategy">
    -        <timeoutMinutes>600</timeoutMinutes>
    -      </strategy>
    -      <operationList>
    -        <hudson.plugins.build__timeout.operations.AbortOperation />
    -      </operationList>
    -    </hudson.plugins.build__timeout.BuildTimeoutWrapper>
    >>>
Creating job 'nightly_linux-aarch64_release'
Creating job 'nightly_linux-aarch64_repeated'
Skipped 'ci_osx' because the config is the same
Skipped 'packaging_osx' because the config is the same
Skipped 'nightly_osx_debug' because the config is the same
Skipped 'nightly_osx_release' because the config is the same
Skipped 'nightly_osx_repeated' because the config is the same
Skipped 'ci_windows' because the config is the same
Skipped 'packaging_windows' because the config is the same
Skipped 'nightly_win_deb' because the config is the same
Skipped 'nightly_win_rel' because the config is the same
Skipped 'nightly_win_rep' because the config is the same
Updating job 'ci_launcher'
    <<<
    --- remote config
    +++ new config
    @@ -135,0 +136,12 @@
    +          <projects>ci_linux-aarch64</projects>
    +          <condition>SUCCESS</condition>
    +          <triggerWithNoParameters>false</triggerWithNoParameters>
    +        </hudson.plugins.parameterizedtrigger.BuildTriggerConfig>
    +      </configs>
    +    </hudson.plugins.parameterizedtrigger.BuildTrigger>
    +    <hudson.plugins.parameterizedtrigger.BuildTrigger plugin="parameterized-trigger@2.32">
    +      <configs>
    +        <hudson.plugins.parameterizedtrigger.BuildTriggerConfig>
    +          <configs>
    +            <hudson.plugins.parameterizedtrigger.CurrentBuildParameters />
    +          </configs>
    >>>
Updating job 'ci_turtlebot-demo'
    <<<
    --- remote config
    +++ new config
    @@ -249 +248,0 @@
    -export EXTRA_MOUNT=""
    @@ -251 +250 @@
    -docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $EXTRA_MOUNT $CONTAINER_NAME
    +docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
    >>>
~~~

### DIFF
--- a/job_templates/ci_launcher_job.xml.em
+++ b/job_templates/ci_launcher_job.xml.em
@@ -45,6 +45,16 @@
         <hudson.plugins.parameterizedtrigger.BuildTriggerConfig>
           <configs>
             <hudson.plugins.parameterizedtrigger.CurrentBuildParameters/>
+@[if os_name == 'linux-aarch64']@
+            <hudson.plugins.parameterizedtrigger.BooleanParameters>
+              <configs>
+                <hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
+                  <name>CI_USE_CONNEXT</name>
+                  <value>false</value>
+                </hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
+              </configs>
+            </hudson.plugins.parameterizedtrigger.BooleanParameters>
+@[end if]
           </configs>
           <projects>@(os_data['job_name'])</projects>
           <condition>SUCCESS</condition>
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -1,4 +1,4 @@
-FROM aarch64/ubuntu:xenial
+FROM ubuntu:xenial
 MAINTAINER William Woodall <william@osrfoundation.org>
 ARG BRIDGE=false
 ARG INSTALL_TURTLEBOT2_DEMO_DEPS=false
@@ -63,12 +63,12 @@ RUN curl --silent http://s3.amazonaws.com/RTI/Bundles/5.2.3/Evaluation/rti_conne
 ADD rti_web_binaries_install_script.py /tmp/rti_web_binaries_install_script.py
 
 # Add the RTI license file.
-#ADD rticonnextdds-license/rti_license.dat /tmp/rti_license.dat
+ADD rticonnextdds-license/rti_license.dat /tmp/rti_license.dat
 
 # Add the RTI binaries we made.
-#ADD rticonnextdds-src/librticonnextdds52_5.2.3-1_amd64.deb /tmp/librticonnextdds52_5.2.3-1_amd64.deb
-#ADD rticonnextdds-src/librticonnextdds52-dev_5.2.3-1_amd64.deb /tmp/librticonnextdds52-dev_5.2.3-1_amd64.deb
-#ADD rticonnextdds-src/rticonnextdds-tools_5.2.3-1_amd64.deb /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
+ADD rticonnextdds-src/librticonnextdds52_5.2.3-1_amd64.deb /tmp/librticonnextdds52_5.2.3-1_amd64.deb
+ADD rticonnextdds-src/librticonnextdds52-dev_5.2.3-1_amd64.deb /tmp/librticonnextdds52-dev_5.2.3-1_amd64.deb
+ADD rticonnextdds-src/rticonnextdds-tools_5.2.3-1_amd64.deb /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
 
 # Install the eProsima dependencies.
 RUN (apt-get update || true) && apt-get install -y libasio-dev libboost-chrono-dev libboost-date-time-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libtinyxml2-dev valgrind
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -8,18 +8,18 @@ ARG PLATFORM=x86
 # See: http://askubuntu.com/questions/506158/unable-to-initialize-frontend-dialog-when-using-ssh
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN (apt-get update || true) && apt-get install -y locales
+RUN apt-get update && apt-get install -y locales
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 
-RUN (apt-get update || true) && apt-get install -y lsb-release
-RUN (apt-get update || true) && apt-get install -y sudo
+RUN apt-get update && apt-get install -y lsb-release
+RUN apt-get update && apt-get install -y sudo
 
 # Get curl for fetching the repo keys.
-RUN (apt-get update || true) && apt-get install -y curl
+RUN apt-get update && apt-get install -y curl
 
 # Get https transport for APT.
-RUN (apt-get update || true) && apt-get install -y apt-transport-https
+RUN apt-get update && apt-get install -y apt-transport-https
 
 # Add the ROS repositories to the apt sources list.
 RUN echo "deb http://repositories.ros.org/ubuntu/testing/ `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list
@@ -35,25 +35,25 @@ RUN if test ${PLATFORM} = x86; then echo "deb https://packagecloud.io/github/git
 RUN if test ${PLATFORM} = x86; then curl --silent https://packagecloud.io/gpg.key | apt-key add -; fi
 
 # Install some development tools.
-RUN (apt-get update || true) && apt-get install -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
+RUN apt-get update && apt-get install -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
-RUN (apt-get update || true) && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-flake8 python3-mock python3-nose python3-pep8 uncrustify
+RUN apt-get update && apt-get install -y clang-format-3.8 cppcheck git pydocstyle pyflakes python3-coverage python3-flake8 python3-mock python3-nose python3-pep8 uncrustify
 
 # Install and self update pip/setuptools to the latest version.
-RUN (apt-get update || true) && apt-get install -y python3-pip
+RUN apt-get update && apt-get install -y python3-pip
 RUN pip3 install -U setuptools pip virtualenv
 
 # Install coverage build dependencies.
-RUN (apt-get update || true) && apt-get install -y gcovr
+RUN apt-get update && apt-get install -y gcovr
 
 # Install the OpenSplice binary from the OSRF repositories.
-RUN if test ${PLATFORM} = x86; then (apt-get update || true) && apt-get install -y libopensplice64; fi
+RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install -y libopensplice64; fi
 # Update default domain id.
 RUN if test ${PLATFORM} = x86; then sed -i "s/<Id>0<\/Id>/<Id>108<\/Id>/" /usr/etc/opensplice/config/ospl.xml; fi
 
 # Install the RTI dependencies.
-RUN if test ${PLATFORM} = x86; then (apt-get update || true) && apt-get install -y default-jre-headless; fi
+RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install -y default-jre-headless; fi
 
 # Install dependencies for RTI web binaries install script.
 RUN pip3 install pexpect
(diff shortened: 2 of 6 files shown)


gerkey:
These builds were initiated by `ci_launcher` using this branch, and they worked fine:
* http://ci.ros2.org/job/ci_windows/2513/
* http://ci.ros2.org/job/ci_osx/1848/
* http://ci.ros2.org/job/ci_linux/2391/

In that run, the aarch64 build failed because of a bug that I fixed; this subsequent is in flight with that fix:
* http://ci.ros2.org/job/ci_linux-aarch64/31/

For completeness, here's another `ci_linux` job running with the latest fixes (which I don't expect to change the outcome):
* http://ci.ros2.org/job/ci_linux/2394/

The new aarch64 packaging job failed for no apparent reason:
* http://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/1/console
There's a new one in the queue.

linux_docker_resources/Dockerfile
```diff
@@ -8,17 +8,18 @@ ARG PLATFORM=x86
 # See: http://askubuntu.com/questions/506158/unable-to-initialize-frontend-dialog-when-using-ssh
 ENV DEBIAN_FRONTEND noninteractive
 
+RUN (apt-get update || true) && apt-get install -y locales
```
wjwwood: Yeah, that `apt-get update || true` is not ideal since it also hides other kinds of update failures too. I have the suspicion that this will come back to bite us, but if you don't see a straightforward way to avoid the docker hub issue, which you described in a comment, then this the best solution.

wjwwood: Also if we have to do this, can we do it in a sed line like above for the `FROM ` so that it doesn't affect the normal linux jobs.

linux_docker_resources/entry_point.sh
```diff
@@ -12,41 +12,49 @@ export ORIG_GID=$(echo $ORIGPASSWD | cut -f4 -d:)
...
 
-echo "Initializing Git-LFS..."
-if ! sudo -H -u rosbuild -- git lfs install; then
-  echo "Git-LFS failed to initialize (this is expected on ARM platforms)."
+if [ $ARCH = "x86_64" ]; then
+  echo "Initializing Git-LFS..."
+  sudo -H -u rosbuild -- git lfs install
+  echo "done."
+else
+  echo "Not installing git-lfs because we're on a non-x86_64 machine."
```
wjwwood: Is it not available for aarch64? I think our use of lfs is not contingent on the architecture.

linux_docker_resources/entry_point.sh
```diff
@@ -12,41 +12,49 @@ export ORIG_GID=$(echo $ORIGPASSWD | cut -f4 -d:)
...
+            echo "Connext not installed correctly." >&2
+            exit 1
+          fi
+          mv /tmp/rti_license.dat /home/rosbuild/rti_license.dat
+          export RTI_LICENSE_FILE=/home/rosbuild/rti_license.dat
+          ;;
+      esac
+      echo "done."
+    else
+      echo "Not installing Connext because we're on a non-x86_64 machine."
```
wjwwood: I think this should maybe fail right here, because if `--connext` is passed as an option in `CI_ARGS`, then this job will just fail later because it cannot find Connext, right?

linux_docker_resources/entry_point.sh
gerkey: Apparently it's not available:
~~~
root@ros2-testing:~/code/ci# apt-get install -y git-lfs
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package git-lfs
~~~
I'm not sure why it's not available, but this isn't a change; we were previously not installing `git-lfs` on arm for the qemu builds.

wjwwood: Yeah, I think it's fine because we're not actually using git-lfs anywhere atm. Just curious -- it shouldn't block this pr.

dhood:
You'll need to update the packaging template as well if you're going to run packaging jobs for arm: https://github.com/ros2/ci/blob/59575a285a653fb3e0ab34876dec61737bc596b4/job_templates/packaging_job.xml.em#L111

Also keep in mind that as it stands, packaging jobs will try to build the bridge for the new job [because of this line](https://github.com/ros2/ci/blob/59575a285a653fb3e0ab34876dec61737bc596b4/ros2_batch_job/__main__.py#L310) that equates linux* to linux.

linux_docker_resources/entry_point.sh
gerkey: I thought about that, but I want to chain into this job from the standard `ci_launcher` job, without special-casing the params for this architecture. I don't think that there will be a later failure because when we can't find Connext we just won't build that RMW flavor.

It's potentially misleading in that the aarch64 `ci` job is testing a smaller subset of functionality than are the other `ci` jobs, but I don't think that that's a big deal.

linux_docker_resources/Dockerfile
wjwwood: Specifically I'm imagining someone adding a new PPA to the normal Linux machines which then cause apt-get update to fail, but the only thing they see is that their package they expected to find in the PPA isn't installable. They would have seen the issue with apt-get update, but it was being suppressed by the `|| true`.

mikaelarguedas:
> Also keep in mind that as it stands, packaging jobs will try to build the bridge for the new job.

Is this something that we don't want to do ?

linux_docker_resources/entry_point.sh
wjwwood: Yeah, but if I explicitly ask for connext and my build passes then I'll assume it worked, which is not great, even when starting a single `ci_linux_aarch64` and not using the launcher. I understand the difficulty of doing this better given how the launcher works, I was just pointing out the issue.

linux_docker_resources/Dockerfile
gerkey: Capital idea on using sed instead: a2db644

gerkey:
@dhood, aha, that explains why the packaging job fails, thanks! I'll look into the template.

gerkey:
I think that we should try to build the bridge on ARM. I don't see any reason in principle why it shouldn't work.

gerkey:
Here's a packaging job to test both the `sed` changes to the `Dockerfile` and all the updates that I made to support packaging on arm:

http://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/2

linux_docker_resources/entry_point.sh
gerkey: Fair point. And, while it doesn't produce an error, we do get CMake warnings that Jenkins dutifully reports: http://ci.ros2.org/job/ci_linux-aarch64/31/warnings2Result/.

I'll look into making the launcher smarter.

gerkey: 4047ddfbd89653366cb79ec7756557034903fb6e reverts my change to hide the failure to install Connext, replacing it with a more informative error message.

f91bbce19ab26d2cbfff4fd7f41b47831dbcb73c attempts to disable Connext specifically for the aarch64 job when started by the launcher. It's not clear to me how the `Parameterized Trigger Plugin` will merge the current build's params with one that's explicitly set; I have jobs in flight to figure that out.

gerkey:
New builds to test the launcher's ability to selectively disable Connext on aarch64:
* http://ci.ros2.org/job/ci_linux/2396/
* http://ci.ros2.org/job/ci_osx/1850/
* http://ci.ros2.org/job/ci_windows/2515/
* http://ci.ros2.org/job/ci_linux-aarch64/33/ (in the queue)

According to Jenkins, they are configured as desired: the first three have `CI_USE_CONNEXT` true and aarch64 has it false.

linux_docker_resources/entry_point.sh
wjwwood: Sounds good, thanks for looking into it.

gerkey:
The aarch64 packaging job is happy and the `CI_USE_CONNEXT` param was correctly propagated to all the launcher-started jobs.

I'm going to merge this now. Note that we can expect a test failure or two, at least until https://github.com/ros2/rosidl/pull/211 is resolved.

dirk-thomas:
The name of the generated artifact of the packaging job should include the `-aarch64` suffix to distinguish it from the "normal" linux package.

gerkey:
#49 modifies the names of the generated build artifacts.

job_templates/ci_launcher_job.xml.em
```diff
@@ -45,6 +45,16 @@
...
+@[if os_name == 'linux-aarch64']@
+            <hudson.plugins.parameterizedtrigger.BooleanParameters>
+              <configs>
+                <hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
+                  <name>CI_USE_CONNEXT</name>
+                  <value>false</value>
+                </hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
+              </configs>
+            </hudson.plugins.parameterizedtrigger.BooleanParameters>
+@[end if]
```
dirk-thomas: This lacks a trailing `@`. Addressed in #50.

gerkey: Thanks for fixing that!

PR #48, opened 2017-04-05
remove git lfs

dirk-thomas:
Removing for now since it is not being used (see ros2/ros2#173).


dirk-thomas:
[Build Status](http://ci.ros2.org/job/ci_linux/2403/)

PR #49, opened 2017-04-05
add arch to package name

gerkey:


gerkey:
http://ci.ros2.org/view/packaging/job/packaging_linux/637/
http://ci.ros2.org/view/packaging/job/packaging_osx/565/
http://ci.ros2.org/view/packaging/job/packaging_windows/558/
http://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/4/

dirk-thomas:
I am not sure if it is a good idea to rename the artifacts of all previous platforms too. That requires all the instructions using those names to be updated too.

tfoote:
@dirk-thomas can you clarify what you mean by "all previous platforms"?

mikaelarguedas:
I think it means that every artifact produced by the packaging jobs (not aarch64) will be now have "x86_64" appended to their name so every instruction referring to these artifacts should be updated accordingly.

mikaelarguedas:
To clarify I'm not against renaming them all just giving my understanding.
IMO there are a handful of places (if any) where we refer to thos by name so it shouldn't be hard to update them all

gerkey:
I thought about only naming the aarch64 artifact but decided that it's more consistent and future-proof to include the architecture for all of them (what if we start building for Windows on ARM?).

I'm happy to update the instructions. Looks like it's just this handful of mentions:
~~~
ros2.wiki$ grep -- -package- *
Linux-Install-Binary.md:* Download the latest package for Linux; let's assume that it ends up at `~/Downloads/ros2-package-linux.tar.bz2`.
Linux-Install-Binary.md:        tar xf ~/Downloads/ros2-package-linux.tar.bz2
OSX-Install-Binary.md:- Download the latest package for OS X; let's assume that it ends up at `~/Downloads/ros2-package-osx.tar.bz2`.
OSX-Install-Binary.md:        tar xf ~/Downloads/ros2-package-osx.tar.bz2
Release-Howto.md:- Rename each artifact file (an archive file) as `ros2-alpha<alpha number>-package-<platform>.<ext>`
Release-Howto.md:  - E.g. `ros2-alpha4-package-linux.tar.bz2`
Windows-Install-Binary.md:* Download the latest package for Windows, e.g., `ros2-package-windows.zip`.
~~~

gerkey:
I updated the docs:

https://github.com/ros2/ros2/wiki/_compare/4a6dfb3ef6fe0447185da37c042ea86559663a21...50ad7118d95633af69e7c2bbc1152e0144c6ec7a

Note that the instructions already were flexible on file naming, saying that the package names given there are examples and that the actual names might be different.

PR #52, opened 2017-04-16
remove deprecated MAINTAINER tag

dirk-thomas:
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


mikaelarguedas:
@dirk-thomas is this branch still needed?

dirk-thomas:
No, thanks, deleted (I always miss the button in the app...).

PR #53, opened 2017-04-17
Switch the rebase to a merge.

clalancette:
As we have discussed, using a rebase will not work when you are doing a merge-based workflow.  We still want to always test the changes against the latest, though, so switch the rebase to a merge, and add some additional comments about why we are doing this.

connects to ros2/ros2#342

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -426,9 +426,11 @@ def run(args, build_function, blacklisted_package_names=None):
                 info("'{0}' returned exit code '{1}'", fargs=(" ".join(vcs_custom_cmd), ret))
                 print()
 
-                # Attempt to rebase all the repositories to the __ci_default branch
-                info("Attempting to rebase all repositories to the '__ci_default' branch")
-                vcs_custom_cmd = vcs_cmd + ['custom', '.', '--git', '--args', 'rebase', '__ci_default']
+                # Attempt to merge all the repositories to the __ci_default branch.  This is to
+                # ensure that the changes on the branch still work when applied to the
+                # latest version of the default branch.
+                info("Attempting to merge all repositories to the '__ci_default' branch")
+                vcs_custom_cmd = vcs_cmd + ['custom', '.', '--git', '--args', 'merge', '__ci_default']
                 ret = job.run(vcs_custom_cmd)
                 info("'{0}' returned exit code '{1}'", fargs=(" ".join(vcs_custom_cmd), ret))
                 print()


nuclearsandwich:
Is your plan for merge conflicts to let the build fail as it does with a failed rebase now?

clalancette:
Yes, exactly.  I think at that point, it is reasonable to require the CI user to have to do conflict resolution to get their build working.  Do you agree?

nuclearsandwich:
> it is reasonable to require the CI user to have to do conflict resolution to get their build working. Do you agree?

Yep. It's just a bummer that it costs us a build which may or may not be followed closely by the CI user to learn that. It doesn't fit with the workflow, in the past I've seen a pre-build step to merge the default branch into the PR automatically before running CI, which meant we caught the merge conflict early and could notify the CI user via @-mention in the PR comments.

Since this doesn't really change the current fail-on-conflict behavior, I don't think it's something that needs to be addressed now but it's definitely on my wishlist to improve both the UX and branch integration.

clalancette:
Yeah, that's a nice system.  I am just not that ambitious right now :).

clalancette:
The CI here is just to show what happens in the case where there is a merge commit, and in the case where there is a fast-forward.  

linux with merge commit: [Build Status](http://ci.ros2.org/job/ci_linux/2460/)
linux with fast-forward: [Build Status](http://ci.ros2.org/job/ci_linux/2473/)

clalancette:
@ros2/team Thoughts here?  I'll need this (or something like this) to merge up the tf2 changes from indigo-devel.

clalancette:
Thanks @wjwwood .  I'm going to merge this up now; @dhood please let me know ASAP if you see anything wonky happening on the build farm because of this.

ros2_batch_job/__main__.py
```diff
@@ -426,9 +426,11 @@ def run(args, build_function, blacklisted_package_names=None):
                 info("'{0}' returned exit code '{1}'", fargs=(" ".join(vcs_custom_cmd), ret))
                 print()
 
-                # Attempt to rebase all the repositories to the __ci_default branch
-                info("Attempting to rebase all repositories to the '__ci_default' branch")
-                vcs_custom_cmd = vcs_cmd + ['custom', '.', '--git', '--args', 'rebase', '__ci_default']
+                # Attempt to merge all the repositories to the __ci_default branch.  This is to
+                # ensure that the changes on the branch still work when applied to the
+                # latest version of the default branch.
```
dirk-thomas: I understand the comment that the custom branch is being merged into the default branch. Isn't the code doing the opposite - the default branch being merged into the custom branch?

clalancette: @dirk-thomas You are absolutely right about that.  As far as the code goes, it doesn't matter; either way will produce the same result.  I'll fix up the comment so that it says the right thing.  Thanks.

PR #60, opened 2017-04-25
Fix and update osrf pycommon subtree

wjwwood:
The subtree for `osrf_pycommon` broken when we moved the parent folder:

https://github.com/ros2/ci/commit/ed5672bf72b4065b9b646d4f7c3866333fab08b8#diff-7022afb2ea0a34e81b004ae7b1458535

I also updated the subtree to the latest upstream version. This should be safe since that's what we're using for everything else in the ROS 2 source tree (this copy in "vendor" is only used to bootstrap the build tool).

I'm going to run CI anyways.


wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2528/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/132/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1990/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2634/)

Issue #69, opened 2017-05-02
Add/change turtlebot demo to build on aarch64


clalancette:
The current turtlebot_demo CI job runs on labels marked as "linux", which in our build farm is implicitly x86_64 machines. However, the real turtlebot demo is almost certainly going to run on aarch64, so we should either change the label on the turtlebot-demo job to be "linux_aarch64", or we should add an additional job to build turtlebot-demo on both architectures. @nuclearsandwich @gerkey does that make sense?

gerkey:
I like the idea of adding a second job labeled `linux_aarch64`. Then we can rest assured that it builds on both architectures.

mikaelarguedas:
:+1: for creating another job "ci_turtlebot-demo-aarch64" rather than modifying the label of the existing one

gerkey:
I'll do this.

gerkey:
Fixed by #70.

PR #70, opened 2017-05-05
add aarch64 flavor of tbot demo job

gerkey:
Addresses #69.

These changes have been deployed and for clarity I deleted the old tbot job (the two new ones have the os/arch in the name).

CI after deployment:
- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/1/)
- Linux aarch64:
    - [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/2/)

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -192,7 +192,11 @@ echo "# BEGIN SECTION: Build Dockerfile"
 @[if os_name == 'linux-aarch64']@
 sed -i 's+^FROM.*$+FROM aarch64/ubuntu:xenial+' linux_docker_resources/Dockerfile
 sed -i 's+apt-get update+(apt-get update || true)+' linux_docker_resources/Dockerfile
+@[  if turtlebot_demo]@
+docker build --build-arg PLATFORM=arm -t ros2_batch_ci_turtlebot_demo linux_docker_resources
+@[  else]@
 docker build --build-arg PLATFORM=arm -t ros2_batch_ci_aarch64 linux_docker_resources
+@[  end if]@
 @[elif os_name == 'linux']@
 @[  if turtlebot_demo]@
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Renaming the existing job would preserve the history. I think it would be valuable to do that in order to be able to identify changes over time.

gerkey:
Ah, I didn't realize that you could rename an existing job. Well, I already deleted the old job, which I assume deleted the history?

gerkey:
New CI after setting the flag to install deps on aarch64
- Linux aarch64:
    - [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/4/)

dirk-thomas:
I am not sure if Jenkins deletes all artifacts with the job. You can check the filesystem on the master to see if there is anything left.

job_templates/ci_job.xml.em
```diff
@@ -192,7 +192,11 @@ echo "# BEGIN SECTION: Build Dockerfile"
 @[if os_name == 'linux-aarch64']@
 sed -i 's+^FROM.*$+FROM aarch64/ubuntu:xenial+' linux_docker_resources/Dockerfile
 sed -i 's+apt-get update+(apt-get update || true)+' linux_docker_resources/Dockerfile
+@[  if turtlebot_demo]@
+docker build --build-arg PLATFORM=arm --build-arg INSTALL_TURTLEBOT2_DEMO_DEPS=true -t ros2_batch_ci_turtlebot_demo linux_docker_resources
```
nuclearsandwich: If this gets any more options it might be worth breaking out a function that generates the correct `docker build` invocation for each combo. I don't think it needs to be refactored yet though.

gerkey: Agreed; let's revisit that the next time we add Docker build args.

gerkey:
The latest aarch64 build succeeded, so I think that we're good to go here.

PR #71, opened 2017-05-08
Add in dependencies for turtlebot demo

clalancette:
In order to successfully build the turtlebot demo with cartographer, we need to add a number of turtlebot-specific dependencies.  Along with this, we also need to add in a couple of workarounds for bugs in the packaging of libpcl in Ubuntu 16.04.  With all of this in place, we are able to successfully build and link all of the pieces necessary for cartographer on the turtlebot.

connects to ros2/ros2#342

CI:
linux: [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/8/)

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -86,7 +86,7 @@ RUN echo "@today_str"
 RUN if test ${BRIDGE} = true; then apt-get update && apt-get install --no-install-recommends -y python-rospkg ros-kinetic-catkin ros-kinetic-common-msgs ros-kinetic-rosbash ros-kinetic-roscpp ros-kinetic-roslaunch ros-kinetic-rosmsg ros-kinetic-roscpp-tutorials ros-kinetic-rospy-tutorials ros-kinetic-tf2-msgs; fi
 
 # Install build dependencies for turtlebot demo.
-RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install -y --no-install-recommends libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libgoogle-glog-dev libpcl-dev libsdl1.2-dev libsdl-image1.2-dev libsuitesparse-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
+RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install -y --no-install-recommends libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libgoogle-glog-dev liblua5.2-dev libpcl-dev libsdl1.2-dev libsdl-image1.2-dev libsuitesparse-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
 
 # Workaround a bug in the Ubuntu 16.04 packaging of libpcl.
 RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then sed -i -e '/vtkproj4;/d' /usr/lib/$(uname -m)-linux-gnu/cmake/pcl/PCLConfig.cmake; fi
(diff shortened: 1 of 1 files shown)


linux_docker_resources/Dockerfile
```diff
@@ -86,7 +86,11 @@ RUN echo "@today_str"
...
 
 # Install build dependencies for turtlebot demo.
-RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install --no-install-recommends -y libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libpcl-dev libsdl1.2-dev libsdl-image1.2-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
+RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install -y --no-install-recommends google-mock libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libgoogle-glog-dev liblua5.2-dev libpcl-dev libprotobuf-dev libsdl1.2-dev libsdl-image1.2-dev libsuitesparse-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev protobuf-compiler python-sphinx ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
+
+# Workaround bugs in the Ubuntu 16.04 packaging of libpcl.
```
nuclearsandwich: Are the two lines below working around one bug or two separate ones?
If both lines are needed to work around the same bug, we should make them part of the same RUN directive and escape the newlines with `\` to keep a sensible line length.

If there's an upstream bug can you link it in the comments. Doing so will help us track when the fix may no longer be needed.

linux_docker_resources/Dockerfile
```diff
@@ -86,7 +86,11 @@ RUN echo "@today_str"
 RUN if test ${BRIDGE} = true; then apt-get update && apt-get install --no-install-recommends -y python-rospkg ros-kinetic-catkin ros-kinetic-common-msgs ros-kinetic-rosbash ros-kinetic-roscpp ros-kinetic-roslaunch ros-kinetic-rosmsg ros-kinetic-roscpp-tutorials ros-kinetic-rospy-tutorials ros-kinetic-tf2-msgs; fi
 
 # Install build dependencies for turtlebot demo.
-RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install --no-install-recommends -y libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libpcl-dev libsdl1.2-dev libsdl-image1.2-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
+RUN if test ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true; then apt-get update && apt-get install -y --no-install-recommends google-mock libboost-iostreams-dev libboost-regex-dev libboost-system-dev libboost-thread-dev libgoogle-glog-dev liblua5.2-dev libpcl-dev libprotobuf-dev libsdl1.2-dev libsdl-image1.2-dev libsuitesparse-dev libudev-dev libusb-1.0.0-dev libyaml-cpp-dev protobuf-compiler python-sphinx ros-kinetic-kobuki-driver ros-kinetic-kobuki-ftdi; fi
```
dirk-thomas: Please avoid reordering `-y` and `--no-install-recommends` compared to other lines.

clalancette: Oops, yeah, that must have happened while cutting-n-pasting.  I'll fix it.

linux_docker_resources/Dockerfile
clalancette: So, this appears to be one bug that manifests in two separate ways.  I'll put a comment above it with a link to the original bug report.  Note that the original bug has already been fixed in Debian, but I don't think the fix has made its way into Ubuntu yet.

nuclearsandwich:
Looks good to me with the latest changes and assuming that you're satisfied with the status of the build. i.e. you'd know better than me whether those failures are part of the work in progress or if they're newly caused by this change.

clalancette:
Most of those failures are because of an expired RTI license.  However, there are 4 new failures in there that I don't understand, so I'm going to hold off merging this until I do.

clalancette:
So with the latest changes, I now have a build that is almost warning-free (the last warning is in the turtlebot2_drivers, which I'll fix presently).  Given that, and that this is approved, I'm going to merge this one up.  Thanks!

PR #76, opened 2017-05-27
Installing + finding openssl

mikaelarguedas:
Opening this for visibility, will put it in review with the other sros2 PRs

connects to ros2/ros2#339

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -77,7 +77,7 @@ def main(argv=None):
         'use_osrf_connext_debs_default': 'false',
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
-        'ament_build_args_default': '--parallel',
+        'ament_build_args_default': '--parallel --cmake-args -DSECURITY=ON--',
         'ament_test_args_default': '--retest-until-pass 10',
         'enable_c_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -77,7 +77,7 @@ def main(argv=None):
         'use_osrf_connext_debs_default': 'false',
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
-        'ament_build_args_default': '--parallel --cmake-args -DSECURITY=ON--',
+        'ament_build_args_default': '--parallel --cmake-args -DSECURITY=ON --',
         'ament_test_args_default': '--retest-until-pass 10',
         'enable_c_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
--- a/ros2_batch_job/osx_batch/__init__.py
+++ b/ros2_batch_job/osx_batch/__init__.py
@@ -41,7 +41,7 @@ class OSXBatchJob(BatchJob):
         if 'OPENSSL_ROOT_DIR' not in os.environ:
             import subprocess
             brew_openssl_prefix = subprocess.run(
-                ["brew", "--prefix", "openssl"], shell=True, stdout=subprocess.PIPE)
+                ['brew --prefix openssl'], shell=True, stdout=subprocess.PIPE)
             if not brew_openssl_prefix.stdout is bytes():
               os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix.stdout.decode().strip('\n')
 
--- a/ros2_batch_job/osx_batch/__init__.py
+++ b/ros2_batch_job/osx_batch/__init__.py
@@ -41,8 +41,9 @@ class OSXBatchJob(BatchJob):
         if 'OPENSSL_ROOT_DIR' not in os.environ:
             import subprocess
             brew_openssl_prefix = subprocess.run(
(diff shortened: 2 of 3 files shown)


ros2_batch_job/osx_batch/__init__.py
```diff
@@ -38,6 +38,8 @@ def pre(self):
             os.environ['LANG'] = 'en_US.UTF-8'
         if 'ROS_DOMAIN_ID' not in os.environ:
             os.environ['ROS_DOMAIN_ID'] = '111'
+        if 'OPENSSL_ROOT_DIR' not in os.environ:
+            os.environ['OPENSSL_ROOT_DIR'] = '/usr/local/Cellar/openssl/1.0.2k'
```
mikaelarguedas: We should be able to move this to the bash_profile (of the henkins user) on each MacOS node and not have to put it in the ci script here

dirk-thomas: This looks fragile. Is there any way to retrieve that path so that it doesn't embed the exact version here?

mikaelarguedas: as mentioned in https://github.com/ros2/ci/pull/76#discussion_r121731426, I'd expect that to be done directly in the "bashrc" of the build machines rather than in this script. Didn't have time to do / test it though.

dirk-thomas: How about looking in `/usr/local/Cellar/openssl`, sort all subfolders, and pick the latest one?

If not I think at least a TODO should be added.

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -38,6 +38,12 @@ def pre(self):
             os.environ['LANG'] = 'en_US.UTF-8'
         if 'ROS_DOMAIN_ID' not in os.environ:
             os.environ['ROS_DOMAIN_ID'] = '111'
+        if 'OPENSSL_ROOT_DIR' not in os.environ:
+            import subprocess
+            brew_openssl_prefix = subprocess.run(
+                ['brew --prefix openssl'], shell=True, stdout=subprocess.PIPE)
```
dirk-thomas: Is `shell` really necessary here?

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -38,6 +38,12 @@ def pre(self):
...
+            if not brew_openssl_prefix.stdout is bytes():
+              os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix.stdout.decode().strip('\n')
```
dirk-thomas: The environment variable is only assigned conditionally. What if the condition used for? What is supposed to happen when it is `False`?

create_jenkins_job.py
```diff
@@ -77,7 +77,7 @@ def main(argv=None):
         'use_osrf_connext_debs_default': 'false',
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
-        'ament_build_args_default': '--parallel',
+        'ament_build_args_default': '--parallel --cmake-args -DSECURITY=ON--',
```
dirk-thomas: Please insert a space before the trailing `--`.

ros2_batch_job/osx_batch/__init__.py
mikaelarguedas: Sorry I'm is still experimenting. I think that it depends what we build in the end,  if none of the packages we build use openssl, the build should not fail even with that env variable empty.
So I think that defaulting to not setting it if openssl config is not found is an ok behaviour. I can add a warning or something though to let the dev know that it was not bound.

dirk-thomas: The condition just looks weird to me. I would expect something like `rc != 0` or `not output` instead.

mikaelarguedas: I tried to address this in https://github.com/ros2/ci/pull/76/commits/6eadb5f204e1047612a4a22b199c624c890e1e9b. How does that look ?

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -38,6 +38,13 @@ def pre(self):
             os.environ['LANG'] = 'en_US.UTF-8'
         if 'ROS_DOMAIN_ID' not in os.environ:
             os.environ['ROS_DOMAIN_ID'] = '111'
+        if 'OPENSSL_ROOT_DIR' not in os.environ:
+            import subprocess
```
dirk-thomas: The import should be at the top of the file. Only modules which can't be loaded globally or are optional should be imported in a local scope.

ros2_batch_job/osx_batch/__init__.py
dirk-thomas: :+1:

ros2_batch_job/osx_batch/__init__.py
mikaelarguedas: My bad I thought it was narrowest scope, fixed in https://github.com/ros2/ci/pull/76/commits/01ec8ee5ee202ff9c7fb7e4fa3547ff653e5c628

Issue #75, opened 2017-05-16
Investigate limited number of layers on aarch64 docker


clalancette:
See ros2/ci#73 for a description of the problem, and a link to problematic builds.  We've currently worked around the problem with ros2/ci#74, but reverting that commit should show the problem again for debugging.

nuclearsandwich:
This is a sort of time bomb / hot potato for whichever buildfarmer's touch to the Dockerfile re-triggers this issue.

https://github.com/moby/moby/issues/1171 looks very promising. I might get a chance to circle back to this during my buildfarmercop reign.

vielmetti:
https://github.com/moby/moby/issues/27384 is perhaps the root cause of this. Docker 1.12.x on ARM64 is compiled with Go 1.7, and Go has a bug that's tickled by Docker.

The good news is that we're looking at Go 1.9 successfully being available as a binary release for ARM64 from the Go project, and in some reasonable timeframe after that there should be a fresh binary for ARM64 of Docker that can be easily consumed.

In the meantime the best strategy is to reduce the number of layers you use. There are various Docker optimization strategies to pursue that would reduce the size of the Dockerfile and thus eliminate layers and build steps, with a performance improvement. Looking at

https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile

In particular successive RUN commands can usually be merged, e.g. as

RUN foo
RUN bar

is almost always equivalent to your purposes to

RUN foo && bar

and that wipes out a layer. You might look at this (vintage) writeup

https://blog.tutum.co/2014/10/22/how-to-optimize-your-dockerfile/

that's pretty useful. Newest Docker has some more tricks up its sleeve, so don't go completely overboard on optimizing, but there are small changes that should help stability.

vielmetti:
My understanding is that this has been resolved by upgrading the version of Docker on the arm64 machines to one that has better file system stability.

nuclearsandwich:
> My understanding is that this has been resolved by upgrading the version of Docker on the arm64 machines to one that has better file system stability.

We haven't upgraded this Jenkins instance yet. We have what was previously referred to as the ROS 2 buildfarm which lives at http://ci.ros2.org and provides continuous integration and build archiving for ROS 2 omnibus builds, and in the past couple weeks have created a ROS 2 buildfarm at http://build.ros2.org based more closely on the ROS buildfarm (build.ros.org) to build the Xenial debs for the beta2 release of ROS 2.

[...]

PR #74, opened 2017-05-16
Move the two separate RUN commands into one statement.

clalancette:
This avoids a bug (only on aarch64) where a certain combinations
of RUN and ADD commands causes Docker to fail to build the
image.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2601/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/182/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2059/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2718/)
* TB aarch64: [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/27/)
* TB amd64: [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/13/)

PR #79, opened 2017-05-31
install net-tools for ifconfig in docker image

wjwwood:
fixes https://github.com/ros2/build_cop/issues/26

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -11,10 +11,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y locales
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 
-RUN apt-get update && apt-get install --no-install-recommends -y lsb-release
-RUN apt-get update && apt-get install --no-install-recommends -y sudo
-# For ifconfig
-RUN apt-get update && apt-get install --no-install-recommends -y net-tools
+# net-tools is for ifconfig
+RUN apt-get update && apt-get install --no-install-recommends -y lsb-release net-tools sudo
 
 # Get curl for fetching the repo keys.
 RUN apt-get update && apt-get install --no-install-recommends -y curl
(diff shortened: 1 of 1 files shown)


dhood:
I think it might need CI for ARM given https://github.com/ros2/ci/issues/75

mikaelarguedas:
> I think it might need CI for ARM given #75

Yeah, this is in the middle of a series of very similar commands without any add commande before so it would be very surprising. I'll leave it up to @wjwwood to run CI or not

wjwwood:
I started a job just to make sure if got rid of the ifconfig missing issue:

http://ci.ros2.org/job/ci_linux/2651/console

It looks good, but I'll also run an ARM job as @dhood suggested.

dhood:
Now that I remember what originally caused that issue to appear, I also launched [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/33/) because it was the extra lines that are only in the turtlebot dockerfile that caused it to show up on ARM (https://github.com/ros2/ci/pull/74)

wjwwood:
Thanks @dhood the ARM job passed docker and the issue was fixed:

http://ci.ros2.org/job/ci_linux-aarch64/206/console

I'll wait for the job you launched to get past docker before merging.

wjwwood:
@dhood did it fail in the same way?

dhood:
yes, unfortunately. there's an issue linked to [in this comment](https://github.com/ros2/ci/issues/75#issuecomment-302186809) that might have the permanent fix, but the temporary workaround is to merge lines in the dockerfile (we're hitting some layer limit)

wjwwood:
I'm going to try again with ab4085fb4e20597c62bf82843eb3fc266a45bb6c

mikaelarguedas:
My bad, I thought it was related to the type of command not the number of layers. Thanks @dhood  for checking

mikaelarguedas:
Maybe we should rename https://github.com/ros2/ci/issues/75 accordingly now that we know more about the issue

wjwwood:
http://ci.ros2.org/job/ci_linux-aarch64/207/console is working, so I'll merge this.

PR #84, opened 2017-06-19
add arguments for build args to packaging job

mikaelarguedas:
Haven't tested this yet.

Should allow to pass default build args to packaging job the same way its done for other ci jobs

### DIFF
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -23,6 +23,7 @@
     default_repos_url=default_repos_url,
     supplemental_repos_url=supplemental_repos_url,
     cmake_build_type=cmake_build_type,
+    ament_build_args_default=ament_build_args_default,
 ))@
         <hudson.model.ChoiceParameterDefinition>
           <name>CI_USED_RMW_IMPL</name>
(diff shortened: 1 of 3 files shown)


job_templates/packaging_job.xml.em
```diff
@@ -35,6 +34,11 @@
             </a>
           </choices>
         </hudson.model.ChoiceParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_AMENT_BUILD_ARGS</name>
+          <description>Additional arguments passed to 'ament build'.</description>
+          <defaultValue>@(ament_build_args_default)</defaultValue>
+        </hudson.model.StringParameterDefinition>
```
mikaelarguedas: So here we can either move the CI_AMENT_BUILD_ARGS parameter inside our [common parameter snippet](https://github.com/ros2/ci/blob/master/job_templates/snippet/property_parameter-definition_common.xml.em) (given that after this PR all jobs will use it) or explicitly add it to the packaging job template. I'd prefer adding it to the template to avoid duplication. I was wondering if there is any objection before doing so

dirk-thomas: No objection :wink:

mikaelarguedas: sounds good: https://github.com/ros2/ci/pull/84/commits/5c8a00e3ac4d81b02ac1e3c5258f67fee2219a43
This will result in a diff on all jobs config, and the position of the parameter will change in the UI (the build args will be just under the build type so much higher than before)

mikaelarguedas: example of what the jobs configuration page would look like:
http://ci.ros2.org/job/ci_packaging_linux-aarch64/build?delay=0sec
I havent deployed the regular ci jobs but they would look the same (CI_AMENT_BUILD_ARGS before any parameter that is job specific v.s. at the bottom before). Is it ok for everyone?

dirk-thomas: Fine with me.

mikaelarguedas:
I have 2 approvals on the new layout so I'm going to merge and deploy this now that the master restarted

Issue #87, opened 2017-06-26
Shebang lines not re-written for executables in libexec


dhood:
Our packaging jobs re-write shebang lines to be portable, however it only seems to do so for files installed to 'bin': https://github.com/ros2/ci/blob/af70aa2b8c6682b97fa286e574e33172dcc04980/ros2_batch_job/packaging.py#L118

Since we install executables to "libexec" as of late, this causes the packaged binaries to give this error (OS X):

```
$ ros2 run demo_nodes_py talker
Traceback (most recent call last):
  File "/Users/dhood/Downloads/ros2-osx-june25-2017/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.0.0', 'console_scripts', 'ros2')()
  File "/Users/dhood/Downloads/ros2-osx-june25-2017/lib/python3.6/site-packages/ros2cli/cli.py", line 64, in main
    rc = extension.main(parser=parser, args=args)
  File "/Users/dhood/Downloads/ros2-osx-june25-2017/lib/python3.6/site-packages/ros2run/command/run.py", line 59, in main
    return run_executable(path=path, argv=args.argv)
  File "/Users/dhood/Downloads/ros2-osx-june25-2017/lib/python3.6/site-packages/ros2run/api/__init__.py", line 75, in run_executable
    completed_process = subprocess.run([path] + argv)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 707, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 1326, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/dhood/Downloads/ros2-osx-june25-2017/lib/demo_nodes_py/talker'

$ cat /Users/dhood/Downloads/ros2-osx-june25-2017/lib/demo_nodes_py/talker
#!/Users/osrf/jenkins/workspace/packaging_osx/venv/bin/python
[...]

PR #88, opened 2017-06-26
Updates to packaging jobs for "libexec" binary installation

dhood:
Connects to ros2/ci#87

Nodes are being installed to libexec as of late

ci packaging osx: [Build Status](http://ci.ros2.org/job/ci_packaging_osx/3/)
ci packaging linux: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/26/)
ci packaging windows: [Build Status](http://ci.ros2.org/job/ci_packaging_windows/1/)

http://ci.ros2.org/job/ci_packaging_osx/3/console#console-section-7 shows additional files have shebang modified now. tested the build artifact and confirmed it fixes ros2/ci#87

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -112,19 +112,30 @@ def build_and_test_and_package(args, job):
     # Only on Linux and OSX Python scripts have a shebang line
     if args.os in ['linux', 'osx']:
         print('# BEGIN SUBSECTION: rewrite shebang lines')
-        # Demo nodes are installed to 'lib', so binaries may be present there in addition to 'bin'
-        for dir_with_binaries in ['bin', 'lib']:
-            path_with_binaries = os.path.join(args.installspace, dir_with_binaries)
-            for path in [os.path.join(root, f) for root, dirs, files in os.walk(path_with_binaries) for f in files]:
-                with open(path, 'rb') as h:
-                    content = h.read()
-                shebang = b'#!%b' % job.python.encode()
-                if content[0:len(shebang)] != shebang:
-                    continue
-                print('- %s' % path)
-                with open(path, 'wb') as h:
-                    h.write(b'#!/usr/bin/env python3')
-                    h.write(content[len(shebang):])
+        filenames = []
+
+        bin_path = os.path.join(args.installspace, 'bin')
+        filenames.append([os.path.join(bin_path, filename) for filename in os.listdir(bin_path)])
+
+        # Demo nodes are installed to 'lib/<package_name>'
+        lib_path = os.path.join(args.installspace, 'lib')
+        for lib_sub_dir in next(os.walk(lib_path))[1]:
+            sub_dir_path = os.path.join(lib_path, lib_sub_dir)
+            sub_dir_contents = [os.path.join(sub_dir_path, name) for name in os.listdir(sub_dir_path)]
+            filenames.extend([filename for filename in sub_dir_contents if os.path.isfile(filename)])
+        for lib_sub_dir in next(os.walk(lib_path))[1]:
+            filenames.append([os.path.join(lib_sub_dir, filename) for filename in os.listdir(lib_sub_dir)])
+
+        for path in filenames:
+            with open(path, 'rb') as h:
+                content = h.read()
+            shebang = b'#!%b' % job.python.encode()
+            if content[0:len(shebang)] != shebang:
+                continue
+            print('- %s' % path)
+            with open(path, 'wb') as h:
+                h.write(b'#!/usr/bin/env python3')
+                h.write(content[len(shebang):])
         print('# END SUBSECTION')
 
     print('# BEGIN SUBSECTION: create archive')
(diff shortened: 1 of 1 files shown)


ros2_batch_job/packaging.py
```diff
@@ -112,30 +112,31 @@ def build_and_test_and_package(args, job):
...
-            if content[0:len(shebang)] != shebang:
-                continue
-            print('- %s' % path)
-            with open(path, 'wb') as h:
-                h.write(b'#!/usr/bin/env python3')
-                h.write(content[len(shebang):])
+        # Demo nodes are installed to 'lib', so binaries may be present there in addition to 'bin'
+        for dir_with_binaries in ['bin', 'lib']:
+            path_with_binaries = os.path.join(args.installspace, dir_with_binaries)
+            for path in [os.path.join(root, f) for root, dirs, files in os.walk(path_with_binaries) for f in files]:
```
dirk-thomas: The algorithm should walk recursively. In the case of `bin` it should only consider files in that folder (no recursion). In the `lib` case it should only consider direct subfolder and then the files in that folder (exactly one level of recursion without considering the parent `lib` folder).

dhood: How it is at the moment, the `bin` and `lib` directories are both being walked recursively with no limit. Could you explain the harm of considering the parent `lib` folder and the subdirectories of `bin`? We don't install anything to either at the moment, but if we were to we would also want the shebang to be re-written I would think.

This can be refactored to act differently for `lib` and `bin` but I would like to understand why it's necessary.

dirk-thomas: The algorithm should only try to touch very specific files. While we currently don't have files in the other locations there could be some in the future and they shouldn't be changed.

dhood: that's what I am trying to understand. If in the future we install something else that has `#!/Users/osrf/jenkins/workspace/packaging_osx/venv/bin/python` at the top of it (that's all that gets matched), why we would ever want to leave it? I'll make it behave differently for the two directories to be conservative as you suggest, but it sounds like you are aware of a situation that I'm not so I am trying to pick your brain :smile:

dirk-thomas: Because other locations are not confirming with FHS. So if a file ends of there we shouldn't touch it since it shouldn't work / be there in the first place.

dhood: ok, thank you, that's what I was missing

ros2_batch_job/packaging.py
```diff
@@ -112,9 +112,21 @@ def build_and_test_and_package(args, job):
     # Only on Linux and OSX Python scripts have a shebang line
     if args.os in ['linux', 'osx']:
         print('# BEGIN SUBSECTION: rewrite shebang lines')
+        filenames = []
```
dirk-thomas: Nitpick: it might be good to name this variable `paths` since it contains full paths rather than just filenames.

ros2_batch_job/packaging.py
```diff
@@ -112,9 +112,21 @@ def build_and_test_and_package(args, job):
...
+
         bin_path = os.path.join(args.installspace, 'bin')
-        for filename in os.listdir(bin_path):
-            path = os.path.join(bin_path, filename)
+        filenames.append([os.path.join(bin_path, filename) for filename in os.listdir(bin_path)])
```
dirk-thomas: It might be good to add the same check as below `os.path.isfile` (just in case someone creates a subfolder under `bin`).

ros2_batch_job/packaging.py
```diff
@@ -112,9 +112,21 @@ def build_and_test_and_package(args, job):
...
+
+        # Demo nodes are installed to 'lib/<package_name>'
+        lib_path = os.path.join(args.installspace, 'lib')
+        for lib_sub_dir in next(os.walk(lib_path))[1]:
+            sub_dir_path = os.path.join(lib_path, lib_sub_dir)
+            sub_dir_contents = [os.path.join(sub_dir_path, name) for name in os.listdir(sub_dir_path)]
+            filenames.extend([filename for filename in sub_dir_contents if os.path.isfile(filename)])
+        for lib_sub_dir in next(os.walk(lib_path))[1]:
+            filenames.append([os.path.join(lib_sub_dir, filename) for filename in os.listdir(lib_sub_dir)])
```
dirk-thomas: Aren't these two loops doing something similar / overlapping?

dhood: ha, yeah, forgot to delete one of them :sweat:

dhood:
Thanks for iterating with me: I appreciate you answering my questions so I could understand the 'why'. I'll wait for the CI before merging:

ci packaging linux: [Build Status](http://ci.ros2.org/job/ci_packaging_linux)
ci packaging osx: [Build Status](http://ci.ros2.org/job/ci_packaging_osx/5/)

dhood:
these are still not removed: https://github.com/ros2/ci/pull/88/files#diff-ff5a69e3fa8788343837457e9dd458f8R146 so I'll do another round of iteration on this. There should not be any `__rmw_` suffixes around anymore so I will remove that check.

dhood:
Tested locally and CI's happy now for each platform so I'm going to merge this before the nightlies start provided there are no objections to the commits since the approval

mikaelarguedas:
:+1:

PR #90, opened 2017-06-28
pass ament_build_args to ament invocation

mikaelarguedas:
Follow up of: #84 
The parameter ament-build-args was actually not used by the packaging job resulting in security not being compiled and shipped for Fast-RTPS.

packaging job with this change: 
* macOS [Build Status](http://ci.ros2.org/job/ci_packaging_osx/8/)

@Kukanani FYI


mikaelarguedas:
Note that the default value of the build arguments being shared bewteen the jobs, the packaging job is now using `--parallel`

mikaelarguedas:
I'm going to merge this and rerun the packaging jobs for the testing today. If there is any further comment please feel free to comment on this PR

mikaelarguedas:
Note that the new packaging artifacs will include the hotfix/besteffort changes from eProsima that just got merged on master

PR #91, opened 2017-07-12
use OpenSplice 6.7.0 Debian package

dirk-thomas:
Waiting for the final Debian package to become available in packages.osrfoundation.org (@j-rivero).


dirk-thomas:
The new Debian is available (https://build.osrfoundation.org/job/opensplice-debbuilder/27/) and the CI build is going well: [Build Status](http://ci.ros2.org/job/ci_linux/2886/)

Ready for review.

PR #92, opened 2017-07-12
Add nightly turtlebot

clalancette:
The change I really wanted to do was to add a nightly build job for turtlebot.  However, I found the way the current jobs are created to be confusing, so I also rewrote the loop in the first patch.  See the commit message for the first patch for an explanation.

I tested this in a dry-run locally; the only things that were going to be different from before were a small typo fix in the nightly_*_repeated jobs (an extra space at the end of the spec), and the creation of the two new nightly turtlebot jobs.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -38,6 +38,8 @@ except ImportError:
     sys.exit("Could not import symbol from ros_buildfarm, please update ros_buildfarm.")
 
 DEFAULT_REPOS_URL = 'https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos'
+DEFAULT_MAIL_RECIPIENTS = 'ros2-buildfarm@googlegroups.com'
+PERIODIC_JOB_SPEC = '30 7 * * *'
 
 template_prefix_path[:] = \
     [os.path.join(os.path.abspath(os.path.dirname(__file__)), 'job_templates')]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please double check the current configuration by comparing the newly generated config with the existing jobs (without actually overwriting them). You will notice that the patch has a couple of bugs atm.

clalancette:
Ah, good call.  I missed that.

create_jenkins_job.py
```diff
@@ -114,73 +116,87 @@ def main(argv=None):
...
+        })
+
+        # configure turtlebot jobs on Linux only for now
+        if os_name in ['linux', 'linux-aarch64']:
+            create_job(os_name, 'ci_turtlebot-demo_%s' % (os_name), 'ci_job.xml.em', {
+                'cmake_build_type': 'None',
+                'turtlebot_demo': True,
+                'supplemental_repos_url': 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos',
+            })
+            create_job(os_name, 'nightly_turtlebot-demo-%s' % (os_name), 'ci_job.xml.em', {
```
dhood: this needs to set turtlebot_demo and supplemental_repos_url, right? I'd also recommend the name include debug for consistency

mikaelarguedas: `'turtlebot_demo': True,` << I think that this is needed as well

clalancette: Yep, you are exactly right, totally missed.  Thanks, I've added that now, along with the "debug" name.  Please take another look when you have a moment.

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
         # configure nightly triggered job
-        job_name = 'nightly_' + os_name + '_debug'
-        if os_name == 'win':
-            job_name = job_name[0:-2]
-        job_data['cmake_build_type'] = 'Debug'
-        job_config = expand_template('ci_job.xml.em', job_data)
-        configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
+        short_name = 'nightly_' + os_name + '_debug'
+        if os_name == 'windows':
+            short_name = 'nightly_win_deb'
```
mikaelarguedas: Is it necessary to change the logic of how the names shorten for windows ? (same everywhere the job names are hardcoded)

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
 
         # configure nightly coverage job on x86 Linux only
         if os_name == 'linux':
-            job_name = 'nightly_' + os_name + '_coverage'
-            job_data['cmake_build_type'] = 'Debug'
-            job_data['enable_c_coverage_default'] = 'true'
-            job_config = expand_template('ci_job.xml.em', job_data)
-            configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
-            job_data['enable_c_coverage_default'] = 'false'
+            create_job(os_name, 'nightly_linux_coverage', 'ci_job.xml.em', {
```
mikaelarguedas: Same remark asabove, I'm not sure hardcoding the job names is a good idea. If we want to create a coverage job for osx we will need to create another conditional bloc while we could just us this one if it stayed as it was before

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
+
+        # configure turtlebot jobs on Linux only for now
+        if os_name in ['linux', 'linux-aarch64']:
+            create_job(os_name, 'ci_turtlebot-demo_%s' % (os_name), 'ci_job.xml.em', {
+                'cmake_build_type': 'None',
+                'turtlebot_demo': True,
+                'supplemental_repos_url': 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos',
+            })
+            create_job(os_name, 'nightly_turtlebot-demo_%s_debug' % (os_name), 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
```
mikaelarguedas: Note: Cartographer doesn't build in `Debug` AFAIK so I would expect this job to fail every night

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
+            'cmake_build_type': 'None',
+            'time_trigger_spec': PERIODIC_JOB_SPEC,
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            'ament_test_args_default': '--retest-until-fail 20 --ctest-args -LE linter --',
+        })
+
+        # configure turtlebot jobs on Linux only for now
+        if os_name in ['linux', 'linux-aarch64']:
+            create_job(os_name, 'ci_turtlebot-demo_%s' % (os_name), 'ci_job.xml.em', {
+                'cmake_build_type': 'None',
```
mikaelarguedas: We usually never have None nightlies, is there a reason to have one for the turtlebot job ?

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
+        create_job(os_name, short_name, 'ci_job.xml.em', {
+            'cmake_build_type': 'None',
+            'time_trigger_spec': PERIODIC_JOB_SPEC,
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            'ament_test_args_default': '--retest-until-fail 20 --ctest-args -LE linter --',
+        })
+
+        # configure turtlebot jobs on Linux only for now
+        if os_name in ['linux', 'linux-aarch64']:
+            create_job(os_name, 'ci_turtlebot-demo_%s' % (os_name), 'ci_job.xml.em', {
```
mikaelarguedas: every other job uses a `job_name` variable and uses `+` for concatenation, can you do the same here to stay consistent?

create_jenkins_job.py
clalancette: OK, yeah.  On this one, I agree with you.  I'll change this back to building it up from the os_name.

create_jenkins_job.py
clalancette: Yeah, that's fine, I'll fix that.  This was a cut-n-paste from below, but I agree with you, it should be consistent.

create_jenkins_job.py
clalancette: This was a cut-n-paste from the old turtlebot job.  I'm not sure where that configuration originally came from.

create_jenkins_job.py
clalancette: Ah, I had forgotten about that fact, good call.  Will change it to release instead.

create_jenkins_job.py
clalancette: Not really necessary, no, but the previous way seemed overly complicated to me.  It was, in effect, hard-coding the short names, just in a less obvious way.

clalancette:
> If the goal is to confirm that the code still builds I'm not sure we need to create several build types because it will take a lot of buildfarm time.

Sorry, I missed this bit before.  As it stands, this patch will create 2 new jobs: nightly_turtlebot-demo_linux and nightly_turtlebot-demo_linux-aarch64.  Are you worried that adding those two nightly builds will tax the build system, or were you concerned about something else?

mikaelarguedas:
> Sorry, I missed this bit before. As it stands, this patch will create 2 new jobs: nightly_turtlebot-demo_linux and nightly_turtlebot-demo_linux-aarch64. Are you worried that adding those two nightly builds will tax the build system, or were you concerned about something else?

I was concerned about mutliple turtlebot jobs **per architecture** (e.g. before one Debug and one None). If the current patch "create 2 new jobs: nightly_turtlebot-demo_linux and nightly_turtlebot-demo_linux-aarch64" that's good :+1:

create_jenkins_job.py
```diff
@@ -114,73 +116,89 @@ def main(argv=None):
...
+        })
+
+        # configure turtlebot jobs on Linux only for now
+        if os_name in ['linux', 'linux-aarch64']:
+            create_job(os_name, 'ci_turtlebot-demo_' + os_name, 'ci_job.xml.em', {
+                'cmake_build_type': 'None',
+                'turtlebot_demo': True,
+                'supplemental_repos_url': 'https://raw.githubusercontent.com/ros2/turtlebot2_demo/master/turtlebot2_demo.repos',
+            })
+            create_job(os_name, 'nightly_turtlebot-demo_' + os_name, 'ci_job.xml.em', {
```
mikaelarguedas: > I'd also recommend the name include debug for consistency

This still applies, we should add `_release` to the job name here

create_jenkins_job.py
mikaelarguedas: the previous logic was creating the job_name the same way for everyone and then modifying it to make it shorter for windows. This new approach replace the variable content with a hard-coded string literal discarding the previous content so I think it's more likely that the naming becomes inconsistent if we change one but not the other

clalancette:
All right, I added release back to the name, and I've changed it up so that we are building the Windows names again.  I still made some changes in how we built the names; in particular I made it so that they have a common set of rules.  Ideally we'd apply those rules to all jobs, but that would currently replace some of the windows jobs with other jobs (ci_windows -> ci_win, packaging_windows -> packaging_win, etc).  While that is technically more consistent, I don't think we need to do that right now, so I just special cased the ones that were special-cased before.  @mikaelarguedas please take another look when you get a chance.

mikaelarguedas:
> I made it so that they have a common set of rules. Ideally we'd apply those rules to all jobs,

:+1: agreed that a common rule makes more sense but appreciate that you apply them only to the jobs that needed it to keep the naming. Ideally the rule of cropping after the 15th character should be shared as well rather than duplicated for all the jobs needing cropping

clalancette:
Thanks for the review.  I'm going to merge this and then deploy it.

PR #93, opened 2017-07-27
Less hardcoded opencv libraries path

mikaelarguedas:
Path of opencv libraries to install is hardcoded in the CI script. Thiis PR reads opencv directory from environment variable. Should fix failing packaging job:

[Build Status](http://ci.ros2.org/view/packaging/job/packaging_windows/677) 

With this fix:
[Build Status](http://ci.ros2.org/view/All/job/ci_packaging_windows/7/)

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import glob
 import os
 import platform
 import shutil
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Ok I went a more generic way and install all opencv dlls rather than hardcoded ones
CI seems happier now:
[Build Status](http://ci.ros2.org/view/All/job/ci_packaging_windows/9/) 

Putting this in review

ros2_batch_job/packaging.py
```diff
@@ -166,10 +167,9 @@ def exclude(filename):
     elif args.os == 'windows':
         archive_path = 'ros2-package-windows-%s.zip' % platform.machine()
         # NOTE(esteve): hack to copy our custom built VS2015-compatible OpenCV DLLs
-        opencv_libdir = os.path.join('c:/', 'opencv', 'build', 'x64', 'vc14', 'bin')
-        for libfile in ['opencv_core2412.dll', 'opencv_highgui2412.dll']:
-            libpath = os.path.join(opencv_libdir, libfile)
-            shutil.copy(libpath, os.path.join(args.installspace, 'bin', libfile))
+        opencv_libdir = os.path.join(os.environ['OpenCV_DIR'], 'x64', 'vc14', 'bin')
+        for libpath in glob.glob(r'%s/*.dll' % opencv_libdir):
```
dirk-thomas: Just a question: why does it need to be a *raw* string?

mikaelarguedas: good point #94

PR #96, opened 2017-08-22
Update ci_launcher job description

dhood:
I noticed the original description is no longer accurate.

### DIFF
--- a/job_templates/ci_launcher_job.xml.em
+++ b/job_templates/ci_launcher_job.xml.em
@@ -3,6 +3,7 @@
   <actions/>
   <description>
     Launches a build of the CI job for each supported platform with the same set of parameters.
+    Use of RTI Connext will always be disabled on aarch64 jobs.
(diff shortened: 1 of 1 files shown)


job_templates/ci_launcher_job.xml.em
```diff
@@ -2,7 +2,7 @@
 <project>
   <actions/>
   <description>
-    Note: The Windows jobs ignore the middleware selection flags.
+    Launches a build of the CI job for each supported platform with the same set of parameters.
```
dirk-thomas: `aarch64` ignore Connext atm. Not sure if it makes sense to mention it here though...

dhood: while we'll have to maintain this, I think it's good to document it: https://github.com/ros2/ci/pull/96/commits/4b819d4c044b335e138d957178df706a9c5ec8d3

PR #98, opened 2017-08-29
add some flake8 plugins

dirk-thomas:
From the many available plugins those caught my eye since they check for either "unclean" code, more modern Python code, and style we usually have to look for manually. Hopefully this helps reducing the need for "nitpick" comments.

I update the code in the `ament` repositories to have some diffs to review so that it is easier to get an impression what those plugins are checking for. Please see the referenced PR below.

Before this can be merged all the packages need to pass cleanly.


nuclearsandwich:
Are these plugins automatically detected by flake8 or is some configuration required to use them locally?

dirk-thomas:
> Are these plugins automatically detected by flake8

Yes.

mikaelarguedas:
the list looks sensible to me, it seems that several of them didn't flag any error on the set of referenced PRs. I'm in favor of moving forward with making other packages comply with this set of rules and we can refine later based on the changes it introduces

ros2_batch_job/__main__.py
```diff
@@ -49,7 +49,15 @@
     'EmPy',
     'coverage',
     'flake8',
+    'flake8-blind-except',
+    'flake8-builtins',
+    'flake8-class-newline',
+    'flake8-commas',
```
dirk-thomas: We talked about the list of plugins in todays meeting and will keep all but this one.

dirk-thomas:
I update this as well as all referenced PR. I removed the forced commas but kept them in a few places or not wrapped a closing parenthesis where I thought it makes sense. I also added the specific error code to the new `# noqa` comments.

Ready for a second round of review. At least the current set of repos (everything in the `ament` org unit) passes: [Build Status](http://ci.ros2.org/job/ci_linux/3105/).

[Still needs similar changes for other packages before this can be merged.]

dirk-thomas:
With the additional PRs this passes on all platforms:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3112/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/478/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2486/) (unrelated test failures)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3172/) (unrelated test failures)

Ready for merge after review.

mikaelarguedas:
Can you run a turtlebot job as well? Otherwise I'm expecting test failures related to this in the next nightlies

dirk-thomas:
> Can you run a turtlebot job as well

I didn't update any turtlebot related repositories therefore I expect them to have failing linter tests: [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/92/) It would be great if someone else could take a look at those and make PRs for these repos where necessary.

mikaelarguedas:
Ok then we should not merge this until the turtlebot PRs are created and reviewed

dirk-thomas:
> Ok then we should not merge this until the turtlebot PRs are created and reviewed

Sounds fine with me (if that happens in the near future). The current PRs can be reviewed though.

dirk-thomas:
`teleop_twist_keyboard` seems to be the only affected package:

[Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/94/)

mikaelarguedas:
Yeah I think that's the only python package in the tb2 stack. Thanks for fixing this one too :+1:

mikaelarguedas:
this looks good to me overall.
Not sure what we should do about the multiline string quoting thought. Maybe it would be better to use "real" multiline string using triple `"` rather than having strings split across lines using different quotes according to their content

dirk-thomas:
Multiline strings using triple `"` have the disadvantage of "breaking" the indentation level. I don't they are a viable alternative.

mikaelarguedas:
fair enough :+1:

nuclearsandwich:
Is the [Python style guide](https://github.com/ros2/ros2/wiki/Developer-Guide#python) otherwise up to date? If so I can just add that we're using these flake8 plugins but if it's not correct in its current form I'm not sure I know enough to fix it.

mikaelarguedas:
I think that we'll need to update the bullet list where we "chose the following more precise rule where PEP 8 leaves some freedom"

Not sure all of the newly enforced rules have a pep8 section so we may need to add some bullets outside of the PEP8 list as well (I'm guessing here I didnt actually check what was missing)

mikaelarguedas:
Follow-up on this:
- these plugins should be added to the installations instructions (I'm on it)
- as mentioned by @nuclearsandwich the python style guide should be updated accordingly

PR #99, opened 2017-09-11
seaparate checkboxes for fastrtps and opensplice in packaging job

dirk-thomas:
Separate the combo box to select one rmw impl. with separate checkboxes to enable selecting both when supported by the platform.

* Linux [Build Status](http://ci.ros2.org/job/packaging_linux/807/)
* Windows [Build Status](http://ci.ros2.org/job/packaging_windows/724/)

### DIFF
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -25,16 +25,16 @@
     cmake_build_type=cmake_build_type,
     ament_build_args_default=ament_build_args_default,
 ))@
-        <hudson.model.ChoiceParameterDefinition>
-          <name>CI_USED_RMW_IMPL</name>
-          <description/>
-          <choices class="java.util.Arrays$ArrayList">
-            <a class="string-array">
-              <string>FastRTPS</string>
-              <string>OpenSplice</string>
-            </a>
-          </choices>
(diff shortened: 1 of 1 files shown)


job_templates/packaging_job.xml.em
```diff
@@ -25,16 +25,16 @@
...
-        </hudson.model.ChoiceParameterDefinition>
+        <hudson.model.BooleanParameterDefinition>
+          <name>CI_USE_FASTRTPS</name>
+          <description>By setting this to True, the build will attempt to use eProsima&apos;s FastRTPS.</description>
+          <defaultValue>@(use_fastrtps_default)</defaultValue>
+        </hudson.model.BooleanParameterDefinition>
+        <hudson.model.BooleanParameterDefinition>
+          <name>CI_USE_OPENSPLICE</name>
+          <description>By setting this to True, the build will attempt to use PrismTech&apos;s OpenSplice.</description>
+          <defaultValue>@(use_opensplice_default)</defaultValue>
```
mikaelarguedas: Question: If we check both boxes, fast-RTPS will still be the default right ?

dirk-thomas: Yes.

PR #100, opened 2017-09-13
Install opensplice on aarch64

sloretz:
This updates the dockerfile to install libopensplice67 on arm. The package would come from `http://packages.osrfoundation.org/gazebo/ubuntu-stable/dists/xenial/main/binary-arm64/Packages`. It is intented to fix the failure seen in http://ci.ros2.org/job/packaging_linux-aarch64/171/console

* Packaging Linux [Build Status](http://ci.ros2.org/job/packaging_linux-aarch64/173/)

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -42,9 +42,9 @@ RUN pip3 install -U setuptools pip virtualenv
 RUN apt-get update && apt-get install --no-install-recommends -y gcovr
 
 # Install the OpenSplice binary from the OSRF repositories.
-RUN if test \( ${PLATFORM} = aarch64 -o ${PLATFORM} = x86 \) ; then apt-get update && apt-get install --no-install-recommends -y libopensplice67=6.7.0+osrf2-2~xenial; fi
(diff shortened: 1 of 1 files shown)


sloretz:
First CI will fail, looks like the platform is `arm` instead of `aarch64` :(. Rerunning.

dhood:
@sloretz FYI we have CI versions of the packaging jobs for testing changes to the jobs (we try to avoid triggering the standard packaging jobs with non-default parameters)

dhood:
Looks like this change pushed us back over [the layer limit encounted building the turtlebot demo](https://github.com/ros2/ci/issues/75): [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/99/). We could only install it if it's selected which would buy us some time, or follow up on proposals in https://github.com/ros2/ci/issues/75

dirk-thomas:
> We could only install it if it's selected

That would also imply that we don't reuse the docker images when toggling the option. Therefore I would suggest to "collapse" some other line to get back under the limit as a short term workaround.

sloretz:
There are quite a few layers that could be collapsed. These 3 could become one.

https://github.com/ros2/ci/blob/3c2fe6ac7e0520e6832f468d6b1560b87bc9889a/linux_docker_resources/Dockerfile#L14-L21

dirk-thomas:
If the first block is enough for now I would start with those.

PR #101, opened 2017-09-14
Switch to Connext 5.3.0

dirk-thomas:
The first commit uses the S3 URL to the newly available package of Connext 5.3.0.

The second commit disables the option to use the OSRF-built Debian packages of Connext for now (since we only have them for version 5.2.3). I don't think we need to iterate on this giving the new way of using the installer.

The third and fourth commit switch to Connext 5.3.0 on Windows and Mac OS as well (requires ros2/rmw_connext#250).

For now I don't plan to update the wiki pages since users are likely using Beta 3 which needs 5.2.x. Currently the default branch should mostly work with both. Updating the security part might change that in the future. Then we likely need to separate the instructions somehow.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3239/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_windows/3300/)
* Windows [Build Status](http://ci.ros2.org/job/ci_osx/2581/)

RTI doesn't provide Visual Studio 2015 packages of Connext 5.3.0 for Windows 8. Therefore I propose to take `eatable` offline until it has been upgraded to Windows 10.

### DIFF
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -32,7 +32,7 @@ case "${CI_ARGS}" in
         ;;
       *)
         echo "Installing Connext binaries off RTI website..."
-        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti-installer.run /home/rosbuild
+        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild
         if [ $? -ne 0 ]
         then
           echo "Connext not installed correctly (maybe you're on an ARM machine?)." >&2
--- a/job_templates/snippet/property_parameter-definition.xml.em
+++ b/job_templates/snippet/property_parameter-definition.xml.em
@@ -30,11 +30,13 @@ This tests the robustness to whitespace being within the different paths.</descr
           <description>By setting this to True, the build will disable the Connext dynamic rmw implementation.</description>
           <defaultValue>@(disable_connext_dynamic_default)</defaultValue>
         </hudson.model.BooleanParameterDefinition>
+        <!--
         <hudson.model.BooleanParameterDefinition>
           <name>CI_USE_OSRF_CONNEXT_DEBS</name>
           <description>By setting this to True, the build will use the deb packages built by OSRF for Connext, instead of the binaries off the RTI website (applies to linux only).</description>
           <defaultValue>@(use_osrf_connext_debs_default)</defaultValue>
         </hudson.model.BooleanParameterDefinition>
+        -->
         <hudson.model.BooleanParameterDefinition>
           <name>CI_USE_FASTRTPS</name>
           <description>By setting this to True, the build will attempt to use eProsima&apos;s FastRTPS.</description>
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -53,7 +53,7 @@ RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-insta
 RUN pip3 install pexpect
 
 # Get and install the RTI web binaries.
-RUN curl --silent http://s3.amazonaws.com/RTI/Bundles/5.2.3/Evaluation/rti_connext_dds-5.2.3-eval-x64Linux3gcc4.8.2.run -o /tmp/rti-installer.run && chmod +x /tmp/rti-installer.run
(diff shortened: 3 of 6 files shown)


linux_docker_resources/entry_point.sh
```diff
@@ -32,7 +32,7 @@ case "${CI_ARGS}" in
         ;;
       *)
         echo "Installing Connext binaries off RTI website..."
-        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti-installer.run /home/rosbuild
+        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild
```
dhood: IMO it'd be better to rename the installer at the download step so the version doesn't need to be referred to in two files

mikaelarguedas:
> RTI doesn't provide Visual Studio 2015 packages of Connext 5.3.0 for Windows 8. Therefore I propose to take eatable offline until it has been upgraded to Windows 10.

Slightly related Connext doesnt provide VS2017 packages, that may impact our considerations to upgrade

linux_docker_resources/entry_point.sh
dirk-thomas: > IMO it'd be better to rename the installer at the download step so the version doesn't need to be referred to in two files

The download is not a single file anymore but an archive containing multiple files. So I would rather not touch the files from within the archive.

dhood: I understand it's an archive but I don't see how it affects what I'm saying, that now we have to update the name of the connext web installation in multiple files. I haven't understood the reason against renaming the installer after extracting the archive?

dirk-thomas: There are currently many locations in which the exact Connext version is hard coded. I guess I just don't see it as that valuable to reduce one of the many.

Please feel free to modify the patch in place and trigger a new CI build for it.

dhood: no worries, was checking if I was missing something about the reasoning

dirk-thomas:
I have already updated all Jenkins nodes (OS X as well as Windows) except:
* `mini2` which is unable to download the archive atm (I will try later / tomorrow again) and
* `eatable` (which will be disabled since it is still using Windows 8).

I think it would be good to get some more :+1: from the @ros2/team on this before merging it.

mikaelarguedas:
Oh and iterate on the nodes that need some love to be able to use it...

dirk-thomas:
> I guess the only remaining question is how to document the connext install/sourcing for people tracking master and not beta3

With this merged it should still be fine to build and use the default branches with both Connext 5.2.x and 5.3. Only the upcoming security PR will actually require 5.3.

sloretz:
Will merging this resolve [this ci failure](http://ci.ros2.org/job/ci_linux/3253/console)?

```
17:37:41 Step 23 : RUN curl --silent http://s3.amazonaws.com/RTI/Bundles/5.2.3/Evaluation/rti_connext_dds-5.2.3-eval-x64Linux3gcc4.8.2.run -o /tmp/rti-installer.run && chmod +x /tmp/rti-installer.run
17:37:41  ---> Running in 5bce308ae2e7
17:45:36 The command '/bin/sh -c curl --silent http://s3.amazonaws.com/RTI/Bundles/5.2.3/Evaluation/rti_connext_dds-5.2.3-eval-x64Linux3gcc4.8.2.run -o /tmp/rti-installer.run && chmod +x /tmp/rti-installer.run' returned a non-zero code: 56
17:45:36 Build step 'Execute shell' marked build as failure
```

dirk-thomas:
> Will merging this resolve this ci failure?

That looks like a temporary network glitch. The URL seems to work fine for me. Have you tried it again?

This patch won't add any retry logic to the download - just change the URL to the new version.

sloretz:
Ah, I thought the file might have been removed in preparation for the switch to 5.3. CI is re-running now.

mikaelarguedas:
@dirk-thomas can you give us an update here ? mini2 has been successfully updated and we take eatable offline until it's upgraded to windows 10?

dirk-thomas:
> can you give us an update here ? mini2 has been successfully updated and we take eatable offline until it's upgraded to windows 10?

Yes, since these are the prerequisites to merge this. Both has happened and I updated the [previous comment](https://github.com/ros2/ci/pull/101#issuecomment-329619325) with that info.

mikaelarguedas:
Cool thanks. I missed the update :confused: , sorry for the noise

wjwwood:
The wiki instructions were never updated, I just updated them now.

PR #102, opened 2017-09-15
Squash a few docker layers

sloretz:
Should get image below layer limit for building turtlebot demo on arm

(see https://github.com/ros2/ci/pull/100#issuecomment-329560974)

[Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/100/)

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -12,12 +12,11 @@ RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 
 # net-tools is for ifconfig
+# Get curl for fetching the repo keys.
+# Get https transport for APT.
 RUN apt-get update && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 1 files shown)


clalancette:
Agreed with @dirk-thomas .  Personally, I would just move/combine all of those comments above the RUN statement itself, which just makes it a bit easier to read.

sloretz:
@dirk-thomas It's an abuse of command substitution, but it works. @clalancette comments moved above `RUN` in 48026bec703499640af12bced4141fdea63b09c1.

dirk-thomas:
Yeah, very "creative" with the backticked comments :wink:

This looks much simpler now :+1:

sloretz:
I'm going to squash/merge this. While the CI job isn't finished and it's not testing the last commit, I think the last commit changes are low risk and the build has gone far enough to show that enough layers have been squashed.

PR #105, opened 2017-10-10
Install rti security plugins

mikaelarguedas:
follow up of https://github.com/ros2/ci/pull/101

This installs the connext security plugin but can be invoked with a list of `.rtipkg` files if we need to install more of them

### DIFF
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -32,7 +32,7 @@ case "${CI_ARGS}" in
         ;;
       *)
         echo "Installing Connext binaries off RTI website..."
-        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild
+        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild --rtipkg_paths /tmp/rti_security_plugins-5.3.0-eval-x64Linux3gcc5.4.0.rtipkg
         if [ $? -ne 0 ]
         then
           echo "Connext not installed correctly (maybe you're on an ARM machine?)." >&2
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -32,7 +32,7 @@ case "${CI_ARGS}" in
         ;;
       *)
         echo "Installing Connext binaries off RTI website..."
-        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild --rtipkg_paths /tmp/rti_security_plugins-5.3.0-eval-x64Linux3gcc5.4.0.rtipkg
+        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild/rti_connext_dds-5.3.0 --rtipkg_paths /tmp/rti_security_plugins-5.3.0-eval-x64Linux3gcc5.4.0.rtipkg
         if [ $? -ne 0 ]
         then
           echo "Connext not installed correctly (maybe you're on an ARM machine?)." >&2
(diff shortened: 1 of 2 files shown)


linux_docker_resources/entry_point.sh
```diff
@@ -32,7 +32,7 @@ case "${CI_ARGS}" in
         ;;
       *)
         echo "Installing Connext binaries off RTI website..."
-        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild
+        python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.0-eval-x64Linux3gcc5.4.0.run /home/rosbuild/rti_connext_dds-5.3.0 --rtipkg_paths /tmp/rti_security_plugins-5.3.0-eval-x64Linux3gcc5.4.0.rtipkg
```
dirk-thomas: Why does the `install_directory` change?

mikaelarguedas: Right now the new function takes the install location and append `/bin/rtipkginstall`.
We could work around it in different ways but this seems to be the least invasive and simpler approach.

Current approach: path the full installation directory given that we know what it is based on the installer we use
Alternatives:
- add another argument to the script to pass the installation folder name
- guess it based on the installer filename we pass to the script
- hardcode it in the script

dirk-thomas: Before the code was passing `/home/rosbuild` which made `install_connext()` install into `/home/rosbuild/rti_connext_dds-5.3.0`.

If this patch passes `/home/rosbuild/rti_connext_dds-5.3.0` doesn't the function install Connext into `/home/rosbuild/rti_connext_dds-5.3.0/rti_connext_dds-5.3.0`?

mikaelarguedas: The file location after installation is the same if you pass `/home/rosbuild` or `/home/rosbuild/rti_connext_dds-5.3.0`

CI passing with the current patch: http://ci.ros2.org/job/ci_linux/3316 (ignore the security test failures they're due to an unrelated domain_id issue)

dirk-thomas: > The file location after installation is the same if you pass `/home/rosbuild` or `/home/rosbuild/rti_connext_dds-5.3.0`


Interesting... Sounds good to me then.

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -46,10 +48,34 @@ def install_connext(installer_path, install_directory):
...
+    parser.add_argument('--rtipkg_paths', nargs='*', default=[])
 
     args = parser.parse_args()
     install_connext(args.installer_path, args.install_directory)
+    # Install successful, now installing rti plugins if any
+    for plugin_path in args.rtipkg_paths:
+        if os.path.isfile(plugin_path):
+            install_plugin(plugin_path, args.install_directory)
+        else:
+            print("RTI package '%s' not found, Skipping..." % plugin_path, file=sys.stderr)
```
dirk-thomas: Should we enforce that the passed plugins are being installed rather than skip them with just a warning message?

mikaelarguedas: Yeah we definitely can do that, I took the conservative approach in case we had some "optional" plugins in the future. I can raise here with the same error message if it's preferred

dirk-thomas: Sounds "safer" to me. If we need optional plugins in the future we can either relax it or pass them using a different option.

mikaelarguedas: Sounds good https://github.com/ros2/ci/pull/105/commits/5a6daba2a141d572482e2e534c1d87f21cae54ea

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -46,10 +47,34 @@ def install_connext(installer_path, install_directory):
...
+    parser.add_argument('--rtipkg_paths', nargs='*', default=[])
 
     args = parser.parse_args()
     install_connext(args.installer_path, args.install_directory)
+    # Install successful, now installing rti plugins if any
+    for plugin_path in args.rtipkg_paths:
+        if os.path.isfile(plugin_path):
+            install_plugin(plugin_path, args.install_directory)
+        else:
+            raise RuntimeError("RTI package '%s' not found, Skipping..." % plugin_path)
```
dirk-thomas: Remove `, Skipping...`?

mikaelarguedas:
Thanks for the cleanup

PR #109, opened 2017-11-10
install pytest and pytest-cov

dirk-thomas:
Connect to ament/ament_tools#169.

The new dependencies need to be added to the install instructions in the wiki once this is merged.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -62,6 +62,9 @@ pip_dependencies = [
     'pep8',
     'pydocstyle',
     'pyflakes',
+    'pytest',
+    'pytest-cov',
+    'pytest-runner',
     'pyyaml',
     'vcstool',
 ]


mikaelarguedas:
Is the version from apt not acceptable ?

dirk-thomas:
> Is the version from apt not acceptable ?

The Debian package of `pytest` and `pytest-cov` should be fine on Xenial. `pytest-runner` is not available as a Debian package in Xenial though. Our pipeline always installs the pip packages (mostly because of the non-Debian platforms) so I didn't bother adding those two packages also from Debian.

mikaelarguedas:
> Our pipeline always installs the pip packages (mostly because of the non-Debian platforms) so I didn't bother adding those two packages also from Debian.

Agreed it doesnt change much for our CI given that they install from scratch every time.
I'm not sure but I though the motivation for relying on the apt versions wherever we can is to allow us and the users to have updates on this packages through their usual upgrade process and not having to update each of them by manually invoking pip.
Do you remember the motivation for installing all the other test dependencies from pip and apt ? (nose or mock for example?)

dirk-thomas:
> Do you remember the motivation for installing all the other test dependencies from pip and apt ?

I guess we just didn't bother to skip some of the PIP installs on Ubuntu for packages which are installed from Debian packages.

mikaelarguedas:
I'd suggest installing via `apt` whatever we recommend users to use via `apt`. This also allows the dependency part of the instructions to be aligned with the content of the Dockerfiles and the one of our machines.
I'm happy to provide a PR if you agree

dirk-thomas:
> I'm happy to provide a PR if you agree

Sounds good to me. :+1:

PR #110, opened 2017-11-10
add pytest and pytest-cov to apt packages to install

mikaelarguedas:
Follow-up of https://github.com/ros2/ci/pull/109

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -31,7 +31,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this achieves what your are describing in the referenced ticket. It installs the Debian packages of `pytest` but later it still installs the packages from `pip` so the environment will still *not* match what a users has when installing only the Debian packages.

mikaelarguedas:
> I don't think this achieves what your are describing in the referenced ticket. It installs the Debian packages of pytest but later it still installs the packages from pip so the environment will still not match what a users has when installing only the Debian packages.

yes but if I build this container I have the same setup as the one of a user following the installation instructions. And these are also the instructions I'll follow on my native machine.

I agree a better outcome would be to not install the pip packages we can get from apt when we run Linux jobs but that's not what I was aiming in my original comment.

dirk-thomas:
With 7f057ebbe94a8b17ef3b9ba2fee28a967295192a the patch reflect the instructions as well as what we actually need.

mikaelarguedas:
Ok merging as is as this matches what is in the instructions

PR #111, opened 2017-11-14
Revert "add pytest and pytest-cov to apt packages to install"

mikaelarguedas:
Reverts ros2/ci#110

We need features available since pytest3.2 and xenial ships with 2.8.7. See https://github.com/ament/ament_cmake/pull/116#discussion_r150923774 for more details


sloretz:
Looks like this just removes the dependency from apt. Won't CI be broken unless it installs `pytest` via pip too?

mikaelarguedas:
CI is installing it via pip for all platforms that's why it was working on CI https://github.com/ros2/ci/commit/9c1daf6929ab3c18a06f1d64725d40c508e53aa9. Adding it to apt was for completness if we decide to install pip packages only on OSX and Windows in the future

PR #114, opened 2017-12-02
install curl for resource_retriver

wjwwood:
Connects to ros2/ros2#435


nuclearsandwich:
> I was initially concerned about the new layer added, I think one more layer is fine but would prefer to see a build passing on arm with this change before merging.
> 
> I'll try to reduce the number of layers as I think some of them are not necessary anymore (like poco that we build from source anyway)

I also almost raised a concern about the layer count. In the end I didn't feel the need to muddy my approval of the current state. But I do think we can consolidate apt dependencies of packages in ros2.repos into a single invocation of apt-get.

PR #117, opened 2017-12-04
Add qt5 dependencies for rviz2 on Linux

nuclearsandwich:
Using the package manifests in ros2/rviz as a template add the necessary qt dependencies for rviz.

I'm not sure how much this is going to add to image generation time, but without recommends hopefully it won't be too extreme.

### DIFF
--- a/ros2_batch_job/osx_batch/__init__.py
+++ b/ros2_batch_job/osx_batch/__init__.py
@@ -49,7 +49,7 @@ class OSXBatchJob(BatchJob):
             os.environ['OSPL_HOME'] = os.path.join(os.environ['HOME'], 'opensplice', 'HDE', 'x86_64.darwin10_clang')
         # TODO(wjwwood): remove this when qt5 is linked on macOS by default
         # See: https://github.com/Homebrew/homebrew-core/issues/8392#issuecomment-334328367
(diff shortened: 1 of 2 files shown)


wjwwood:
Let's hold this until I can test it against my pr to add rviz to the repos file.

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -47,6 +47,9 @@ def pre(self):
               os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix_result.stdout.decode().strip('\n')
         if 'OSPL_HOME' not in os.environ:
             os.environ['OSPL_HOME'] = os.path.join(os.environ['HOME'], 'opensplice', 'HDE', 'x86_64.darwin10_clang')
+        # TODO(wjwwood): remove this when qt5 is linked on macOS by default
+        # See: https://github.com/Homebrew/homebrew-core/issues/8392#issuecomment-334328367
+        os.environ['CMAKE_PREFIX_PATH'] = os.environ.get('CMAKE_PREFIX_PATH', '') + ':/usr/local/opt/qt'
```
nuclearsandwich: It's a minor nit as this script is inherently macos specific but above `os.pathsep` is used rather than the literal ':'.

wjwwood: No worries, it's an easy fix :+1:

PR #118, opened 2017-12-12
Add info for configuring process supervision with launchd.

nuclearsandwich:
I tried once before to alleviate the pain of disconnecting MacOS nodes by enabling process supervision but the supervised process ran more poorly than the manually started one and this change was reverted. (See https://github.com/ros2/build_cop/issues/10)

When I reimaged Dosa I did a bit more digging into the launchd docs and found some flags that should help prioritize the supervised processes. Dosa has been running with supervision enabled for the past couple weeks and build times are stable.

With supervision dosa is able to bounce back online as soon as its able as opposed to the other two nodes which currently stay offline until someone manually restarts them.

Once the files here get reviewed, I think it would be worth deploying this to mini1 or mini2 for another few days to make sure there are no performance regressions for them, rather than rolling this out to both at once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,4 +69,4 @@ The private submodule is still pulled on OS X and Linux.
 
 ### Notes about MacOS Agents
 
-Configuration details for running the Jenkins agent process via launchd can be found in the [configuration](https://github.com/ros2/ci/tree/confuguration) branch of this repository.
+Configuration details for running the Jenkins agent process via launchd can be found in the [configuration](https://github.com/ros2/ci/tree/configuration) branch of this repository.
--- a/README.md
+++ b/README.md
@@ -66,3 +66,7 @@ Repositories which have this branch will be switched to it, others will be left 
 
 I had trouble setting up the ssh-agent on Windows, so I disabled the submodule init on Windows which removes the need for cloning from private repositories and there removes the need for ssh keys and the ssh agent.
 The private submodule is still pulled on OS X and Linux.
+
+### Notes about MacOS Agents
+
+Configuration details for running the Jenkins agent process via launchd can be found in the [configuration](https://github.com/ros2/ci/tree/confuguration) branch of this repository.


nuclearsandwich:
Requested review from @mikaelarguedas as the current buildfarmer and @dhood as the one who had to revert my last attempt at doing this because of the perf issue.

macos_resources/jenkins-agent.plist
```diff
@@ -0,0 +1,29 @@
...
+<plist version="1.0">
+<dict>
+	<key>Label</key>
+	<string>org.ros2.ci.REPLACE_HOSTNAME-agent</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>/usr/bin/java</string>
+		<string>-Djava.awt.headless=true</string>
+		<string>-jar</string>
+		<string>/Users/osrf/jenkins-agent/slave.jar</string>
```
mikaelarguedas: I suppose this can't expand env vars by itself ? If it cant, maybe we could also make this a field to replace (REPLACE_HOME) so that it's easily reusable by us (if we rename our users `jenkins` for example) and others. Then I understand you made modular only the machine specific parts and were not aiming to make everything modular so I'm totally fine not acting on this or the CI URL in this PR.

nuclearsandwich:
Thanks for correcting the typos. I'm having a tough time typing today.

macos_resources/jenkins-agent.plist
nuclearsandwich: Yeah, we can't use environment variables in the argument list. I kept things that were currently universal hard-coded, but I have no problem adding more steps to the setup process to make it more universally applicable.

Happy to make the changes if they're desired.

mikaelarguedas: I'd say that fine for now to keep the replace steps to what is strictly necessary. If anybody asks, it's easy to add later on.

macos_resources/jenkins-agent.plist
```diff
@@ -0,0 +1,29 @@
...
+		<string>-jnlpUrl</string>
+		<string>http://ci.ros2.org/computer/REPLACE_JENKINS_AGENT_NAME/slave-agent.jnlp</string>
+		<string>-secret</string>
+		<string>REPLACE_JENKINS_AGENT_SECRET</string>
+	</array>
+	<key>KeepAlive</key>
+	<true/>
+	<key>ProcessType</key>
```
dhood: is it correct to say that "The process is set to `Interactive` type and given high scheduling priority so that when ROS 2 tests are run those related to performance (timing etc) do not suffer."? Both are necessary for that right? If so I'll add that as a comment for whoever finds this without the context of this PR.

dhood:
The content looks fine to me (I don't have heaps of experience with launchd to critically evaluate the plist), but I'm not sure that this is the right "place". I know that you added it here so it could be version controlled, but everything else in this repo is used on the buildfarm itself (e.g. the linux resources are used for running jobs) while this is one particular step of node configuration (and the rest of the steps are elsewhere).

On the [roswiki repo](https://github.com/ros-infrastructure/roswiki/tree/conf) there's a separate branch for configuration stuff that we want version controlled. What do you think about this living on a different branch or just in a gist?

macos_resources/jenkins-agent.plist
nuclearsandwich: > "The process is set to Interactive type and given high scheduling priority so that when ROS 2 tests are run those related to performance (timing etc) do not suffer

The process type is set to interactive so that the supervised process is afforded the same priority as the previous manually started agents.
The niceness value is really just for emphasis that this is process should be allowed to run with minimal hinderance on our build machines. It may not be necessary (manually spawned processes have the default niceness).

If we wanted performance tests to be representative of default background task behavior on MacOS we could drop these configuration keys and instead treat the resulting performance hit and possible perf-test failures as bugs.


> Both are necessary for that right?

nuclearsandwich:
> I'm not sure that this is the right "place"

Nor am I.

If we had config management for the Ci hosts it would belong there.

I'm fine with closing the PR and dumping this in a separate branch or even just sticking it in a gist, as my earlier revisions were gists, but this conversation (and future ones) would be very challenging over a gist.

mikaelarguedas:
I don't have strong feelings about where this should live. I think it's better to have them in a rpeo than a gist. This repo seems fine until we have a better place (a deployment repo with config files / scripts for CI nodes management).
As of if it should live on the default branch of elsewhere I think both are fine. Having them on the default branch gives them more visibility and I don't see a strong drawback to having them there.

Maybe renaming the subfolder to avoid confusion would make sense (regardless of the branch it lands on).

nuclearsandwich:
If anyone has strong feelings about the subdirectory name, I don't mind any direct changes. `macos_configuration` is the only thing that comes to mind from me.

nuclearsandwich:
@dhood my meeting presence was cut short by hardware failure so I didn't get a chance to ask you. I think you had the strongest thoughts on where and how this should land. I'm happy to retarget this to a new `configuration` branch or anywhere else for that matter.

dhood:
yep, moving to `configuration` branch and referencing it from a new step in the OS X node setup instructions sounds good to me

nuclearsandwich:
The orphan configuration branch now exists with the contents of this PR. https://github.com/ros2/ci/tree/configuration

dhood:
could you propagate the additional commits from this PR please

nuclearsandwich:
> could you propagate the additional commits from this PR please

whoops. Yes. Forgot to pull changes before cherry picking. It is up to date now.

nuclearsandwich:
What do we want to do next here? I had proposed rolling the change out to the next two hosts one at a time to make sure that Dosa isn't a fluke.

dhood:
yep that seems appropriate to me too

dhood:
we can add that to the rolling buildfarmer list of tasks so it can get prioritised with respect to other tasks

nuclearsandwich:
mini1 has been updated to use launchd and a test build is running to verify timing info: [Build Status](http://ci.ros2.org/job/test_ci_osx/44/)

nuclearsandwich:
All three MacOS hosts are now using supervision. Since the configuration docs themselves are stored in a separate branch, this PR just contains a commit linking to that branch from the readme.

nuclearsandwich:
> typo

Had to be one. Thanks for the catch.

PR #120, opened 2018-01-03
use comment in em

dirk-thomas:
Follow up of #101.

Instead of using an xml comment the block should be commented out in the template so that it never gets send to Jenkins.


dirk-thomas:
Deployed on ci.ros2.org.

PR #124, opened 2018-02-14
Configuration: Add Windows notes and switch to swarm client.

nuclearsandwich:
I added a section for configuring our Windows nodes as services. We were already doing this but the documentation did not match the found implementation.

As part of readying the CI host migration I switched CI to use the Swarm client plugin like the ROS buildfarm does. In addition to operational consistency increasing my sanity, this change also makes it easier to scale up CI nodes on demand (though it isn't quite as easy as the buildfarm yet).

### DIFF
--- a/macos/README.md
+++ b/macos/README.md
@@ -9,6 +9,7 @@ For our CI hosts we consider the jenkins agent process to be the most critical t
 
 The agent plist was written when dosa was reimaged December 2017 and updated to use the Swarm client in Feb 2018.
 It expects that the directory `$HOME/jenkins-agent` exists and that it contains the Jenkins agent program `swarm-client-3.8.jar`.
+If you need the swarm client, which when combined with the [Swarm plugin](https://plugins.jenkins.io/swarm) allows for easier addition and removal of temporary Jenkins nodes, you can get it from [this repository](https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/swarm-client/).
 This directory will also contain the agent `stdout.log` and `stderr.log` streams from launchd.
 
 The file contains a few fields that must be updated for the specific CI host.
--- a/macos/README.md
+++ b/macos/README.md
@@ -7,8 +7,8 @@ For our CI hosts we consider the jenkins agent process to be the most critical t
 
 ## Preparing the Jenkins agent property list
 
-The agent plist was written when dosa was reimaged December 2017.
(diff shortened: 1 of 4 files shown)


windows/README.md
```diff
@@ -0,0 +1,34 @@
+# Windows Resources
+
+Windows agents can be wrapped with [winsw][] to run as Windows services.
```
clalancette: Missing URL?

windows/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Preparing the Jenkins agent
+
+To set up a Windows agent requires downloading the [winsw .NET4 executable](https://github.com/kohsuke/winsw/releases/).
+To keep paths as short as possible, we use the directory `C:\J` as our Jenkins home directory.
```
clalancette: Maybe add a little bit more explanation on why this is desirable, i.e. Windows doesn't like really long pathnames.

macos/README.md
```diff
@@ -7,8 +7,8 @@ For our CI hosts we consider the jenkins agent process to be the most critical t
 
 ## Preparing the Jenkins agent property list
 
-The agent plist was written when dosa was reimaged December 2017.
-It expects that the directory `$HOME/jenkins-agent` exists and that it contains the Jenkins agent program `slave.jar`.
+The agent plist was written when dosa was reimaged December 2017 and updated to use the Swarm client in Feb 2018.
+It expects that the directory `$HOME/jenkins-agent` exists and that it contains the Jenkins agent program `swarm-client-3.8.jar`.
```
clalancette: What exactly is the swarm client?  Where can I get it?  Some documentation and/or links would be helpful to someone like me who isn't following all of this stuff.

windows/README.md
nuclearsandwich: This is actually the markdown syntax for a footnote style link. the empty second bracket indicates that the link text is also the footnote key. The URL is on line 33.

macos/README.md
nuclearsandwich: f181e7b4bf8106704d94e72380f2ffe8fc6109fd adds a sentence about the swarm client jar to the macos docs and links to it in both docs. Happy to add more if you think it's valuable to do so. I've been enmired in this for a bit so I trust your judgement as to what explanations need improvement.

clalancette: Excellent, thanks.

PR #125, opened 2018-03-06
Raise KeyError if ROS_DOMAIN_ID is unset

sloretz:
Follow up to discussion in ros2/build_cop#89. This PR makes a CI job fail if the machine does not have the ROS_DOMAIN_ID environment variable set.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -345,6 +345,10 @@ def run(args, build_function, blacklisted_package_names=None):
         # printing ANSI color codes on Windows.
         os.environ['ConEmuANSI'] = 'ON'
 
+    # ROS_DOMAIN_ID must be unique to each CI machine on a network to avoid crosstalk
+    if 'ROS_DOMAIN_ID' not in os.environ:
+        raise KeyError('ROS_DOMAIN_ID environment variable must be set')
+
     info("Using workspace: @!{0}", fargs=(args.workspace,))
     # git doesn't work reliably inside qemu, so we're assuming that somebody
     # already checked out the code on the host and mounted it in at the right
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -345,10 +345,6 @@ def run(args, build_function, blacklisted_package_names=None):
         # printing ANSI color codes on Windows.
         os.environ['ConEmuANSI'] = 'ON'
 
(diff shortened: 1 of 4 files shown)


sloretz:
CI 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4025/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1119/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3352/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4116/)

sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4134/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1212/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3414/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4239/)

sloretz:
CI again
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4135/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1213/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3415/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4240/)

sloretz:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3418/)

sloretz:
@mikaelarguedas I'm not sure if it currently sets ROS_DOMAIN_ID on the linux nodes, but I think I remember @nuclearsandwich saying provisioning currently can't set unique ROS_DOMAIN_ID.

nuclearsandwich:
> is it that our provisioning doesn't set the domain IDs and we want to keep them working? if the provisioning does take care of it already, maybe we should not allow same domain ids in that case either

The linux machines are all provisioned using the same base attributes and using chef in a standalone configuration. In order to easily allow per-host variables we would need to set up some way for hosts to query for what their unique attributes should be (the most logical thing to use is a chef-server but there are other options if we want). The linux hosts do all exist in the same subnet and virtual private network within AWS so it is possible for them to collide. One thing we could try doing is adding a script that plucks the host's internal IP address and sets the ROS_DOMAIN_ID to the value of the final octet. This would keep them unique within the local network while we figure out what to do long-term.

PR #131, opened 2018-03-20
use newer Debian of libopensplice67

dirk-thomas:
This build is using the new Debian package `libopensplice67_6.7.0+osrf2-3~xenial_amd64.deb`: [Build Status](https://ci.ros2.org/job/ci_linux/4117/)

It addresses the missing config file in `etc` reported in ros2/rmw_opensplice#205.


mikaelarguedas:
I suggest to merge this right away as the build step passed and ci will be broken as soon as the cache is invalidated given that the previous deb has been deleted from the repo.

dhood:
For future readers: this was re-applied in https://github.com/ros2/ci/pull/138

PR #132, opened 2018-03-21
invoke colcon instead of ament_tools

dirk-thomas:
I don't intended to merge this patch - it will stay "in progress" until the direction is decided. I will keep the branch up-to-date though to allow anyone to opt-in to use this on CI.

---

With this branch you can set `CI_SCRIPTS_BRANCH` to `colcon` in the job parameters to use `colcon` for the build. When using this branch you need to change the default value for `CI_AMENT_BUILD_ARGS` from:

> --parallel --cmake-args -DSECURITY=ON --

to

> --cmake-args " -DSECURITY=ON"

* Parallel builds are the default (`--executor parallel`).
* Since `colcon` uses `argparse` without custom parsing of greedy arguments the `-D` needs to be prefixed with a space so that `argparse` doesn't consider it an argument. The `--cmake-args` argument itself will then take care of removing the leading space. The terminating `--` is not needed anymore.

I also often use the option `--event-handler console_cohesion+` to print the output of each package "on block" once the package has finished. That makes sure the output is not interleaves but the build log still contains "all" the information.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4114/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1201/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3404/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4226/)
 * ~~built without `SECURITY=ON` since that require changing the handling of the job parameters~~ the PR has been updated to change the default parameters to keep security on the same way as before
  * only using FastRTPS since the RTI code generator keeps failing when building "many" packages in parallel
* Linux coverage [Build Status](https://ci.ros2.org/job/nightly_linux_coverage_colcon/3/)

---

Comparison of two Linux CI builds using different build tools:
* Using `ament_tools`: [1 hr 5 min](https://ci.ros2.org/job/ci_linux/4121/)
* Using `colcon`: [47 min](https://ci.ros2.org/job/ci_linux/4123/) (27% faster)


mikaelarguedas:
Not stating an opinion here but I think it would be more fair to compare build time instead of the time including repeating flaky tests:

ament build time: 00:21:30
colcon build time: 00:20:04 (slightly faster)

dirk-thomas:
> Not stating an opinion here but I think it would be more fair to compare build time instead of the time including repeating flaky tests:

With "build" time I was referring to the time the Jenkins build takes. I would expect both builds to have a similar rate of flakiness and would be surprised if they preferred happen in the `ament_tools` one. The overall result is still a significant speed up - I haven't looked into where the time difference originates from...

dirk-thomas:
> Is this able to build connext without crashing on the code generator ?

At least the builds I ran on Jenkins didn't had that problem. But since the problem happened only very rarely in the first place I don't think I can answer that question after just a few builds.

> I see that the windows job is not using connext

I triggered another Windows build with Connext: [Build Status](https://ci.ros2.org/job/ci_windows/4293/)

mikaelarguedas:
I think I'm missing information to review this so forgive if the following questions have been answered elsewhere and are planned to be addressed (pointers appreciated)

If it's not desired to have the console output displayed by default in colcon. Is there a plan to add `--event-handler console_cohesion+` or at least the error output to the jobs by default?

What is the plan for `DSECURITY` ? Do you plan on having the jobs fetch the meta information from somewhere to add the flag for `fastrtps` and `test_security` ? Or should the default values of the jobs also be modified accordingly?

Why is `catkin_pkg` required? Does this mean that we don;t need ament_package anymore?

dirk-thomas:
> If it's not desired to have the console output displayed by default in colcon. Is there a plan to add `--event-handler console_cohesion+` or at least the error output to the jobs by default?

Yes, we could make that the default. I can add more commits to this PR - I was mostly worried when also altering the job generation (to use different defaults and change the naming of some of the parameters) it might get too much for a review.

> What is the plan for `DSECURITY` ? Do you plan on having the jobs fetch the meta information from somewhere to add the flag for `fastrtps` and `test_security` ? Or should the default values of the jobs also be modified accordingly?

I would keep that separate (to not make a single PR address too many things at the same time) and keep the current default value passing it to all packages.

> Why is `catkin_pkg` required? Does this mean that we don;t need ament_package anymore?

`colcon` uses the API of `catkin_pkg` to parse the manifests and doesn't use `ament_package`. Either of the two would be sufficient. It just happens that the first is already available from Debian in a ROS distro independent way which isn't the case for the later.

mikaelarguedas:
> I would keep that separate (to not make a single PR address too many things at the same time) and keep the current default value passing it to all packages.

So if we merge this we don't build or test security anymore, is that correct ?

> I can add more commits to this PR - I was mostly worried when also altering the job generation (to use different defaults and change the naming of some of the parameters) it might get too much for a review.

I understand the concern for the review, nicely crafted commits should allow to reduce the review burden. I we want to convince ourselves about the job generation, what about a set of test jobs generated with the new config and manually triggered ?


I think that the behavior of the jobs should be the same after merging this PR, so not having console output or not building/testing part of our stack is a blocker for merging this in my opinion.

mikaelarguedas:
> Currently the default job parameters contain -DSECURITY=ON. The same will be the case after this change.

I must have misunderstood the description of this PR then "built without SECURITY=ON since that require changing the handling of the job parameters"

> I updated the PR to change the default parameters to match the same behavior for e.g. output handling

:+1: , without that change our jobs would have been broken: https://ci.ros2.org/job/ci_linux/4197/

mikaelarguedas:
> I intentionally didn't include any other job configuration changes in this PR. So I don't think we need to generate any temporary jobs. You can simply run trigger the current job using this branch and using the default values from this patch:
> `--cmake-args " -DSECURITY=ON" --event-handler console_cohesion+`

I tried to run a job with these arguments and the job crashed on argument parsing: https://ci.ros2.org/job/ci_windows/4301/

dirk-thomas:
This is now colliding with the special handling from #86. I removed all of that logic in dc6cc3a2c44efae3bf60dfe7f26c8bb439cff41b which now requires separate jobs :disappointed::
* https://ci.ros2.org/view/All/job/colcon_ci_linux/
* https://ci.ros2.org/view/All/job/colcon_ci_linux-aarch64/
* https://ci.ros2.org/view/All/job/colcon_ci_osx/
* https://ci.ros2.org/view/All/job/colcon_ci_windows/

dirk-thomas:
For the latest builds see https://ci.ros2.org/view/colcon/

ros2_batch_job/__main__.py
```diff
@@ -237,58 +254,51 @@ def process_coverage(args, job):
...
-        '--build-tests',
-        '--build-space', '"%s"' % args.buildspace,
-        '--install-space', '"%s"' % args.installspace,
-    ] + (['--isolated'] if args.isolated else []) +
-        (
-            ['--cmake-args', '-DCMAKE_BUILD_TYPE=' +
-                args.cmake_build_type + ' --']
+        args.colcon_script, 'build',
+        '--base-paths', '"%s"' % args.sourcespace,
+        # '--build-tests',
```
mikaelarguedas: This can be removed now that we pass BUILD_TESTING to cmake

dirk-thomas:
Default CI builds:
* Linux [Build Status](https://ci.ros2.org/job/colcon_ci_linux/14/)
* Linux aarch [Build Status](https://ci.ros2.org/job/colcon_ci_linux-aarch64/4/)
* macOS [Build Status](https://ci.ros2.org/job/colcon_ci_osx/6/)
* Windows [Build Status](https://ci.ros2.org/job/cc_ci_win/18/)
* Linux coverage: [Build Status](https://ci.ros2.org/job/colcon_linux_coverage/3/)

Default packaging builds (FastRTPS):
* Linux [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_linux/11/)
* Linux aarch [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_linux-aarch64/8/)
* macOS [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_osx/14/)
* Windows [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_windows/8/)

Packaging builds (OpenSplice):
* Linux [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_linux/10/)
* Linux aarch [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_linux-aarch64/9/)
* macOS [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_osx/13/)
* Windows [Build Status](https://ci.ros2.org/job/colcon_ci_packaging_windows/9/)

With the recent changes and the latest "green" results (except the hanging Linux builds) I am putting this back in review.

create_jenkins_job.py
```diff
@@ -188,7 +188,7 @@ def create_job(os_name, job_name, template_file, additional_dict):
             'cmake_build_type': 'None',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
-            'ament_test_args_default': '--retest-until-fail 20 --ctest-args -LE linter --',
+            'test_args_default': '--retest-until-fail 20 --ctest-args -LE linter --',
```
mikaelarguedas: My understanding was that any arg not known by argparse should be escaped (in this case `--ctest-args ' -LE' linter`)
Did you test this set of argument on one of the jobs?

ros2_batch_job/__main__.py
```diff
@@ -237,58 +254,65 @@ def process_coverage(args, job):
...
     print('# END SUBSECTION')
     if ret_test:
         return ret_test
 
-    print('# BEGIN SUBSECTION: ament test_results')
+    print('# BEGIN SUBSECTION: colcon test-result')
     # Collect the test results
     ret_test_results = job.run(
-        ['"%s"' % job.python, '-u', '"%s"' % ament_py, 'test_results', '"%s"' % args.buildspace],
+        [args.colcon_script, 'test-result', '--build-base', '"%s"' % args.buildspace, '--all'],
```
mikaelarguedas: It is pretty inconvenient to have to open the full console log just to see the summary of the test results. Could we either not pass `--all` or print the summary after the list of test result files?
Ideally I'd like to have both (not passing all and displaying summary after rather than before) as this provides us the same behavior as ament that I find more convenient

mikaelarguedas: https://github.com/colcon/colcon-test-result/pull/2

create_jenkins_job.py
dirk-thomas: Fixed in b2c5358ec1b0840e8cb2f4d730d86fb4d7885875.

ros2_batch_job/__main__.py
dirk-thomas: Moving the summary below the individual results sounds good to me.

Imo the list of all results is still valuable. There is no other easy way to see which tests ran otherwise. Therefore I would prefer keeping `--all`.

If you think moving the summary below isn't sufficient for readability we could consider to print the test results once with `--all` (for the complete list) and once without (for a brief list of results with errors/failures).

mikaelarguedas: I guess it's pretty subjective. I find it more valuable to have a quick summary of what failed directly at the end of the job and to go to the Jenkins test interface if I want the full detail of tests (passing or failing).
But I guess your point is that the jenkins interface already provides an easy way to see failed tests separated from the ran tests? that makes the segregation in the console output less useful/redundant ?

I would be in favor of running the command twice then as it wouldn't impact CI time much and provide easily redable information as (as a developer) I care more often about what failed than what has been ran.

dirk-thomas: Ok: 4030b517f62c4f49bb358fc5dca9591509bbf612

dirk-thomas:
Rebased to fix conflicts with the recently merged PR.

ros2_batch_job/__main__.py
```diff
@@ -402,6 +438,11 @@ def run(args, build_function, blacklisted_package_names=None):
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
         job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U'] + pip_dependencies, shell=True)
+        if sys.platform != 'win32':
+            colcon_script = os.path.join(venv_path, 'bin', 'colcon')
+        else:
+            colcon_script = 'c:\\python36\\Scripts\\colcon.exe'
```
mikaelarguedas: Does this mean that we will now need to update the ci scripts on new python minor version?
Could we infer that from the environment ?

dirk-thomas: If that can be done automatically even better.

dirk-thomas: Fixed in 4fe9cfdefb4ae828ce8599a6d4ef8318be5c27cf.

job_templates/packaging_job.xml.em
```diff
@@ -202,47 +186,42 @@ echo "# BEGIN SECTION: Run packaging script"
 echo "# END SECTION"
 @[end if]@
 @[elif os_name == 'windows']@
+setlocal enableDelayedExpansion
 rmdir /S /Q ws workspace
 
 echo "# BEGIN SECTION: Determine arguments"
-set "PATH=%PATH:"=%"
+set "PATH=!PATH:"=!"
```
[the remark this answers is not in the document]
dirk-thomas: Using `%` didn't work for me. I didn't spend too much effort trying to figure out why. Using `!` made it work.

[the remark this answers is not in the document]
mikaelarguedas: ok, I was wondering if this didnt work *before* you enabled the delayed expansion but would have worked *after* enabling it.

That's fine, I'm not planning on looking into it to find out

ros2_batch_job/packaging.py
```diff
@@ -24,33 +24,37 @@
 
 
 def build_and_test_and_package(args, job):
-    print('# BEGIN SUBSECTION: ament build')
+    print('# BEGIN SUBSECTION: build')
```
mikaelarguedas: this is "BEGIN SUBSECTION: colcon build" in the other file. I suggest homogenizing by just using "build", "test" and "test-result" everywhere

dirk-thomas: Done in 78d6cab9de6dcd9ca098875a175f78f6604e2975.

mikaelarguedas:
This looks good to me now. :+1: 

Not sure if we prefer to roll this out before the weekend to have the time to see all jobs types running with it before Monday or if on the opposite we want to be around to monitor the jobs after deploying this

dirk-thomas:
I am planning on merging this in an hour so that the nightlies can run and there is time over the weekend to fix minor issues. Imo that is preferred over potentially disturbing everyone during the week. If the problems are too big (or my time too short) I will just revert it before the nightlies Sunday evening and will try again some time later... Let me know if you have any objections.

dirk-thomas:
The first round of manually triggered builds (after reconfiguring the jobs) look promising:

CI builds:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/4277/)
* Linux aarch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1288/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/3516/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/4361/)

Packaging builds:
*  Linux [Build Status](https://ci.ros2.org/job/ci_packaging_linux/81/)
* Linux aarch [Build Status](https://ci.ros2.org/job/ci_packaging_linux-aarch64/15/)
* macOS [Build Status](https://ci.ros2.org/job/ci_packaging_osx/26/)
* Windows [Build Status](https://ci.ros2.org/job/ci_packaging_windows/22/)

dirk-thomas:
To summarize the current state:

* all `CI` builds are green
* all `packaging` builds are green
* the `nightly` builds:
  * all `release` builds are green
  * the `debug` builds:
    * Linux, Linux aarch are green
    * macOS: the [26 test failures](https://ci.ros2.org/view/nightly/job/nightly_osx_debug/829/testReport/) are all related to rviz rendering and plugins
    * Windows: the [4 test failures](https://ci.ros2.org/view/nightly/job/nightly_win_deb/859/testReport/) are from the `test_showimage` timing out
  * the `repeated` builds:
    * Linux: the [20 test failiures](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1003/) are composition tests, `test_parameter_server_cpp`, `test_tutorial_parameter_events`
    * Linux aarch: the [8 test failures](https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/397/testReport/) are composition tests
    * macOS: the [10 test failures](https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/1082/) are composition, `test_services_cpp`, `rclcpp.TestNode.now`
    * Windows: the [2 test failures](https://ci.ros2.org/view/nightly/job/nightly_win_rep/1058/) are from the `test_parameter_server_cpp`

Neither of the failing tests strikes me as likely being related to this patch.

mikaelarguedas:
Few additions that may help us segregate the failures:

It indeed looks like the various fixes in testing and packaging yesterday reduced the number of failures

> macOS: the 26 test failures are all related to rviz rendering and plugins

:+1: from today's meeting, @nuclearsandwich is about to disable them 

> Windows: the 4 test failures are from the test_showimage timing out

:+1: not related to this patch, these failures are due to OpenCV libraries not being installed properly on the build machines:
see https://github.com/ros2/ci/pull/146#issuecomment-381278308 and https://github.com/ros2/build_cop/issues/113 for details

> all packaging builds are green

Packaging jobs are green, but the packaging artifacts are still unusable ATM. @dhood has been investigating and @mjcarroll @sloretz are testing the ones from #158 to see if the resulting workspaces can be source and used as underlays

dirk-thomas:
> Users of nightlies also haven't been informed of the switch that has occurred.

I couldn't find anywhere in the docs where we point users to the nightly packaging jobs. Do you have a pointer for me? I am more than happy to add a note for the changed script name.

PR #133, opened 2018-03-21
Use --rm on docker run invocation

dhood:
To prevent having to remove old containers manually. Our linux nodes run out of disk space regularly because of a build up of containers.

Will give this a test on the farm in case it has any unintended side effects

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -229,7 +229,7 @@ export CONTAINER_NAME=ros2_batch_ci_aarch64
 @[else]@
 @{ assert False, 'Unknown os_name: ' + os_name }@
 @[end if]@
-docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache $CONTAINER_NAME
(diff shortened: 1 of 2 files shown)


dhood:
ps I'm aware that this doesn't resolve the buildup of old images.

mikaelarguedas:
This also prevents introspection of containers on job failure as they will disappear instantly.
Integrating something like this [cleanup docker images script](https://github.com/ros-infrastructure/buildfarm_deployment/blob/2a3d6b409c0e85336ecbcd6e7a3663f913872e27/modules/agent_files/files/home/jenkins-agent/cleanup_docker_images.py) would be more appropriate. (it would likely need to be extended to prune all old containers and not only the ones tied to old images)

nuclearsandwich:
> Integrating something like this cleanup docker images script would be more appropriate.

I think we're going to move new deployments to scheduling `docker image prune -f` once a week. That removes "dangling" images, it won't remove images from running or stopped containers. Which means that we have the chance to investigate images for reproduction even if the container is immediately removed.

If we schedule the weekly image removal for Thursday nights, that gives us nearly the full week to investigate issues that crop up on the weekend (Friday and Saturday) nightlies before those images get pruned.

mikaelarguedas:
I pointed it out because that was the conclusion last time we discussed that approach.
I don't remember ever introspecting containers on the build machines so I'm personally in favor of this approach in addition to pruning the dangling images periodically. Just wanted to make sure that we now all agree that the usefulness of having them around is outweighed by the space they eat up on the machines.

dhood:
CI on a linux test job using `--rm`: [Build Status](https://ci.ros2.org/job/test_ci_linux/22/)

merging and will comment when it's deployed

dhood:
deployed

PR #136, opened 2018-03-22
Add ROS_DOMAIN_ID env var

sloretz:
Update template to add setting the environment variable ROS_DOMAIN_ID before launching the jenkins agent on OSX nodes.

connects to ros2/ci#125

### DIFF
--- a/macos/README.md
+++ b/macos/README.md
@@ -22,6 +22,7 @@ They are documented here
 - `REPLACE_AGENT_DESCRIPTION`: The description fof the agent in the Jenkins UI.
 - `REPLACE_USERNAME`: The Jenkins username of the user with node creation privileges.
 - `REPLACE_PASSWORD`: The password (usually a GitHub auth token for us) for the above user.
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
Are there any other variables we should set in this fashion? the `OSPL_HOME` variable comes to mind. We set a default one if none is set so maybe that's preferred. https://github.com/ros2/ci/blob/da3575dd50932720c6c89053be9ead7e32b7c3f4/ros2_batch_job/osx_batch/__init__.py#L48

It makes sense to abort if the domain id is missing, should we warn if the default OSPL_HOME is used or just be happy?

sloretz:
The three environment variables set in .bashrc on mini2 are ROS_DOMAIN_ID, PATH (adds `/usr/local/opt/ccache/libexec`), and OSPL_HOME. It think it's painless to print warning if they're not set, and some future problem might be easier to solve.

https://github.com/ros2/ci/blob/da3575dd50932720c6c89053be9ead7e32b7c3f4/ros2_batch_job/osx_batch/__init__.py#L31-L37

PR #137, opened 2018-03-22
Warn when defaults for env vars are used

sloretz:
From comments on #136, this PR adds warnings when environment variables aren't set on OSX. Windows and linux don't appear to set these environment variables. ROS_DOMAIN_ID isn't touched because #125 aborts the job if it's not set.

### DIFF
--- a/ros2_batch_job/osx_batch/__init__.py
+++ b/ros2_batch_job/osx_batch/__init__.py
@@ -46,7 +46,9 @@ class OSXBatchJob(BatchJob):
                 ['brew', '--prefix', 'openssl'],
                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             if not brew_openssl_prefix_result.stderr:
-              os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix_result.stdout.decode().strip('\n')
+                os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix_result.stdout.decode().strip('\n')
+            else:
+                raise KeyError('Failed to find openssl')
         if 'OSPL_HOME' not in os.environ:
             warn('OSPL_HOME not set; using default value')
             os.environ['OSPL_HOME'] = os.path.join(os.environ['HOME'], 'opensplice', 'HDE', 'x86_64.darwin10_clang')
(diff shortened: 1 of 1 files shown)


sloretz:
CI
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3419/)

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -36,16 +36,19 @@ def pre(self):
         else:
             warn('ccache does not appear to be installed; not modifying PATH')
         if 'LANG' not in os.environ:
+            warn('LANG unset; using default value')
             os.environ['LANG'] = 'en_US.UTF-8'
         if 'ROS_DOMAIN_ID' not in os.environ:
             os.environ['ROS_DOMAIN_ID'] = '111'
         if 'OPENSSL_ROOT_DIR' not in os.environ:
+            warn('OPENSSL_ROOT_DIR not set; installing openssl')
```
mikaelarguedas: I think this is just **finding** openssl installed from brew but not performing the installation.

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -36,16 +36,19 @@ def pre(self):
...
             brew_openssl_prefix_result = subprocess.run(
                 ['brew', '--prefix', 'openssl'],
                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             if not brew_openssl_prefix_result.stderr:
               os.environ['OPENSSL_ROOT_DIR'] = brew_openssl_prefix_result.stdout.decode().strip('\n')
```
mikaelarguedas: maybe we should warn or raise in the `else` case though (openssl env var not set AND not found using `brew --prefix`)

ros2_batch_job/osx_batch/__init__.py
sloretz: Fixed text
1986152

ros2_batch_job/osx_batch/__init__.py
sloretz: I made it raise in 66980c6 because I assume this is a case that is not expected to happen in practice.

mikaelarguedas: :+1: I think the build would fail if it's not set, so better fail early

sloretz:
CI still waiting for an executor, so It should have the latest commits when it runs
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3419/)

PR #138, opened 2018-03-23
Re-apply PRs 131 and 134

dhood:
These had been merged but then an accidental force-push to master occurred with a branch in which they had not been merged. @dirk-thomas FYI.

Failing nightly that tipped me off: [Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/382/). Otherwise this probably would have gone unnoticed and would have led to a lot of confusion further down the line!


dirk-thomas:
Sorry for the mess.

dhood:
I re-enabled the "merge commit" option like we have on our other repos.

dhood:
Depends how interesting you want to make the build cop's job :smile: It was a nice puzzle to start the morning with. Imagine if you hadn't have mentioned the force push yesterday!

dirk-thomas:
The least I can do is take the blame publically :bowing_man: :wink:

PR #141, opened 2018-04-03
[configuration] Modify node labels to not use _slave

dhood:


nuclearsandwich:
> I'm assuming you've looked into that and they are truly redundant.

Yeah, I added the updated labels (dropped _slave and changed osx to macos) when I migrated the linux hosts to chef and the windows and mac hosts to the swarm client with the intention that I'd open a PR like #142 afterwards. The linux machines were already using the label `linux` and so need no changes.

dhood:
merged but not deployed

PR #142, opened 2018-04-03
Modify node labels to not use _slave

dhood:
As I understand it this swap was a WIP by @nuclearsandwich so the nodes already have both labels. As such I don't think there are any issues with swapping this over.

Then we can remove the old label from the nodes, see https://github.com/ros2/ci/pull/141


dhood:
Realised that there are doc changes needed for this too, on their way.

dhood:
deployed

PR #144, opened 2018-04-11
Opensplice version update

rohitsalem:
Updated the linux dockerfile  with latest version of opensplice deb

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -41,7 +41,7 @@ RUN pip3 install -U setuptools pip virtualenv
 RUN apt-get update && apt-get install --no-install-recommends -y gcovr
 
 # Install the OpenSplice binary from the OSRF repositories.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The patch looks good to me. Can you please link to a job which uses this branch.

rohitsalem:
Here : [Build Status](https://ci.ros2.org/job/ci_linux/4196/)

mikaelarguedas:
From the job: `Get:1 http://packages.osrfoundation.org/gazebo/ubuntu xenial/main amd64 libopensplice67 amd64 6.7.180404+osrf1-1~xenial [3,278 kB]`

I'm assuming that the test failures are due to your changes and not the new binary. Is that correct ?
Do you happen to have a job with the new binary and without your changes to confirm it ?

rohitsalem:
Yes, the test failures are due to my changes, No, I ran only a single job with this new binary, I will start a new job with the new binary, with master and let you know.

rohitsalem:
New job [Build Status](https://ci.ros2.org/job/ci_linux/4199/)

PR #146, opened 2018-04-11
use Connext 5.3.1 and Visual Studio 2017

dirk-thomas:
For Windows I didn't bother to install the VS 2015 version but go for the VS 2017 version right away.

### DIFF
--- a/linux_docker_resources/rti_web_binaries_install_script.py
+++ b/linux_docker_resources/rti_web_binaries_install_script.py
@@ -36,11 +36,18 @@ def install_connext(installer_path, install_directory):
         child.sendline(install_directory)
         child.expect_exact('Do you want to continue? [Y/n]:')
         child.sendline('y')
-        result_index = child.expect_exact([
-            'Create an RTI Launcher shortcut on the Desktop [y/N]: ', pexpect.EOF], timeout=120)
-        if result_index == 0:
-            child.sendline('n')
-            child.expect(pexpect.EOF)
+        while True:
+            result_index = child.expect_exact([
+                pexpect.EOF,
+                'Disable copying of examples to rti_workspace [y/N]: ',
+                'Create an RTI Launcher shortcut on the Desktop [y/N]: ',
+                ], timeout=120)
+            if result_index == 0:
+                return
+            elif result_index == 1:
+                child.sendline('y')
+            elif result_index == 2:
+                child.sendline('n')
 
     except (pexpect.TIMEOUT, pexpect.EOF):
         raise RuntimeError(
--- a/linux_docker_resources/rti_web_binaries_install_script.py
+++ b/linux_docker_resources/rti_web_binaries_install_script.py
@@ -48,6 +48,8 @@ def install_connext(installer_path, install_directory):
                 child.sendline('y')
             elif result_index == 2:
                 child.sendline('n')
+            else:
+                raise RuntimeError('Unexpected result index: %d' % result_index)
 
     except (pexpect.TIMEOUT, pexpect.EOF):
         raise RuntimeError(
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -52,7 +52,7 @@ RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-insta
 RUN pip3 install pexpect
 
 # Get and install the RTI web binaries.
-RUN cd /tmp && curl --silent https://s3.amazonaws.com/RTI/Bundles/5.3.0/Evaluation/rti_connext_dds_secure-5.3.0-eval-x64Linux3gcc5.4.0.tar.gz | tar -xz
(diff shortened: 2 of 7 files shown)


dirk-thomas:
@dhood Since you wrote the `rti_web_binaries_install_script.py` script to install Connext I was wondering if you could patch it to work for the new Connext version. Currently it fails on a new question: see https://ci.ros2.org/job/ci_linux/4200/console

dhood:
sure thing

dirk-thomas:
The commit 892373e4b91551b98de591f7bbd965d616d3bbe5 adds the bits and pieces from #122 to use VS 2017.

mikaelarguedas:
Lots of bug fixes in there :yum:
https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/manuals/connext_dds/RTI_ConnextDDS_CoreLibraries_ReleaseNotes.pdf

dirk-thomas:
I have installed Connext 5.3.1 on all running mac OS (mini1+2+3) and Windows machines (icecube+portable+windshield):

* macOS: [Build Status](https://ci.ros2.org/job/ci_osx/3482/)
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/4314/)

dhood:
@dirk-thomas it installs fine on linux after 479d6fb0c8dd7e74e2ef169ed736b4f953ef0d2c (and a necessary rebase): [Build Status](https://ci.ros2.org/job/ci_linux/4218/)

I'll pass this back over to you for QA.

Launched the osx and windows jobs including fastrtps looking for cross-rmw issues (I also haven't seen the osx failure above before so it may be a regression):
- os x: [Build Status](https://ci.ros2.org/job/ci_osx/3491/)
- windows: [Build Status](https://ci.ros2.org/job/ci_windows/4324/)
- linux: above includes fastrtps.

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -36,11 +36,18 @@ def install_connext(installer_path, install_directory):
...
+                pexpect.EOF,
+                'Disable copying of examples to rti_workspace [y/N]: ',
+                'Create an RTI Launcher shortcut on the Desktop [y/N]: ',
+                ], timeout=120)
+            if result_index == 0:
+                return
+            elif result_index == 1:
+                child.sendline('y')
+            elif result_index == 2:
+                child.sendline('n')
```
dirk-thomas: I assume the index should never be not 0, 1, or 2? For safety I would propose to still add:

```
else:
    assert False, 'Unexpected result_index: %d' % result_index
```

dhood: it'll return an index in the list or timeout (which is caught below), but there's no harm in adding it. How come not raise though? (not that we intend to run this optimised, but who knows.)

dirk-thomas: > How come not raise though?

That is exactly what the `assert` does.

dhood: not if run with -O, right? Again it's "not that we intend to run this optimised, but who knows." :smile:

dhood: => what I meant was that today we don't, but for some unrelated reason in the future we might set `PYTHONOPTIMIZE` or something

dirk-thomas:
The CI builds are green. Waiting for final review (preferably from multiple people since this not only bumps Connext but also Visual Studio).

mikaelarguedas:
I didn't follow this closely.
How does it work for future Ardent releases?
- Is the goal to keep VS2015 installed on all machines and to build next Ardent patches with VS2015? If so can you please create an ardent branch here that we can use to that effect?
- Do we also have 5.3.1 for VS2015 ? Or do future Ardent release need 5.3.0 to still be installed?

mikaelarguedas:
@dirk-thomas Just checking: Can you confirm that the following been installed on all Windows nodes?
- openCV compiled with VS2017
- connext 5.3.1 security plugins

dirk-thomas:
> Is the goal to keep VS2015 installed on all machines and to build next Ardent patches with VS2015?

Yes.

> If so can you please create an ardent branch here that we can use to that effect?

I will do that once this is merged (to branch as late as possible).

> Do we also have 5.3.1 for VS2015 ?

Yes, we do.

> Or do future Ardent release need 5.3.0 to still be installed?

Until someone take the initiative nothing changes - so Ardent releases can still use 5.3.0. If we see a reason to change that we can do that on the to-be-created ardent branch (which will require 5.3.1 for VS2015 to be installed on the build nodes.

dirk-thomas:
> openCV compiled with VS2017

That has already happened for #122 a while ago.

> connext 5.3.1 security plugins

Yes, of course.

mikaelarguedas:
Thanks for confirming :+1:

dirk-thomas:
@mikaelarguedas @nuclearsandwich Thank you for reviewing this. Can you please also review the referenced PR ros2/rmw_connext#283.

PR #147, opened 2018-04-12
install system version of liburdfdom-headers-dev

mikaelarguedas:
To be able to detect urdf include order issues on CI pending the resolution of https://github.com/ament/ament_cmake/issues/130


mikaelarguedas:
Note I didnt add a new layer for this as I'm not sure how close the the layer limit edge we are on the packet machines

dirk-thomas:
Please add a comment above the line why that is being added.

PR #150, opened 2018-04-18
display console output for colcon tests as well

mikaelarguedas:
Looking at the last colcon builds, I couldnt see the console output ensuring that the tests were actually being ran.
While this information is available in the test-results section, it is not as convenient (or complete) as jumping to the test <PACKAGE_NAME> section of the jenkins output

This targets #132


dirk-thomas:
I can regenerate the custom colcon jobs when this is merged.

mikaelarguedas:
> I can regenerate the custom colcon jobs when this is merged.

sounds good to me :+1:

PR #151, opened 2018-04-18
Add JNLP agent configuration notes for Mac and Windows.

nuclearsandwich:
When attempting to rule out issues with the swarm client I reverted the mac and windows nodes to the agent.jar and JNLP configurations they had previously, but kept using the process supervision mechanisms that I had deployed for the swarm clients.

Since there are some situations where we may prefer the JNLP agent, at least for now, I've documented the setup that I used.


macos/README.md
```diff
@@ -5,28 +5,48 @@ By default it seems that launchd de-prioritizes processes as "background" tasks
...
+```
+cd ~/jenkins-agent && wget https://ci.ros2.org/jnlpJars/agent.jar
+```
 
-The agent plist was written when dosa was reimaged December 2017 and updated to use the Swarm client in Feb 2018.
-It expects that the directory `$HOME/jenkins-agent` exists and that it contains the Jenkins agent program `swarm-client-3.8.jar`.
-The swarm client, when combined with the [Swarm plugin](https://plugins.jenkins.io/swarm) on the buildfarm, allows for easier addition and removal of temporary Jenkins nodes.
-It can be downloaded from [this repository](https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/swarm-client/).
 The agent must have at least Java JDK 8 installed.
+For best results use the same version as our Jenkins master, currently `1.8.0_162`.
```
dhood: maybe replace this with how to infer the version instead so it's less likely to get outdated.

nuclearsandwich: Excellent point! Updated in 4772aaa

dhood: cool thanks!

PR #152, opened 2018-04-20
Set default CI_ISOLATED parameter to false on Windows.

nuclearsandwich:
Jobs using isolation are starting to fail on Windows as we're coming up
against the maximum environment variable and input line lengths.

Since other platforms are still building with isolation we don't lose
too much coverage.

Adds a template variable for the default setting of the CI_ISOLATED
parameter and keeps it set to true by default (previously the default
was hard-coded) and sets it to false for Windows only. It's still
possible to request a build with isolation by changing the build
parameter when starting a new build.

Addresses https://github.com/ros2/build_cop/issues/111

<details>
<summary>Dry run output (skipped jobs omitted)</summary>

```
Updating job 'ci_windows' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Skipped 'ci_packaging_windows' because the config is the same (dry run)
Skipped 'packaging_windows' because the config is the same (dry run)
Updating job 'nightly_win_deb' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Updating job 'nightly_win_rel' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Updating job 'nightly_win_rep' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
```
</details>

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -79,6 +79,7 @@ def main(argv=None):
         'use_osrf_connext_debs_default': 'false',
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
+        'use_isolated_default': 'true',
         'ament_build_args_default': '--parallel --cmake-args -DSECURITY=ON --',
         'ament_test_args_default': '--retest-until-pass 10',
         'enable_c_coverage_default': 'false',
@@ -104,6 +105,7 @@ def main(argv=None):
         'windows': {
             'label_expression': 'windows',
             'shell_type': 'BatchFile',
+            'use_isolated_default': 'false',
         },
         'linux-aarch64': {
             'label_expression': 'linux_aarch64',
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -26,6 +26,7 @@
     use_osrf_connext_debs_default=use_osrf_connext_debs_default,
     use_fastrtps_default=use_fastrtps_default,
     use_opensplice_default=use_opensplice_default,
+    use_isolated_default=use_isolated_default,
     cmake_build_type=cmake_build_type,
     ament_build_args_default=ament_build_args_default,
     ament_test_args_default=ament_test_args_default,
--- a/job_templates/ci_launcher_job.xml.em
+++ b/job_templates/ci_launcher_job.xml.em
@@ -102,6 +102,16 @@ for (item in build_numbers) {
               </configs>
             </hudson.plugins.parameterizedtrigger.BooleanParameters>
 @[end if]@
+@[if os_name == 'windows']@
+            <hudson.plugins.parameterizedtrigger.BooleanParameters>
+              <configs>
+                <hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
+                  <name>CI_ISOLATED</name>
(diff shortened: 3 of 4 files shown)


dirk-thomas:
Does it work when triggering the builds using the `ci_launcher` job`?

nuclearsandwich:
ceda24d adds the template snippet to pass the Windows default parameter. I used the connext default change for linux-arm64 above as the example. One thing I changed from that code is that rather than hard code the OS default of `false` I pull the value from the os_data dict. If we ever change the windows default this will still be dead code in effect but at least it won't result in behavior different from the Windows CI jobs. If there's a preference to just hard code false I can change that.


<details><summary>Dry run output (skipped jobs omitted)</summary>

```
Updating job 'ci_windows' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>

> Does it work when triggering the builds using the ci_launcher job`?

Skipped 'ci_packaging_windows' because the config is the same (dry run)
Skipped 'packaging_windows' because the config is the same (dry run)
Updating job 'nightly_win_deb' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Updating job 'nightly_win_rel' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Updating job 'nightly_win_rep' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -98 +98 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
Updating job 'ci_launcher' (dry run)
    <<<
    --- remote config
    +++ new config
    @@ -215,0 +216,8 @@
    +            <hudson.plugins.parameterizedtrigger.BooleanParameters>
    +              <configs>
    +                <hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
    +                  <name>CI_ISOLATED</name>
    +                  <value>false</value>
    +                </hudson.plugins.parameterizedtrigger.BooleanParameterConfig>
    +              </configs>
    +            </hudson.plugins.parameterizedtrigger.BooleanParameters>
    >>>
```
</details>

nuclearsandwich:
Deployed  just before this writing. I triggered a ci_launcher job to verify that windows correctly had CI_ISOLATED = false and that was the case with https://ci.ros2.org/job/ci_windows/4359/ (it and all other triggered jobs were aborted as they served no other purpose than to smoke-test the changes)

PR #153, opened 2018-04-20
place event handlers before

mikaelarguedas:
Pretty subjective as well but this allows me to add cmake args more easily (append at the end of the line rather than placing them in the middle), and I don't expect people to tweak event handlers parameters often


mikaelarguedas:
this targets the colcon branch used in https://github.com/ros2/ci/pull/132

PR #158, opened 2018-04-24
add shell specific packages

### LINKED ISSUE #157: Fat archives not relocatable
The linux and osx packaging jobs are not currently generating re-locatable `prefix.sh` scripts.

Output on OS X:
```
$ source prefix.sh
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/ament_cppcheck/package.sh"
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/ament_flake8/package.sh"
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/ament_pep8/package.sh"
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/console_bridge/package.sh"
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/fastcdr/package.sh"
not found: "/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install/share/gtest_vendor/package.sh"
...
```

Contents:
```
# generated from colcon_core/shell/template/prefix.sh.em

# function to source another script with conditional trace output
# first argument: the name of the script file
colcon_prefix_source_shell_script() {
  # arguments
  _colcon_prefix_source_shell_script="$1"

  # source script with conditional trace output
  if [ -f "$_colcon_prefix_source_shell_script" ]; then
    if [ -n "$COLCON_TRACE" ]; then
      echo ". \"$_colcon_prefix_source_shell_script\""
    fi
    . "$_colcon_prefix_source_shell_script"
  else
    echo "not found: \"$_colcon_prefix_source_shell_script\"" 1>&2
  fi

  unset _colcon_prefix_source_shell_script
}


# since a plain shell script can't determine its own path when being sourced
# it uses the destination set at build time
COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/ament_cppcheck/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/ament_flake8/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/ament_pep8/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/console_bridge/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/fastcdr/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/gtest_vendor/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/orocos_kdl/package.sh"

COLCON_CURRENT_PREFIX=/Users/osrf/jenkins-agent/workspace/packaging_osx/ws/install
colcon_prefix_source_shell_script "$COLCON_CURRENT_PREFIX/share/osrf_pycommon/package.sh"

...

unset COLCON_CURRENT_PREFIX
unset colcon_prefix_source_shell_script
```

Contents for linux is similar but uses `COLCON_CURRENT_PREFIX=/home/rosbuild/ci_scripts/ws/install`

dirk-thomas:
Fixes #157. Connect to ros2/ci#157.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -81,12 +81,18 @@ pip_dependencies = [
     'git+https://github.com/colcon/colcon-package-information.git',
     'git+https://github.com/colcon/colcon-package-selection.git',
     'git+https://github.com/colcon/colcon-parallel-executor.git',
+    'git+https://github.com/colcon/colcon-powershell.git',
     'git+https://github.com/colcon/colcon-python-setup-py.git',
     'git+https://github.com/colcon/colcon-recursive-crawl.git',
     'git+https://github.com/colcon/colcon-test-result.git',
     'git+https://github.com/colcon/colcon-cmake.git',
     'git+https://github.com/colcon/colcon-ros.git',
 ]
+if sys.platform != 'win32':
+    pip_dependencies += [
+        'git+https://github.com/colcon/colcon-bash.git',
+        'git+https://github.com/colcon/colcon-zsh.git',
+    ]
 
 gcov_flags = " -fprofile-arcs -ftest-coverage "
 


dirk-thomas:
* Linux [Build Status](https://ci.ros2.org/job/ci_packaging_linux/84/)
* macOS [Build Status](https://ci.ros2.org/job/ci_packaging_osx/29/)
* Windows [Build Status](https://ci.ros2.org/job/ci_packaging_windows/27/)

dhood:
I might be calling it incorrectly, but it doesn't seem like the powershell script works even on the windows build machines:
```
<powershell terminal>
PS C:\Users\osrf\Downloads\ros2-windows-colcon\ros2-windows> .\prefix.ps1
At C:\Users\osrf\Downloads\ros2-windows-colcon\ros2-windows\prefix.ps1:16 char:66
+     echo "not found: '$_colcon_prefix_source_powershell_script'" 1>&2
+                                                                  ~~~~
The '1>&2' operator is reserved for future use.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : RedirectionNotSupported
```

Are you testing the _result_ of the windows packages somehow?

dhood:
edit: the machine I tested on (windshield) has colcon-powershell==0.0.0 while the machine the archive was built on (icecube) had 0.1.0, which might be related if there are runtime dependency minimums

dirk-thomas:
> Is the only way to get bash support by adding argcomplete? I would have thought that these would be separate notions (and that I could get prefix.bash files generated even if I don't want to/can't install argcomplete)

These two feature (bash and argcomplete) come in two separate packages. `colcon-bash` provides the bash scripts, `colcon-argcomplete` provides completion for `bash` (and `zsh`). The first is usable without the second. I just thought that installing the second makes sense two. If the users have argcomplete installed they will also get completion functionality (if they don't it is just doing nothing for them).

dirk-thomas:
> Are you testing the result of the windows packages somehow?

We are not using the PowerShell scripts yet. We are still relying on Batch. It has certainly worked in the past. Maybe it is a regression of colcon/colcon-powershell#2.

dirk-thomas:
> The '1>&2' operator is reserved for future use.

Addressed by colcon/colcon-powershell#3.

dhood:
OK I have a better understanding now of the impact of the dependencies.

> These two feature (bash and argcomplete) come in two separate packages. colcon-bash provides the bash scripts, colcon-argcomplete provides completion for bash (and zsh). The first is usable without the second. I just thought that installing the second makes sense two. If the users have argcomplete installed they will also get completion functionality (if they don't it is just doing nothing for them).

I didn't follow the reasoning for installing optional dependencies on the buildfarm machines so I assumed it was required. We were trying to keep the dependencies minimal, so if argcomplete isn't required, we should remove it from this PR then, right? Or is it added so that the archives have argcomplete support or something? (in the past that was done based on user-installed dependencies not build-machine deps)

> We are not using the PowerShell scripts yet. We are still relying on Batch. It has certainly worked in the past. Maybe it is a regression of colcon/colcon-powershell#2.

Given that powershell is not supported yet (e.g. we don't have [matching env hooks](https://github.com/ros2/rmw_opensplice/tree/5520a90919995ddf30a3a22e7813bb2462524968/opensplice_cmake_module/env_hook) for powershell) I think it's misleading to provide the powershell scripts in the archives, and think the `colcon-powershell` package should be removed from this PR and from the machines that have had it installed, so that the archives don't generate a `prefix.ps1`.

dirk-thomas:
I removed the `argcomplete` specific packages.

> Given that powershell is not supported yet (e.g. we don't have matching env hooks for powershell) I think it's misleading to provide the powershell scripts in the archives, and think the `colcon-powershell` package should be removed from this PR and from the machines that have had it installed, so that the archives don't generate a `prefix.ps1`.

The fact that some packages don't have an environment hook atm is imo not a reason to remove it. We currently also don't recommend using the `.ps1` file anywhere so it is unlikely that users will just use it and expect it to work.

I think it should stay in order to actually be able to test and use it. And potentially add the missing environment hooks where necessary.

dhood:
> I think it should stay in order to actually be able to test and use it. And potentially add the missing environment hooks where necessary.

I agree that this could be useful one day, but I don't agree with this statement:

> We currently also don't recommend using the .ps1 file anywhere so it is unlikely that users will just use it and expect it to work.

Users will notice the file, just as I did, and try to use it, whether it is recommended or not. If it's shipped in our archives then yes, as a user, I would expect it to work. Since we ship archives with opensplice and I am of the understanding that opensplice support will not be available for powershell, having the system fail on a user in a strange way IMO is not an acceptable trade-off so that we can hypothetically test powershell at some time in the future.

If there isn't full support for powershell with the packages that are included in the archive, then the archive should not include `prefix.sh1` in my opinion.

Easiest way to achieve that is to not install the dependency on the machines. If we think that's unavoidable then perhaps the packaging job should remove the file manually.

I don't believe this applies to zsh for example, because it wraps the sh env hooks from my understanding. If this is not correct, then we should remove the zsh file too is support is not complete.

dirk-thomas:
> we can hypothetically test powershell at some time in the future.

We are talking about an imminent need to use PowerShell because of the limitations of the `cmd` shell. So this is not some distant idea but something we try to do in the coming days.

dhood:
OK, sure, if we plan on adding _full_ powershell support soon then we can leave it as is. If the next release rolls around and it's not complete, I suggest we remove the .ps1 file.

PR #159, opened 2018-04-25
reduce --retest-until-fail to 10

dirk-thomas:
A single test e.g. `test_client_scope_cpp__rmw_connext_cpp` takes ~27s. Times 20 that is already 10 min. for just a single test.

A retry of 20 has become unsustainable with the current accumulated run time of the repeated nightly jobs. They are running for up to 13+h.

Therefore cutting this down to half.


dhood:
I don't think this has been deployed (perhaps on purpose?)

dirk-thomas:
Deployed now for the next round of nightlies.

PR #164, opened 2018-05-05
Stop shipping OpenCV libraries in packaging artifact

mikaelarguedas:
As we provide a separate archive with OpenCV libraries and tell users to install it in the instructions, we don't need to include them in the packaging artifact anymore.

I tested this with the artifacts from https://ci.ros2.org/job/ci_packaging_windows/37/ and could successfully run the demos on windows nodes that have our OpenCV binaries installed.

Context: @sloretz noticed a significant bump in packaging artifact size because we now have both the Release and Debug libraries of OpenCV. By removing them from our archives we save around 50MB archive size

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -175,10 +175,6 @@ def build_and_test_and_package(args, job):
             h.add(args.installspace, arcname=folder_name, exclude=exclude)
     elif args.os == 'windows':
         archive_path = 'ros2-package-windows-%s.zip' % platform.machine()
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
One potential downside of this is that if people use a different version of OpenCV on their machines our the demos we ship in our fat binaries will not work (as it will be looking for DLLs with the exact name that contains the OpenCV version number).
If we're fine with that, I can go ahead and merge this.
@ros2/team thoughts ?

dirk-thomas:
Fine with me.

PR #166, opened 2018-05-08
pass --no-warn-unused-cli to CMake to avoid unwanted warnings

mikaelarguedas:
WIP: opening for visibility, not tested yet

Based on suggestion in meeting. Pass `--no-warn-unused-cli` to CMake to avoid the false positive warnings on CI

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -271,7 +271,7 @@ def build_and_test(args, job):
     ] + (['--merge-install'] if not args.isolated else []) + \
         args.build_args
 
-    cmake_args = ['" -DBUILD_TESTING=1"']
+    cmake_args = ['" -DBUILD_TESTING=1" " --no-warn-unused-cli"']
     if args.cmake_build_type:
         cmake_args.append(
             '" -DCMAKE_BUILD_TYPE=' + args.cmake_build_type + '"')
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -44,7 +44,7 @@ def build_and_test_and_package(args, job):
     ] + (['--merge-install'] if not args.isolated else []) + \
         args.build_args
 
-    cmake_args = []
+    cmake_args = ['" --no-warn-unused-cli"']
     if args.cmake_build_type:
         cmake_args.append(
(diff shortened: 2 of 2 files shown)


nuclearsandwich:
Because cmake arguments are passed to all cmake-based builds regardless of their relevance the value this warning has is in telling us that

1. A job that we expected to use a build variable didn't.
2. No jobs used a build variable.

Since the warning is suppressed at parse-time it would require either grunt work or manual scripting to re-parse the build log and find out if a variable was universally unused (2).

Knowing that a build you expected to use a variable didn't(1) is a nice thing to know but it again requires actively examining the build log and it's still possible to determine that by running a build locally without the warning supression.

As I understand the proposed change it's not possible to override the suppression of this warning in a CI job. Is that correct?

I think given the low value of this warning in our CI, suppressing the warning output via CMake arg now is worth doing. Even if we identify why the error suppression stopped working 100% of the time it's still a bit of a whack-a-mole game.

mikaelarguedas:
> As I understand the proposed change it's not possible to override the suppression of this warning in a CI job. Is that correct?

Correct, an alternative I considered was adding it to the default build args instead, this would allow us to override the warning suppression. I ended up not doing it because I had the feeling that it would clutter the already pretty long build args and thought we may never use the override. But I dont feel strongly about it.

nuclearsandwich:
Nor do I. I think it's fine to make it an innate default. If someone very badly wants to enable the warning they can still use a custom CI scripts branch to do so.

mikaelarguedas:
Rebased and resolved conflicts.

I ran a couple jobs with this and didnt see any false positive CMake warnings. That doesnt mean it fixes the issue but at least allows us to reduce the risk of false positive and to get rid of stderr for most packages.

https://ci.ros2.org/job/ci_linux/4409/
https://ci.ros2.org/job/ci_linux/4410/

Running CI on all platforms:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4412/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1360/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3605/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4468/)

PR #168, opened 2018-05-09
remove trailing forward slash in default jenkins url

mikaelarguedas:
Maybe it's just me but all the entries in my buildfarm.ini file don't have a trailing slash. So I suggest removing it from here

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -50,7 +50,7 @@ def main(argv=None):
         argv = sys.argv[1:]
     parser = argparse.ArgumentParser(description="Creates the ros2 jobs on Jenkins")
     parser.add_argument(
-        '--jenkins-url', '-u', default='https://ci.ros2.org/',
+        '--jenkins-url', '-u', default='https://ci.ros2.org',
         help="Url of the jenkins server to which the job should be added")
     parser.add_argument(
         '--ci-scripts-repository', default='git@github.com:ros2/ci.git',


mikaelarguedas:
Yeah it must be because my entries were based on the ROS 1 buildfarm that doesn't have one: https://github.com/ros-infrastructure/ros_buildfarm_config/blob/f79c0d3616ddec6a34d4740e6e3f95d827b7f433/index.yaml#L78

I'm don;t have a preference as long as it;s consistent across farms. I can update my buildfarm.ini file accordingly if the decision is that the trailing slash should be used everywhere.

dirk-thomas:
When I added the slash to the default value in the referenced PR it was *not* working for me without - even when my local config didn't have a trailing slash. That is why I created the PR changing the default value.

dirk-thomas:
> If you have a preference for "no trailing slash"

I don't have a preference for either or. It just needs to work out-of-the-box with the default value of this argument. The patch seems to work when I remove the trailing slash from my local config again (which wasn't the case when it was added in the first place).

Whatever the end state is it should be consistent across our Jenkins instances. Also any unnecessary change would be good.

dirk-thomas:
It works for me for `build.ros.org`, `build.ros2.org` and `ci.ros2.org` without a trailing slash in the config atm. So I am fine merging this.

mikaelarguedas:
thanks @dirk-thomas for testing this.
And it currently doesn't work for you with the trailing slash on build.ros.org, right?

I'll leave it up to @nuclearsandwich to give the final review / approval in case there's anything we're not aware of (like why it stopped working, or what could break it in the future if we move back to the version without the slash)

dirk-thomas:
> And it currently doesn't work for you with the trailing slash on build.ros.org, right?

Correct.

nuclearsandwich:
Okay so I did some more testing with this on build.ros.org and ci.ros2.org and I think the underlying issue is actually related to jenkins response time on certain HTTP requests.

I need to try again a little later today to see what happens when caches are no longer warm on the jenkins hosts.

Here's a copy of the stacktrace which is what I saw when running without a trailing slash https://gist.github.com/nuclearsandwich/3f186a204d045c118640b3cd0c43ac36 on build.ros.org. Does anyone else have an old stacktrace handy so I can make sure that we're targeting the same issue. I know that previously @dirk-thomas was seeing

```
requests.exceptions.HTTPError: 500 Server Error: Server Error for url: http://build.ros2.org/api/python?tree=jobs%5Bname%2Ccolor%2Curl%5D
```

But I think that was less about the presence or absence of a trailing slash and more about consistent slash presence between the request url and the config url which we attempted to resolve at the time.

nuclearsandwich:
lol apparently I don't recall reviews issued before coffee... :bow:

mikaelarguedas:
:smile: Merging then

PR #170, opened 2018-05-10
allow packaging jobs to run in parallel?

mikaelarguedas:
We discussed this during Ardent release but didn't address it.

Does anyone know of a problem that can occur if we allow packaging jobs to run in parallel?


mikaelarguedas:
> On different machines: no problem.

What about different executors on the same machine (e.g. the packet machine)?

dirk-thomas:
As for any other jobs: running multiple tests on the same machine will result in "cross talk".

mikaelarguedas:
deployed

PR #175, opened 2018-05-23
add parameter to choose colcon version/branch

dirk-thomas:
Add a new parameter to choose the colcon branch:

* Default / empty value: use the latest released version of the packages (https://ci.ros2.org/view/All/job/dirk_ci_linux/7/)
* Pass `master`: uses the master / default branch of all packages (https://ci.ros2.org/view/All/job/dirk_ci_linux/4/)
* Pass a custom branch name: uses the custom branch where available, otherwise the default branch (https://ci.ros2.org/view/All/job/dirk_ci_linux/8/)

Merging / deploying should wait until after the next `colcon` release since the default value (using the latest release) wouldn't work with the current default arguments.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -447,12 +447,12 @@ def run(args, build_function, blacklisted_package_names=None):
         # Print the pip version
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
-        install_pip_packages = list(pip_dependencies)
+        pip_packages = list(pip_dependencies)
         if not args.colcon_branch:
-            install_pip_packages += colcon_packages
+            pip_packages += colcon_packages
         job.run(
             ['"%s"' % job.python, '-m', 'pip', 'install', '-U'] +
-            install_pip_packages, shell=True)
+            pip_packages, shell=True)
 
         # OS X can't invoke a file which has a space in the shebang line
         # therefore invoking vcs explicitly through Python
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -450,6 +450,10 @@ def run(args, build_function, blacklisted_package_names=None):
         pip_packages = list(pip_dependencies)
         if not args.colcon_branch:
             pip_packages += colcon_packages
+        if sys.platform == 'win32':
+            job.run(
+                ['"%s"' % job.python, '-m', 'pip', 'uninstall', '-y'] +
+                pip_packages, shell=True)
         job.run(
             ['"%s"' % job.python, '-m', 'pip', 'install', '-U'] + pip_packages,
             shell=True)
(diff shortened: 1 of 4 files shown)


ros2_batch_job/__main__.py
```diff
@@ -441,7 +447,52 @@ def run(args, build_function, blacklisted_package_names=None):
         # Print the pip version
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
-        job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U'] + pip_dependencies, shell=True)
+        install_pip_packages = list(pip_dependencies)
```
mikaelarguedas: this variable name sounds like a command more than a package list.
We should renaming to something like `pip_packages_to_install`

mikaelarguedas: Why is that temporary variable needed in the first place ? Isn't `pip_dependencies` already a  list that we can extend?
Would something like that do the job?
```
if not args.colcon_branch:
    pip_dependencies.extend(colcon_packages)
```

ros2_batch_job/__main__.py
```diff
@@ -441,7 +447,52 @@ def run(args, build_function, blacklisted_package_names=None):
...
+            vcs_cmd = [
+                '"%s"' % job.python,
+                '"%s"' % os.path.join(venv_path, 'bin', 'vcs')]
+        else:
+            vcs_cmd = ['vcs']
+
+        if args.colcon_branch:
+            # create .repos file for colcon repositories
+            os.makedirs('colcon', exist_ok=True)
+            with open('colcon/colcon.repos', 'w') as h:
```
mikaelarguedas: Could we just use existing repos file rather than constructing it ? https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/master/colcon.repos

Is this because of the packages that are platform specific ?

ros2_batch_job/__main__.py
dirk-thomas: `pip_dependencies` is a global variable. Altering its content within this function doesn't sounds like a good idea to me.

Renamed to `pip_packages` in 5d377b8cefa7cb951a6bd28ab5b309a8c7b6f84e.

ros2_batch_job/__main__.py
dirk-thomas: > Could we just use existing repos file rather than constructing it ? 

No, because the set is not identical.

> Is this because of the packages that are platform specific ?

The platform specific ones are another reason why the file is being generated.

mikaelarguedas: sounds good :+1:

dirk-thomas:
On Windows this needs to [uninstall the packages](85a7fefdec5e5bbebfc31fff606d63cfc8a33fd6) before installing potentially older versions since we don't start from a clean environment: https://ci.ros2.org/view/All/job/dirk_ci_windows/

dirk-thomas:
The PIP releases of `colcon-*` version 0.2.0 are out. Therefore merging and deploying this.

PR #177, opened 2018-05-30
add pattern to match pytest.xml result files

dirk-thomas:
Fixes ros2/build_cop#120.

See https://ci.ros2.org/job/ci_linux/4497/

### DIFF
--- a/job_templates/snippet/publisher_xunit.xml.em
+++ b/job_templates/snippet/publisher_xunit.xml.em
@@ -16,6 +16,13 @@
           <stopProcessingIfError>true</stopProcessingIfError>
         </GoogleTestType>
         <JUnitType>
+          <pattern>@(prefix)/*/pytest.xml</pattern>
+          <skipNoTestFiles>true</skipNoTestFiles>
+          <failIfNotNew>true</failIfNotNew>
+          <deleteOutputFiles>true</deleteOutputFiles>
+          <stopProcessingIfError>true</stopProcessingIfError>
+        </JUnitType>
+        <JUnitType>
           <pattern>@(prefix)/*/test_results/**/*.xunit.xml</pattern>
           <skipNoTestFiles>true</skipNoTestFiles>
           <failIfNotNew>true</failIfNotNew>


dhood:
is this argument to pytest not being used as expected? https://github.com/ament/ament_cmake/blob/f8469fda468493206b3512246204ba4c6885403c/ament_cmake_pytest/cmake/ament_add_pytest_test.cmake#L79
I understood it would customise the output file to match our expected format

dhood:
Ah, perhaps it's only used if called from cmake packages, not pure python packages.

dirk-thomas:
The code you are referring to is for tests which are invoked via CMake. The tests in questions are coming from pure Python packages.

dhood:
For completeness, there's (currently) no way to modify the output path of test results for packages with the ament_python build type (or any general packages using pytest) in colcon as there [was for ament_tools](https://github.com/ament/ament_tools/blob/476dd8934bc14d18c1224a736e2c24313affa02d/ament_tools/build_types/ament_python.py#L90)?

dhood:
... which is why you chose this approach instead? (Note: I didn't mean it as an accusatory tone on the state of colcon, just to confirm why the approach was taken instead of one more "local" to the build)

dirk-thomas:
The code in `ament_tools` you referenced also contains a hard-coded path. So I don't see how that is different from `colcon` (except that the hard-coded path is different).

dirk-thomas:
> ... which is why you chose this approach instead? (Note: I didn't mean it as an accusatory tone on the state of colcon, just to confirm why the approach was taken instead of one more "local" to the build)

I don't understand what you mean with:

> more "local" to the build

In both cases the test results of a package are being save in the build directory of that package.

dhood:
I would think the benefit of having the hard-coded path is that you get the same output file whether you have ament_python or ament_cmake as your build type, so tools processing the files don't need to behave differently for the different build types.

With "local" to the build I mean that the approach we are taking with this PR is to change the parsing in tools that want to use the files, as opposed to standardising the output with the build tool.

We already have different processing for the different test report types, so incrementally it's not a big deal to take this approach, but I was curious why it was chosen.

dhood:
I don't have a strong opinion on this: if it's minimal work to pass the path through, then yes, it would be preferred so that the behaviour is consistent between build types. But from a glance it seems there are some layers of extensions e.g. `PythonTestTask` that would have to be touched to pass the path through, so we can just document the change of behaviour between ament_tools and colcon instead and revisit if it ever causes us/someone else headaches: https://github.com/colcon/colcon.readthedocs.org/pull/16

PR #179, opened 2018-06-01
skip Pytests with a linter marker in repeated jobs

dirk-thomas:
For CTest packages we already skip linter tests in the repeated jobs. This patch proposes to do the same for Pytests.

For that all linter test functions need to be decorated with the corresponding [markers](https://docs.pytest.org/en/latest/example/markers.html). I propose to use the same as for CTest. An example how this would look like:

```
import pytest


@pytest.mark.flake8
@pytest.mark.linter
def test_flake8():
    ...
```

If this PR gets approved I will go ahead and commit the necessary decorators to all Python linter tests. Instead of creating PRs for all of these repositories I will comment with the commits in this ticket.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -219,7 +219,7 @@ def main(argv=None):
             'cmake_build_type': 'None',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
-            'test_args_default': '--event-handler console_direct+ --executor sequential --retest-until-fail 10 --ctest-args " -LE" linter',
+            'test_args_default': '--event-handler console_direct+ --executor sequential --retest-until-fail 10 --ctest-args " -LE" linter --pytest-args " -m" "not linter"',
         })
 
         # configure turtlebot jobs on Linux only for now


dirk-thomas:
In the commits `<test_depend>python3-pytest</test_depend>` needs to be added the package manifest if not already the case.

sloretz:
That's a cool feature

dirk-thomas:
The list of committed changes:
* https://github.com/ament/ament_index/commit/7eee635a336e17eb4b374c636e0ea214721ceb40
* https://github.com/ament/ament_lint/commit/ec2de0236d602fed18adf35509a36a21c2663fb4
* https://github.com/ament/ament_package/commit/dd109feadcc9f30996444703c029d9a693aa2968
* https://github.com/ament/ament_tools/commit/a5ab721ca25f7d0550d2c0e8d8f178bd8ad42f50
* https://github.com/ros2/demos/commit/91ded04b981c339068eaf989136fdb04608288a3
* https://github.com/ros2/examples/commit/45c0976e02c379ebe6518e6983053c2a6045ee9b
* https://github.com/ros2/launch/commit/e93ce3df37014ad13c0ee0c3f3608eba107af976
* https://github.com/ros2/ros2cli/commit/75d480b53b89741a8e1c2993506408bff4e31cc3
* https://github.com/ros2/sros2/commit/2bdf6330becbbc9a6dd3af68bfcd8f7a6f613eeb

mikaelarguedas:
- https://github.com/ros2/teleop_twist_keyboard/commit/e6b141e35197ba8f160027572d77863930559ea3

PR #181, opened 2018-06-04
readd option to use Debian pkg of Connext

dirk-thomas:
Currently the Debian package fails to install in a non-interactive environment. Waiting for feedback / an updated Debian package from RTI...

### DIFF
--- a/ros2_batch_job/linux_batch/__init__.py
+++ b/ros2_batch_job/linux_batch/__init__.py
@@ -69,17 +69,18 @@ class LinuxBatchJob(BatchJob):
                 connext_env_file, 'rti_connext_dds-5.3.1',
                 'resource', 'scripts', 'rtisetenv_x64Linux3gcc5.4.0.bash')
 
-            if not os.path.exists(connext_env_file):
-                warn("Asked to use Connext but the RTI env was not found at '{0}'".format(
-                    connext_env_file))
-                connext_env_file = None
-            elif not self.args.connext_debs:
+            if os.path.exists(connext_env_file):
                 # Make script compatible with dash
                 with open(connext_env_file, 'r') as env_file:
                     env_file_data = env_file.read()
                 env_file_data = env_file_data.replace('${BASH_SOURCE[0]}', connext_env_file)
+                connext_env_file = os.path.join(os.getcwd(), 'rti.sh')
                 with open(connext_env_file, 'w') as env_file:
                     env_file.write(env_file_data)
+            else:
+                warn("Asked to use Connext but the RTI env was not found at '{0}'".format(
+                    connext_env_file))
+                connext_env_file = None
         # There is nothing extra to be done for OpenSplice
 
         ros1_setup_file = None
--- a/ros2_batch_job/linux_batch/__init__.py
+++ b/ros2_batch_job/linux_batch/__init__.py
@@ -69,17 +69,17 @@ class LinuxBatchJob(BatchJob):
                 connext_env_file, 'rti_connext_dds-5.3.1',
                 'resource', 'scripts', 'rtisetenv_x64Linux3gcc5.4.0.bash')
 
-            if os.path.exists(connext_env_file):
+            if not os.path.exists(connext_env_file):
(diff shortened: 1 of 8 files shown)


ros2_batch_job/linux_batch/__init__.py
```diff
@@ -61,9 +61,13 @@ def setup_env(self):
         # but the installed official packages
```
dhood: Is it _necessary_ now to source the bash script to setup connext properly if using debian installation? either way, this wording should probably be updated

dirk-thomas: > Is it necessary now to source the bash script to setup connext properly if using debian installation?

Yes, still necessary. The files are being installed to `/opt/rti`.

> either way, this wording should probably be updated

Please propose a wording you have in mind and feel free to commit it directly to this PR.

dhood: sounds like it needs to be removed completely then

dirk-thomas: Updated in https://github.com/ros2/ci/pull/181/files#diff-eb845129d501bf69837f0bd51ec671eeR60

dhood:
(oh this wasn't in review; well, I'll re-review if things change :) )

dirk-thomas:
> Connext from debs is always installed in the (x86) Docker images, but is only detected if the setup file is sourced (if Connext is used in the build).

I chose this approach since it is the same for OpenSplice.

dhood:
> I chose this approach since it is the same for OpenSplice.

yep, was just making the observation so someone at a glimpse can know what changed.

I can't actually remember, and I couldn't find after a quick search, where the opensplice "activation" is only being done when we want to use opensplice on CI (or where it's blacklisted if we don't want to use it, because I think opensplice from debs is always "active"), but I know it must be in there somewhere and working fine :smile: (this is mostly a note to myself, no action needed)

mikaelarguedas:
> > I chose this approach since it is the same for OpenSplice.

> yep, was just making the observation so someone at a glimpse can know what changed.

As the Connext deb is 160MB vs 3MB for Opensplice, it may be worth exploring a different strategy for Connext, e.g. not installing it by default and install it only when specified (that makes sense only if we expect downloading form the website to stay our default CI configuration).
Maybe it's because I'm far from the apt repo but it takes me between 5 and 20min to download just that deb' `Fetched 161 MB in 5min 33s (483 kB/s)`
And it doesn't seem to be en internet connection problem as previous layers were faster: `Fetched 62.3 MB in 2s (30.5 MB/s)`

> I can't actually remember, and I couldn't find after a quick search, where the opensplice "activation" is only being done when we want to use opensplice on CI (or where it's blacklisted if we don't want to use it, because I think opensplice from debs is always "active"), but I know it must be in there somewhere and working fine smile (this is mostly a note to myself, no action needed)

Blacklisting or rmws based on CI arguments happens [here](https://github.com/ros2/ci/blob/e39be5dffbac7d7cb63c030b0bf9d7c50178c13d/ros2_batch_job/__main__.py#L573-L595)

mikaelarguedas:
@dirk-thomas do you happen to have the log of https://ci.ros2.org/view/All/job/connext_debian_ci_linux/10/ somewhere ? did it run full CI including tests?

I cannot get the security stests to pass on CI: https://ci.ros2.org/job/ci_linux/4555/

Looking at the deb it doesnt seem to provide any security library:
```
rosbuild@5e1026fdb92b:~$ dpkg -L rti-connext-dds-5.3.1 | grep security.so | wc -l
0
```

dirk-thomas:
>  may be worth exploring a different strategy for Connext, e.g. not installing it by default and install it only when specified

The Docker line could certainly be made conditional. :+1:

> do you happen to have the log of https://ci.ros2.org/view/All/job/connext_debian_ci_linux/10/ somewhere ?

No, I don't have any of them.

> I cannot get the security stests to pass on CI

The Debian package of Connext doesn't contain the security plugins. That was a conscious decision of RTI. Users will have to get them from RTI as well as a corresponding license. Once we have update the wiki instructions for the new Debian package they will contribute additional information how to do that.

PR #182, opened 2018-06-05
actually stop building tests on packaging jobs

mikaelarguedas:
As `BUILD_TESTING` is [ON by default](https://cmake.org/cmake/help/v3.5/module/CTest.html), we need to explicitly set it to OFF to avoid building the tests.

e.g. test_communication:
in the [last nightly](https://ci.ros2.org/job/packaging_linux/1097): Finished <<< test_communication [135.10s]

[with this change](https://ci.ros2.org/job/ci_packaging_linux/102/): Finished <<< test_communication [6.12s]

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -44,7 +44,7 @@ def build_and_test_and_package(args, job):
     ] + (['--merge-install'] if not args.isolated else []) + \
         args.build_args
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
> Btw there is no need anymore for the leading space.

Cool! I didnt know this was released. I'll open a follow up to remove the extra escaping for all jobs then :+1:

Issue #187, opened 2018-06-11
Build C++ application failed with last successful package on Linux


minggangw:
Hi there,

Recently I noticed that there was a persistent link error when compiling a simple example written in C++.

- ROS 2 Package: [Last Successful Artifacts](https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2)
- Platform: Ubuntu 16.04
- Project: https://github.com/RobotWebTools/rclnodejs/tree/develop/test/cpp OR https://github.com/ros2/examples
- Error:
``` bash
/usr/bin/ld: warning: libcrypto.so.1.1, needed by /home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libtinyxml2.so.6, needed by /home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0, not found (try using -rpath or -rpath-link)
/home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0: undefined reference to `BUF_MEM_free@OPENSSL_1_1_0'
/home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0: undefined reference to `tinyxml2::XMLNode::Value() const'
/home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0: undefined reference to `EVP_PKEY_new@OPENSSL_1_1_0'
/home/minggang/Downloads/ros2-linux/lib/libfastrtps.so.1.5.0: undefined reference to `X509_STORE_CTX_free@OPENSSL_1_1_0'
```
Detailed log see [here](https://travis-ci.org/RobotWebTools/rclnodejs/builds/390638890?utm_source=github_status&utm_medium=notification)

More information:

```bash
$ldd libfastrtps.so.1.5.0 
./libfastrtps.so.1.5.0: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by ./libfastrtps.so.1.5.0)
	linux-vdso.so.1 =>  (0x00007ffce69ba000)
	libfastcdr.so.1 => /home/minggang/Downloads/ros2-linux/lib/libfastcdr.so.1 (0x00007f9326b83000)
[...]

PR #188, opened 2018-06-11
Keep producing xenial artifacts?

mikaelarguedas:
We mentioned it in the past but don't remember the final decision if we made one.

Should we keep producing xenial fat archives for now?

(slightly related to https://github.com/ros2/ci/issues/187)

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -252,6 +252,16 @@ def main(argv=None):
             'ubuntu_distro': ubuntu_distro,
         })
 
+        ubuntu_distro = 'xenial'
+        # configure a nightly xenial packaging job
+        job_name = 'packaging_{0}_{1}'.format(ubuntu_distro, os_name)
+        create_job(os_name, job_name, 'packaging_job.xml.em', {
+            'cmake_build_type': 'RelWithDebInfo',
+            'test_bridge_default': 'true',
+            'time_trigger_spec': PERIODIC_JOB_SPEC,
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+        })
+
     # configure the launch job
     os_specific_data = collections.OrderedDict()
     for os_name in sorted(os_configs.keys()):


dhood:
this wasn't added in https://github.com/ros2/ci/pull/178 because I was under the impression we wanted one xenial job to rule them all (which would require some refactoring to let CI jobs test the bridge) but this approach is more straightforward :+1:

dhood:
However, because it was out of scope, "bionic or xenial" support was not added for the packaging job in that PR, so there will be more work than just enabling this job (e.g. [it still hard-codes melodic](https://github.com/ros2/ci/blob/b0bf9856f073a1d52773a78601cbc50afc553f0d/job_templates/packaging_job.xml.em#L142)) so I don't think this will build as-is. Shall I work on a follow-up or have you started it already @mikaelarguedas ?

mikaelarguedas:
Oh so the option was added to the packaging job parameters but the parameter is ignored.. That's confusing.

> Shall I work on a follow-up or have you started it already @mikaelarguedas ?

No I have not, and am out for the day, a follow-up will be appreciated :bowing_man:

create_jenkins_job.py
```diff
@@ -252,6 +252,16 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
         })
 
+        ubuntu_distro = 'xenial'
+        # configure a nightly xenial packaging job
+        job_name = 'packaging_{0}_{1}'.format(ubuntu_distro, os_name)
+        create_job(os_name, job_name, 'packaging_job.xml.em', {
+            'cmake_build_type': 'RelWithDebInfo',
+            'test_bridge_default': 'true',
+            'time_trigger_spec': PERIODIC_JOB_SPEC,
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
```
dhood: FYI, this accidentally left out passing the ubuntu distro through, which was fixed up in https://github.com/ros2/ci/commit/dce075c47ccad6e5fc397699e56f6c9e9cca36ce

mikaelarguedas:
Thanks @dhood for the follow-up and fixups!

PR #194, opened 2018-06-15
Replace today_str as part of packaging jobs too

dhood:
This was only in ci_job.xml.em but ought to be in both templates.


mikaelarguedas:
https://github.com/ros2/ci/blob/f47f0ef8a6aae0faad63f7e93f71ca63e62c3599/job_templates/ci_job.xml.em#L189

PR #195, opened 2018-06-15
Move daily invalidation to top of Dockerfile

dhood:
This might have been enough to fix https://github.com/ros2/ros1_bridge/issues/120. It's not known with complete confidence, but either way we probably want to update _all_ dependencies daily.


dirk-thomas:
This would "waste" some time for every build node every day. I would rather not invalidate the complete Docker image every day. Instead if we "know" we have to invalidate it we should "change" a line intentionally.

dhood:
we can revisit if the issue comes up again

PR #199, opened 2018-06-25
remove nose

dirk-thomas:
Follow up of ros2/rosidl#287 and ros2/rosidl_python#3.

* Linux only
  * `nose` removed but no code changes: [Build Status](https://ci.ros2.org/job/ci_linux/4819/)
  * including linked PRs: [Build Status](https://ci.ros2.org/job/ci_linux/4820/)

Once this is merged I will remove `nose` from the various install instructions.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -35,7 +35,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Is the goal to not release ament_cmake_nose ? or just to not install it on CI but to still "build it" and release it for users to use?

dirk-thomas:
The "goal" was to only install dependencies which we need. And the only two packages which were using it did so without declaring a dependency on it which made the devel jobs fail.

I didn't even consider `ament_cmake_nose` since no package is using it. If we want it to be available for  future use we can keep `nose` on CI. The main goal of fixing the devel jobs has been achieved with the referenced PRs already.

mikaelarguedas:
Sounds good,
In that case I'd say let's leave the dependency in at the moment as we have at least a package requiring it and if that package builds successfully I wouldn't expect to have to modify the dependencies installed by CI to be able to use it in a workspace.

We can definitely remove it if we decide to deprecate that package and to advise people to use pytest for all their python testing

PR #201, opened 2018-07-05
create a nightly Windows Debug packaging job

mikaelarguedas:
This will allow people using our nightly artifacts to build Debug overlays

Note: I kept all the jobs parameters identical to the other packaging jobs (except the build type) in case we want to enable this for other platforms than Windows in the future.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -169,7 +169,7 @@ def main(argv=None):
 
         # create a nightly Debug packaging job on Windows
         if os_name == 'windows':
-            create_job(os_name, 'packaging_' + os_name + '_deb', 'packaging_job.xml.em', {
+            create_job(os_name, 'packaging_' + os_name + '_debug', 'packaging_job.xml.em', {
                 'cmake_build_type': 'Debug',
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
after offline discussion we decided to keep the full job name for now as we don't build isolated on Windows.

It was suggested to deduplicate the job parameters and use a single dictionary with packaging job parameters that would be used for all packaging jobs. For parameter audit purposes we decided to keep the full list in each job for now.

Ready for review

mikaelarguedas:
> Please add a link to these debug artifacts to the wiki page.

Will do :+1:

mikaelarguedas:
Actually there is no mention of the nightly artifacts anywhere on the wiki. So I just [added a note](https://github.com/ros2/ros2/wiki/Windows-Install-Binary/_compare/e7b03ccbc714c1c3b2397680fc20a3d6c5e47741...d93631ab6d9a7330e5f8dc8378ec532acd837ffd) in the Windows binary install to let people know that we provide a Debug archive as of Bouncy.

PR #208, opened 2018-07-11
create test_ci_packaging jobs

mikaelarguedas:
This allows to give it a specific label and test packaging on a specific node (similar to the `test_ci_<OS>` jobs)

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -154,6 +154,16 @@ def main(argv=None):
             'use_connext_default': 'false' if os_name is 'linux-aarch64' else 'true',
             'use_connext_debs_default': 'true',
         })
+        # configure test jobs for experimenting with packaging job config changes
+        # Keep parameters the same as the manual triggered job above.
+        create_job(os_name, 'test_ci_packaging_' + os_name, 'packaging_job.xml.em', {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Aren't the `ci_package_*` jobs sufficient for that?

mikaelarguedas:
same rationale as for `test_ci_*` applies here.
Currently the only way to test packaging on a specific machine is to launch jobs until the machine we want to test starts a job.
If not lucky (like it happened a few minutes ago) I had to launch 3 jobs to be able to run 1 job:
https://ci.ros2.org/job/ci_packaging_windows/56/
https://ci.ros2.org/job/ci_packaging_windows/57/
https://ci.ros2.org/job/ci_packaging_windows/58/

dirk-thomas:
I though two different jobs per type would be enough. `ci_*` being used in daily operation and `test_ci_*` for custom configurations. The same for `packageing_*` being triggered automatically while `ci_packaging_*` being used for custom builds.

If you think the last one if being used so frequently that it makes it necessary to have a third one (`test_packaging_*`) that is fine with me. (I just didn't see the need.)

mikaelarguedas:
closing as considered unnecessary. We will just modify the job configuration to target specific hosts.

PR #209, opened 2018-07-16
do not install fast-rtps examples

mikaelarguedas:
Now that https://github.com/eProsima/Fast-RTPS/pull/238 is merged, we can skip installing Fast-RTPS examples on all jobs.

Without this patch: https://ci.ros2.org/job/ci_linux/4915/consoleFull#console-section-17
With this patch: https://ci.ros2.org/job/ci_linux/4916/consoleFull#console-section-17


mikaelarguedas:
deployed

PR #211, opened 2018-07-19
enable rviz's “visual tests” on Linux using a small wm and a framebuffer

wjwwood:
I've been told that a patch like this should allow rviz's visual tests to be run on our Linux CI, even on the headless EC2 instances we use. It works by installing a minimal window manager (matchbox) and a frame buffer device to which to render.

I'll run a CI job to see if actually works, but in the meantime any review of the proposed change is welcome:

- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/4934/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1757/)

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -133,6 +133,17 @@ RUN if test \( ${UBUNTU_DISTRO} != xenial -a ${INSTALL_TURTLEBOT2_DEMO_DEPS} = t
     python3-sphinx \
     ros-${ROS1_DISTRO}-kobuki-driver ros-${ROS1_DISTRO}-kobuki-ftdi; fi
 
+# Install dependencies for RViz visual tests
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    libgl1-mesa-dri \
+    libglapi-mesa \
+    libosmesa6 \
+    mesa-utils \
+    xvfb \
+    matchbox-window-manager
+
+ENV DISPLAY=:99
+
 # Create a user to own the build output.
 RUN useradd -u 1234 -m rosbuild
 RUN sudo -H -u rosbuild -- git config --global user.email "jenkins@ci.ros2.org"
@@ -147,4 +158,4 @@ RUN chmod 755 /entry_point.sh
 
 ENTRYPOINT ["/entry_point.sh"]
 
-CMD ["python3 -u run_ros2_batch.py $CI_ARGS"]
+CMD ["matchbox-window-manager > /dev/null 2>&1 & python3 -u run_ros2_batch.py $CI_ARGS"]
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -58,4 +58,4 @@ echo "done."
 
 cd /home/rosbuild/ci_scripts
 
-exec sudo -H -u rosbuild -E -- /bin/sh -c "$*"
+exec sudo -H -u rosbuild -E -- xvfb-run -s "-ac -screen 0 1280x1024x24" /bin/sh -c "$*"


mikaelarguedas:
Looks like it works! Tests built and ran and passing :tada: (no idea if it's testing the right thing but it looks promising)

Not sure I have enough insight to really review this but given that it works I'm in :+1:

wjwwood:
I'll wait for comments from some rviz contributors before deciding to merge it. @mikaelarguedas in your experience does this significantly change the job time?

mikaelarguedas:
doesnt look like it.

+3 min on amd
+10min on arm
But as we don't know how many other tests have been repeated it's not really meaningful.

A better way to know would be to:
- run jobs with an without that patch that:
  - build only up to rviz*
  - test only rviz* without retries

anhosi:
It is looking good. The "display tests" did run successfully. :fireworks: 
The visual tests (graphical end to end tests using screenshots as asserts) did not run as they need to be started explicitely with `-DEnableVisualTests=True`.

We restarted a single CI run with visual tests enabled.
[Build Status](https://ci.ros2.org/job/ci_linux/4937/)

anhosi:
The CI setup is also working for the visual tests! :+1: 

The test run https://ci.ros2.org/job/ci_linux/4937/ shows 2 errors. One (image_display_visual_test) is listed twice. We already have a fix ready for this one. The other one (odometry_display_visual_test) looks like a parsing problem of the logs as it is passing. In our experience such artifacts go away when there are no real failures.

The runtime of the visual tests is around 8 minutes. It is probably not necessary to have them run for every ros2 build as if other projects break some rviz functionality the other tests should already pick up on that. We would suggest to use the visual tests for rviz pull requests, though.

wjwwood:
CI building all packages, testing only rviz packages, without this patch:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4941/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1762/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4100/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4957/)

CI building all packages, testing only rviz packages, with this patch:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4942/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1763/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4101/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4958/)

I'll compare run times after they complete.

wjwwood:
Looks like linux is affected by about 4-7 minutes with this patch (updated previous comment). However, the other jobs are also slower (macOS and Windows), which is weird. I looked the jobs I ran again, but the only difference is which ci branch is used (test settings look the same). So there could be a good deal of variability within the jobs themselves. 

I'll merge this and since I'm buildfarmer I'll watch for regressions and issues. We'll still manually add the `-DEnableVisualTests=True` when desired.

dirk-thomas:
These new test failures seem to be related to this: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/489/testReport/

wjwwood:
@anhosi (or someone else maybe) is looking into this failure. I'll give them another day to see if they can find the issue and then revert this if not.

wjwwood:
I thought those were because I ran with the "rerun failures" option to check the timing (so I thought they were flakey).

greimela-si:
@wjwwood @dirk-thomas Are these test failures still occurring?

wjwwood:
@greimela-si any ideas about this issue, the test failures are still there. I don't think any of the outstanding rviz pull requests will address it (though I will try to get to those soon).

clalancette:
@greimela-si Any updates for these test failures?  They are the cause of most of the test failures we see when running on aarch64 (last night: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/624)

greimela-si:
@anhosi @Martin-Idel-SI might have details.

PR #216, opened 2018-10-23
add rqt dependencies

dirk-thomas:
@brawner FYI

PR #217, opened 2018-10-25
add qt_gui_core dependency

dirk-thomas:
`libboost-filesystem-dev`

Follow up of #216.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -88,7 +88,7 @@ RUN if test ${UBUNTU_DISTRO} != xenial; then apt-get update && apt-get install -
 RUN apt-get update && apt-get install --no-install-recommends -y libassimp-dev libcurl4-openssl-dev libfreetype6-dev libgles2-mesa-dev libglu1-mesa libqt5core5a libqt5gui5 libqt5opengl5 libqt5widgets5 libxaw7-dev libxrandr-dev qtbase5-dev
 
 # Install build dependencies for rqt et al.
-RUN apt-get update && apt-get install --no-install-recommends -y pyqt5-dev python3-pyqt5 python3-pyqt5.qtsvg python3-sip-dev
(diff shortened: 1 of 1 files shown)


wjwwood:
Why is it necessary? The referenced pr has basically no information in it.

dirk-thomas:
The title mentions that this is a dependency for `qt_gui_core` which happens to use `boost::filesystem`.

wjwwood:
But why not use std::filesytem (like with the `pluginlib/filesystem_helper.hpp` header)? There are no details as to why the decision to not refactor to remove the dependency was made.

dirk-thomas:
@brawner I will let you answer - I only wanted to provide the dependency here...

wjwwood:
I mean, I'm fine with getting the dependency if we can't get rid of it now (though it would be better if we did), but there's no memoranda about this (if I look back on this a year from now, I can't answer "why?").

brawner:
Frankly, because I didn't realize filesystem_helper.hpp existed, so that may fit the task just fine. The code in qt_gui_cpp::RosPluginlibPluginProvider requires finding the parent path and if it's reliable enough in pluginlib::filesystem_helper then I can use that. 

I did not want to rely on <filesystem> or even <experimental/filesystem> because <filesystem> is not installed by default on ubuntu, and <experimental/filesystem> is not in xcode. The current version of VS seems to be unable to compile ros2 currently, so I can only assume that Windows would also have similarly difficulty.

Eventually, it seems like std::filesystem will be included as the bleeding edge compilers are starting to support it. However if you all would prefer to use filesystem_helper rather than boost::filesystem, I can test it out.

dirk-thomas:
The dependency has been removed in https://github.com/ros-visualization/qt_gui_core/commit/b8e44dda7e205522c5fd655afb7d3c5963823de0

Therefore this isn't necessary anymore.

PR #219, opened 2018-11-12
Make curl follow redirects when downloading repos file

sloretz:
`gist.github.com` is redirecting to `gist.githubusercontent.com`. This causes an empty repos file, and the error `Input data is not valid format: 'NoneType' object is not subscriptable` when running a CI job with a custom repos file. ([example job](https://ci.ros2.org/job/ci_osx/4555/console))

Adding `-L` asks curl to follow redirects.


sloretz:
Example CI job with repos file url that redirects with this PR [Build Status](https://ci.ros2.org/job/ci_linux/5524/)

PR #220, opened 2018-11-15
install lark-parser

dirk-thomas:
Install the Python package `lark-parser` to parse IDL files.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -59,7 +59,7 @@ pip_dependencies = [
     'flake8-docstrings',
     'flake8-import-order',
     'flake8-quotes',
-    'git+https://github.com/lark-parser/lark.git@earley_sppf',
+    'git+https://github.com/lark-parser/lark.git@0.7b',
     'mock',
     'nose',
     'pep8',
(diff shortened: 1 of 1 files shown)


jacobperron:
> 1. Should we install the branch via pip install git+https://.../lark.git@earley_sppf and create a Debian package of that state for the Linux releases?
> 2. Or should be create a vendor package instead (likely a fork of the upstream repo/branch with a package.xml file added)?

:thinking: Whichever is less work? In any case I presume we'll want to switch to a future version upstream when possible.

dirk-thomas:
Option 1. is certainly what we want long term. Creating the Debian package is some effort but once there is a real release out with the state we need we can even push that one into Debian if we want to.

jacobperron:
I haven't looked too closely at commits, but would it make more sense to use a more recent branch like `0.7` or `0.7b`? Then we'll hopefully closer match the next release version when it's ready. `earley_sppf` is 68 commits behind.

dirk-thomas:
Both of the branches do work for us (`0.7`, `0.7b`). Not sure which one we want to choose - they have different commit(s) on them.

dirk-thomas:
I updated the patch to use the branch `0.7b` recommended by upstream.

PR #222, opened 2018-11-26
use Debian package of lark-parser

dirk-thomas:
Follow up of #220.


mjcarroll:
`python3-lark-parser` not existing in Xenial is causing nightlies to fail: https://ci.ros2.org/view/nightly/job/nightly_xenial_linux_release/173/

PR #224, opened 2018-11-28
Update libopensplice69 package revision.

nuclearsandwich:
Yesterday we updated the package revision to fix a packaging error.


nuclearsandwich:
While these jobs were aborted they show that the package is installable at this rev. Search the build log for `libopensplice69=` to find the relevant section.

bionic amd64: https://ci.ros2.org/job/ci_packaging_linux/149/
bionic arm64: https://ci.ros2.org/job/ci_packaging_linux-aarch64/40/

Xenial jobs are failing due to the absence of python3-lark-parser (#222)

PR #225, opened 2018-11-28
Exclude python3-lark-parser from Xenial

mjcarroll:
Since we don't provide packages for Xenial, the `python3-lark-parser` package won't exist there.

Instead, it will be installed later via `pip`.


linux_docker_resources/Dockerfile
```diff
@@ -33,6 +33,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-lark-parser python3-setuptools python3-vcstool
```
clalancette: It looks like python3-lark-parser still exists on this line?

mjcarroll: Fixed.

PR #227, opened 2018-12-07
Reduce the length of the nightly_win_extra_rmw_release.

clalancette:
This is because Jenkins uses the full pathname to build, and
with the longer name we run into Windows pathname restrictions.
Note that this is longer than what we truncate other windows
jobs to be, but we would lose too much information by
truncating to 15 characters.

Along with https://github.com/ros2/rosidl_typesupport/pull/41 and https://github.com/ros2/rosidl/pull/339, this should fix the problems pointed out in https://github.com/ros2/build_cop/issues/154


clalancette:
Thanks for the reviews.  I'll merge now, then I'll coordinate the deployment.

PR #228, opened 2018-12-12
skip qt_gui_core and its dependencies on Xenial

dirk-thomas:
The patch skips `qt_gui_cpp` (and its downstream packages) on Xenial since it doesn't support that platform.

[Build Status](https://ci.ros2.org/job/ci_packaging_linux/153/)

Should be merged before the next nightlies.


sloretz:
@dirk-thomas FYI possible cause of ros-visualization/qt_gui_core#168

PR #231, opened 2018-12-19
Pin lark-parser to a specific commit

jacobperron:
This is to hold us over until v0.7 is released.

---

It appears there have been recent breaking changes on [lark@0.7b](https://github.com/lark-parser/lark/commits/0.7b) that have also caused failures in `rosidl_parser` for our CI builds: https://ci.ros2.org/job/ci_linux/5926/

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -59,7 +59,7 @@ pip_dependencies = [
     'flake8-docstrings',
     'flake8-import-order',
     'flake8-quotes',
-    'git+https://github.com/lark-parser/lark.git@870a7611cba40bcbdddad17a085984c7d8806d99',
+    'git+https://github.com/lark-parser/lark.git@8415fa26a3d3d2b81e64e4fe440faab15b53db49',
     'mock',
     'nose',
     'pep8',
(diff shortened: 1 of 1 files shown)


jacobperron:
Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/5928/)

jacobperron:
It didn't occur to me until now, but installing [lark@0.7b](https://github.com/lark-parser/lark/commits/0.7b) is also listed in our [source build instructions](https://index.ros.org/doc/ros2/Linux-Development-Setup/#install-development-tools-and-ros-tools). Perhaps we should update the docs too... Seems like it was a bad idea pointing to a volatile branch.

dirk-thomas:
Has the problem been reported upstream so that there is a chance this gets addressed before the release of 0.7?

jacobperron:
> Has the problem been reported upstream so that there is a chance this gets addressed before the release of 0.7?

No. I'm not sure if it's a problem upstream or `rosidl_parser` breaking because of an API change.
I've narrowed it down to [this commit](https://github.com/lark-parser/lark/commit/04d90fa9165741333c2e394592a64e1c3966aa14) that is causing our tests to fail.

dirk-thomas:
With this PR being merged there is no open ticket about the problem. Please create a new ticket to either fix our usage of the API in the `rosidl` repo or about the API breakage upstream (whichever applies).

Issue #233, opened 2019-01-03
Linux jobs failing to run vcs from venv if a system version is installed.


nuclearsandwich:
The batch script expands the vcs binary to its path in the venv and hard codes that path in the command.

https://github.com/ros2/ci/blob/cd00968c0bf70a8657a84fa9744135a2cef17f8f/ros2_batch_job/__main__.py#L469-L471

Recent Linux CI jobs have started omitting vcs from the venv since it is installed on the system already and provided by /usr/bin/vcs

From [ci_linux #5955](https://ci.ros2.org/job/ci_linux/5955/)
```
==> "/home/rosbuild/ci_scripts/venv/bin/python" -m pip install -U EmPy coverage catkin_pkg flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings flake8-import-order flake8-quotes git+https://github.com/lark-parser/lark.git@8415fa26a3d3d2b81e64e4fe440faab15b53db49 mock nose pep8 pydocstyle pyflakes pyparsing pytest pytest-cov pytest-repeat pytest-rerunfailures pytest-runner pyyaml vcstool colcon-core colcon-defaults colcon-library-path colcon-metadata colcon-output colcon-package-information colcon-package-selection colcon-parallel-executor colcon-powershell colcon-python-setup-py colcon-recursive-crawl colcon-test-result colcon-cmake colcon-ros colcon-bash colcon-zsh
... output elided ...
03:30:01 Collecting vcstool
03:30:01   Downloading https://files.pythonhosted.org/packages/9f/23/0839fbd992b5cc5f7dcb3751e9d8f82eabe5923212e76ddead457dd4a6b0/vcstool-0.1.37-py3-none-any.whl
```

From [ci_linux #5956](https://ci.ros2.org/job/ci_linux/5956/)

```
[...]

PR #234, opened 2019-01-03
Force reinstall dependencies if using venv

sloretz:
If asked to use a virtual environment then force reinstall all packages in the virtual env. This makes sure `vcs` is installed there even if it is also installed system wide.

Fixes ros2/ci#233

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -459,8 +459,13 @@ def run(args, build_function, blacklisted_package_names=None):
             job.run(
                 ['"%s"' % job.python, '-m', 'pip', 'uninstall', '-y'] +
                 colcon_packages, shell=True)
+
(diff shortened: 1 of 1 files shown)


sloretz:
CI
[Build Status](https://ci.ros2.org/job/ci_linux/5963/)

tfoote:
Taking a look at this if we're getting interference from the system version do we really want to use the system-site-package added in #216? @dirk-thomas is it not possible to use the venv only when doing the rqt stuff? It seems like if the compiled elements are there there should be python packages that can take advantage of the compiled elements from the system from inside the venv.

dirk-thomas:
> is it not possible to use the venv only when doing the rqt stuff?

It didn't work when we tried it since some of the Python packages are coming from Debian packages and are not found when not using `--system-site-packages`.

PR #235, opened 2019-01-07
Update job templates for Jenkins upgrades.

nuclearsandwich:
Updates the template files with the changes from the Jenkins upgrade in CI. 

Mostly this is just changing the plugin version in the config but there are some other changes:

* As of Jenkins 2.90 there is now a `trim` variable for `model.hudson.StringParameterDefinition` for which the backwards-compatible value is `false`.
* The Cobertura Plugin report now adds tags for coverage targets in all cases. 

Here's a link to the [dry-run logs](https://gist.github.com/nuclearsandwich/2ace0bde63233b945d152206c1c71e5d).

And a test_ci_linux job is running after reconfiguring via Jenkins UI. The only diff in the dryrun above is from #229. [Build Status](https://ci.ros2.org/job/test_ci_linux/25/)

### DIFF
--- a/job_templates/snippet/publisher_cobertura.xml.em
+++ b/job_templates/snippet/publisher_cobertura.xml.em
@@ -8,6 +8,10 @@
       <zoomCoverageChart>false</zoomCoverageChart>
       <maxNumberOfBuilds>0</maxNumberOfBuilds>
       <failNoReports>false</failNoReports>
+      @# CoverageTargets values consist of three decimal numbers.
+      @#  First: The target value above which a build is considered "100% healthy/sunny".
+      @#  Second: The target value below which a build is considered "0% healthy/stormy".
+      @#  Third: The target value below which a build is considered "Unstable".
       <lineCoverageTargets>80, 0, 0</lineCoverageTargets>
       <methodCoverageTargets>80, 0, 0</methodCoverageTargets>
       <conditionalCoverageTargets>70, 0, 0</conditionalCoverageTargets>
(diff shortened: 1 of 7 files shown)


job_templates/snippet/publisher_cobertura.xml.em
```diff
@@ -8,6 +8,9 @@
       <zoomCoverageChart>false</zoomCoverageChart>
       <maxNumberOfBuilds>0</maxNumberOfBuilds>
       <failNoReports>false</failNoReports>
+      <lineCoverageTargets>80, 0, 0</lineCoverageTargets>
+      <methodCoverageTargets>80, 0, 0</methodCoverageTargets>
+      <conditionalCoverageTargets>70, 0, 0</conditionalCoverageTargets>
```
dirk-thomas: I assume these are the default values suggested by the plugin? Can you describe the semantic meaning of these values: first value and two zeros.

nuclearsandwich: 7bc11f7 adds comments describing the values.

First: coverage score above which is considered 100% healthy
Second: coverage score below which is considered 0% healthy
Third: coverage score below which is considered an unstable build.

The current values are the defaults.

dirk-thomas: Please mention that these are the default value in the comment in the template.

nuclearsandwich: I thought about doing so but it seems like an easy comment to forget to delete if we do decide to changes those values.

nuclearsandwich: Added one in de99381 anyway because I don't feel very strongly about it.

nuclearsandwich:
aaaaaand deployed!

PR #240, opened 2019-01-22
Don't configure xenial nightlies.

nuclearsandwich:
This stops the generation/configuration of the nightly jobs for Xenial.

On-demand CI and packaging jobs can still be created targeting Xenial
which can be used for future Crystal patch releases.

Upon merging this PR, let's also disable those jobs. I'd prefer to wait to delete them for at least a few months in order to have the build logs for the next few Crystal patch releases. But if space gets tight again they're a good candidate for removal.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -252,31 +252,6 @@ def main(argv=None):
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
             })
 
-    for os_name in ['linux', 'linux-aarch64']:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I think we should keep these jobs and change them to build the branches used by Crystal. Since we are supporting this platform for almost one more year it would be good to be able to confirm it passes. It could be changed to be triggered manually though.

nuclearsandwich:
> I think we should keep these jobs and change them to build the branches used by Crystal.

I'm not interested in carrying this forward. So I'll close the PR and anyone can pick up the work later.

PR #242, opened 2019-01-25
temporary hack to keep logs on Windows

dirk-thomas:
To debug ros2/build_cop#161.

Should be reverted later.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -225,7 +225,8 @@ echo "# END SECTION"
 @[  end if]@
 @[elif os_name == 'windows']@
 setlocal enableDelayedExpansion
-rmdir /S /Q ws workspace "work space"
+REM rmdir /S /Q ws workspace "work space"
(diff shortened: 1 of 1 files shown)


job_templates/ci_job.xml.em
```diff
@@ -225,7 +225,8 @@ echo "# END SECTION"
 @[  end if]@
 @[elif os_name == 'windows']@
 setlocal enableDelayedExpansion
-rmdir /S /Q ws workspace "work space"
+REM rmdir /S /Q ws workspace "work space"
+rmdir /S /Q ws/build ws/install ws/source ws/*.repos ws/env.bat
```
sloretz: `ws/source` -> `ws/src`, at least that's what I see in `ci_windows/ws` on icecube

dirk-thomas: :+1: Fixed.

dirk-thomas:
Reconfigured all Windows jobs.

There is actually a second location which deletes the workspace: 23bd5c321ec97c648b01f2b307df24b97ffca2c3 disables that one too.

PR #243, opened 2019-01-25
Revert temporary hack to keep logs on Windows

dirk-thomas:
Reverts #242 once the necessary debug information was extracted.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -225,9 +225,7 @@ echo "# END SECTION"
 @[  end if]@
 @[elif os_name == 'windows']@
 setlocal enableDelayedExpansion
-REM rmdir /S /Q ws workspace "work space"
-rmdir /S /Q ws\build ws\install ws\src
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@cottsay I unassigned myself from the ticket since this item is blocked on the problem to happen on the buildfarm again and the build cop to extract the necessary logs. We could assign the current build cop instead?

sloretz:
@dirk-thomas I collected what I could in ros2/build_cop#161, but CI needed windows nodes so I killed the stuck processes and let it run. If that ticket has enough info to debug the issue then I think this can be merged.

dirk-thomas:
> If that ticket has enough info to debug the issue then I think this can be merged.

It will take some time to review the new logs and determine if the information help to come up with a patch. Therefore I would suggest to wait with merging this.

dirk-thomas:
Since we were able to identify the problem and the necessary fix is described in https://github.com/ros2/build_cop/issues/161#issuecomment-463760815 I will go ahead and merge this revert.

PR #244, opened 2019-02-08
Add nightly job for clang compiler on linux

bhatsach:
Building with different compilers on linux will help detect
bugs early. We will not need to rely on os x builds to
detect clang build failures.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -81,6 +81,7 @@ def main(argv=None):
         'use_isolated_default': 'true',
         'build_args_default': '--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON',
         'test_args_default': '--event-handlers console_direct+ --executor sequential --retest-until-pass 10',
+        'compile_on_clang_default': 'false',
         'enable_c_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
         'turtlebot_demo': False,
@@ -199,6 +200,15 @@ def main(argv=None):
             create_job(os_name, 'nightly_' + os_name + '_coverage', 'ci_job.xml.em', {
                 'cmake_build_type': 'Debug',
                 'enable_c_coverage_default': 'true',
+                'time_trigger_spec': PERIODIC_JOB_SPEC,
+                'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            })
+
+        # configure nightly coverage job for clang compiler on linux
+        if os_name == 'linux':
+            create_job(os_name, 'nightly_' + os_name + '_clang', 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
+                'compile_on_clang_default': 'true',
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
             })
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -81,7 +81,7 @@ def main(argv=None):
         'use_isolated_default': 'true',
         'build_args_default': '--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON',
         'test_args_default': '--event-handlers console_direct+ --executor sequential --retest-until-pass 10',
-        'compile_on_clang_default': 'false',
+        'compile_with_clang_default': 'false',
         'enable_c_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
         'turtlebot_demo': False,
@@ -184,6 +184,15 @@ def main(argv=None):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
+        # configure nightly job for compiling with clang on linux
+        if os_name == 'linux':
+            create_job(os_name, 'nightly_' + os_name + '_clang', 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
+                'compile_with_clang_default': 'true',
+                'time_trigger_spec': PERIODIC_JOB_SPEC,
+                'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            })
+
         # configure a manually triggered version of the coverage job
         if os_name == 'linux':
             create_job(os_name, 'ci_' + os_name + '_coverage', 'ci_job.xml.em', {
@@ -200,15 +209,6 @@ def main(argv=None):
             create_job(os_name, 'nightly_' + os_name + '_coverage', 'ci_job.xml.em', {
                 'cmake_build_type': 'Debug',
                 'enable_c_coverage_default': 'true',
-                'time_trigger_spec': PERIODIC_JOB_SPEC,
-                'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
-            })
-
-        # configure nightly coverage job for clang compiler on linux
-        if os_name == 'linux':
-            create_job(os_name, 'nightly_' + os_name + '_clang', 'ci_job.xml.em', {
-                'cmake_build_type': 'Debug',
-                'compile_on_clang_default': 'true',
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
             })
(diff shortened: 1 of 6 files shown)


bhatsach:
### Testing done
* I set up a local buildfarm by following the steps on https://index.ros.org/doc/ros2/Contributing/CI-Server-Setup/
* I created the new job and it ran successfully. Here is the link the successful run: http://ec2-18-144-12-181.us-west-1.compute.amazonaws.com/job/nightly_linux_clang/ . Please ping me directly to get the credentials to log in to the above buildfarm

### Point to note
* I am currently skipping pluginlib and packages which depend on it as they need an additional library to be linked to work with clang compiler on linux. I will be opening the PRs for these changes and adding the links here. If those two PRs are pushed in time, I can update the build and test to not skip any packages

bhatsach:
* PR link for making pluginlib clang linux compatible: https://github.com/ros/pluginlib/pull/144
* PR link for making rviz clang linux compatible: https://github.com/ros2/rviz/pull/376

ros2_batch_job/__main__.py
```diff
@@ -295,6 +299,13 @@ def build_and_test(args, job):
         cmd.append('--cmake-args')
         cmd.extend(cmake_args)
 
+    if compile_on_clang:
+        install_clang_cmd = ['sudo apt-get install --no-install-recommends -y clang;']
```
clalancette: We've typically done package installation for Linux in the Dockerfiles; see https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile#L35 and the lines below for examples.  Both for consistency and because this file is shared between Linux and the other platforms, I think we should move this package installation over to the Dockerfile, conditional on some variable passed in there.

bhatsach: Agreed, I actually tried to make it work but was facing some issue in matching the variable contents to the expected value. Will fix this and update in next revision.

ros2_batch_job/__main__.py
```diff
@@ -274,6 +277,7 @@ def process_coverage(args, job):
 
 def build_and_test(args, job):
     coverage = args.coverage and args.os == 'linux'
+    compile_on_clang = args.compile_on_clang and args.os == 'linux'
```
tfoote: Why is this specific to linux?

ros2_batch_job/__main__.py
```diff
@@ -295,6 +299,13 @@ def build_and_test(args, job):
         cmd.append('--cmake-args')
         cmd.extend(cmake_args)
 
+    if compile_on_clang:
+        install_clang_cmd = ['sudo apt-get install --no-install-recommends -y clang;']
+        export_clang_variables_cmd = ['export CC=clang CXX=clang++;']
```
tfoote: Instead of exporting this at the top I think it might be cleaner to inject this as a `--cmake-argument` to ament similar to how the gcov flags are inserted below. 

It will also be a lot more visible in the console logs if that argument is explicitly passed to the invocations instead of onetime hidden in the environment at the top.

A few resources; https://stackoverflow.com/a/50494125/7328782 and https://gitlab.kitware.com/cmake/community/wikis/FAQ#how-do-i-use-a-different-compiler

create_jenkins_job.py
```diff
@@ -81,6 +81,7 @@ def main(argv=None):
         'use_isolated_default': 'true',
         'build_args_default': '--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON',
         'test_args_default': '--event-handlers console_direct+ --executor sequential --retest-until-pass 10',
+        'compile_on_clang_default': 'false',
```
dirk-thomas: "compile on" doesn't sound write to me. Code is compiled with or by a compiler. So I would suggest renaming this everywhere to "compile with".

create_jenkins_job.py
```diff
@@ -203,6 +204,15 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
 
+        # configure nightly coverage job for clang compiler on linux
+        if os_name == 'linux':
+            create_job(os_name, 'nightly_' + os_name + '_clang', 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
+                'compile_on_clang_default': 'true',
+                'time_trigger_spec': PERIODIC_JOB_SPEC,
+                'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            })
+
```
dirk-thomas: Please move this new block to line 185/186 to be close to the other nightly jobs and not be separate from them by the coverage jobs.

ros2_batch_job/__main__.py
```diff
@@ -190,6 +190,9 @@ def get_args(sysargv=None):
     parser.add_argument(
         '--src-mounted', default=False, action='store_true',
         help="src directory is already mounted into the workspace")
+    parser.add_argument(
+        '--compile-on-clang', default=False, action='store_true',
+        help="compile on clang instead of gnu")
```
dirk-thomas: Wording: `gnu` is not the compiler - it is `GCC` or `GNU Compiler Collection`.

ros2_batch_job/__main__.py
bhatsach: Moved the installation of clang to dockerfile and based it on a build arg "COMPILE_WITH_CLANG"

ros2_batch_job/__main__.py
```diff
@@ -295,6 +299,13 @@ def build_and_test(args, job):
         cmd.append('--cmake-args')
         cmd.extend(cmake_args)
 
+    if compile_on_clang:
+        install_clang_cmd = ['sudo apt-get install --no-install-recommends -y clang;']
+        export_clang_variables_cmd = ['export CC=clang CXX=clang++;']
+        cmd = install_clang_cmd + export_clang_variables_cmd + cmd
+        # TODO: Stop skipping pluginlib and dep packages once they are clang on linux compatible
```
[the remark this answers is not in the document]
bhatsach: Makes sense, I have removed the skipping of pluginlib based packages

ros2_batch_job/__main__.py
bhatsach: Currently we are only setting the cmake args for compiler paths to clang paths based on this flag. Given that os x clang compiler is already working as expected, I want to avoid adding unnecessary cmake args or any such logic to this workflow.

ros2_batch_job/__main__.py
bhatsach: Makes sense, I have switched to setting cmake compiler path args instead

create_jenkins_job.py
bhatsach: Makes sense, changed the name

create_jenkins_job.py
bhatsach: Makes sense, moved the block

ros2_batch_job/__main__.py
bhatsach: Makes sense, changed the wording

create_jenkins_job.py
```diff
@@ -183,6 +184,15 @@ def create_job(os_name, job_name, template_file, additional_dict):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
+        # configure nightly job for compiling with clang on linux
+        if os_name == 'linux':
+            create_job(os_name, 'nightly_' + os_name + '_clang', 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
```
nuclearsandwich: Is there a reason for the Debug build type being used as the default here? The default build type for CI jobs is usually `None`.

dirk-thomas: All nightly jobs use a specific build type. Only the manual triggered ones don't choose one by default. So using `Debug` seems fine - not sure if also adding `Release` would catch more cases.

bhatsach:
@dirk-thomas, @tfoote, @clalancette and @nuclearsandwich, please let me know if you have any comments on the latest revision, thanks

bhatsach:
> LGTM.
> 
> Thanks for the contribution.

Thanks a lot Dirk and rest of the reviewers :)

dirk-thomas:
The new nightly job: https://ci.ros2.org/view/nightly/job/nightly_linux_clang/

One test build with a few packages to check that it works: [Build Status](https://ci.ros2.org/job/ci_linux/6151/)

bhatsach:
> The new nightly job: https://ci.ros2.org/view/nightly/job/nightly_linux_clang/
> 
> One test build with a few packages to check that it works: [Build Status](https://ci.ros2.org/job/ci_linux/6151/)

We will need to push this change in as well: https://github.com/ros/pluginlib/pull/144 else the pluginlib based packages will fail :(

dirk-thomas:
The new clang job still has numerous additional tests failing as well as compiler warnings. We merged the current state under the assumption that these would be addressed shortly afterwards (based on the feedback the new job provides). So please follow up on these to get the job passing green (or at least have only the same test failures as the GCC based equivalent).

PR #248, opened 2019-02-19
use same basepath in Docker

dirk-thomas:
Currently when clicking on any of the source files with line numbers under the details tab (e.g. https://ci.ros2.org/view/nightly/job/nightly_linux_clang/6/warnings23Result/) Jenkins fails to show the file. This is due to the fact that the build output contains the paths used within the Docker container which after the build can't be found on the host using the same path.

This patch modifies the Docker container to use the same basepath as Jenkins on the host. That ensures that the path can be accessed as is after the build.


dirk-thomas:
I duplicated the clang job and verified in the ran build that the links work then.

PR #249, opened 2019-02-21
Post rosbridge pkgs

Karsten1987:
this PR enables to specify packages which are intended to be built after the ros1 bridge is built.
packages are specified as a space separated list of package names (not repos).

see [Build Status](https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/13/) as a reference build. I believe this could be used as a modification to the existing packaging job.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -182,7 +182,6 @@ def main(argv=None):
                 'num_to_keep': 100,
             },
             'cmake_build_type': 'RelWithDebInfo',
-            'test_bridge_default': 'true',
             'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'} if os_name in ['linux-aarch64', 'linux-armhf'] else set(),
             'use_connext_debs_default': 'true',
         })
@@ -194,7 +193,6 @@ def main(argv=None):
                 'num_to_keep': 370,
             },
             'cmake_build_type': 'RelWithDebInfo',
-            'test_bridge_default': 'true',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
             'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'} if os_name in ['linux-aarch64', 'linux-armhf'] else set(),
@@ -209,7 +207,6 @@ def main(argv=None):
                     'num_to_keep': 370,
                     },
                 'cmake_build_type': 'Debug',
-                'test_bridge_default': 'true',
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
                 'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'} if os_name in ['linux-aarch64', 'linux-armhf'] else set(),
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -134,7 +134,6 @@ def main(argv=None):
     os_config_overrides = {
         'linux-centos': {
             'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp'},
-            'test_bridge_default': 'false',
             'use_connext_debs_default': 'false',
         },
     }
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -40,11 +40,13 @@
     ignore_rmw_default=ignore_rmw_default,
     use_connext_debs_default=use_connext_debs_default,
 ))@
-        <hudson.model.BooleanParameterDefinition>
-          <name>CI_TEST_BRIDGE</name>
-          <description>By setting this to True, the tests for the ros1_bridge will be run.</description>
-          <defaultValue>@(test_bridge_default)</defaultValue>
-        </hudson.model.BooleanParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_MIXED_ROS_OVERLAY_PKGS</name>
+          <description>
+A list of packages - separated by spaces - which explicitly have to be built after sourcing the ROS 1 environment.
+All packages listed here have to be available from either the primary or supplemental repos file.
+          </description>
+        </hudson.model.StringParameterDefinition>
         <hudson.model.StringParameterDefinition>
           <name>CI_TEST_ARGS</name>
           <description>Additional arguments passed to the 'test' verb if testing the bridge.</description>
@@ -117,7 +119,7 @@ use_opensplice: ${build.buildVariableResolver.resolve('CI_USE_OPENSPLICE')}, <br
 colcon_mixin_url: ${build.buildVariableResolver.resolve('CI_COLCON_MIXIN_URL')}, <br/>
 cmake_build_type: ${build.buildVariableResolver.resolve('CI_CMAKE_BUILD_TYPE')}, <br/>
 build_args: ${build.buildVariableResolver.resolve('CI_BUILD_ARGS')}, <br/>
-test_bridge: ${build.buildVariableResolver.resolve('CI_TEST_BRIDGE')}, <br/>
+mixed ros overlay packages: ${build.buildVariableResolver.resolve('CI_MIXED_ROS_OVERLAY_PKGS')}, <br/>
 test_args: ${build.buildVariableResolver.resolve('CI_TEST_ARGS')}\
 """);]]>
         </script>
@@ -162,8 +164,11 @@ if [ -z "${CI_ROS2_REPOS_URL+x}" ]; then
   CI_ROS2_REPOS_URL="@default_repos_url"
 fi
 export CI_ARGS="$CI_ARGS --repo-file-url $CI_ROS2_REPOS_URL"
-if [ "$CI_TEST_BRIDGE" = "true" ]; then
-  export CI_ARGS="$CI_ARGS --test-bridge"
+if [ -n "${CI_ROS2_SUPPLEMENTAL_REPOS_URL+x}" ]; then
+  export CI_ARGS="$CI_ARGS --supplemental-repo-file-url $CI_ROS2_SUPPLEMENTAL_REPOS_URL"
+fi
+if [ -n "${CI_MIXED_ROS_OVERLAY_PKGS+x}" ]; then
+  export CI_ARGS="$CI_ARGS --mixed-ros-overlay-pkgs $CI_MIXED_ROS_OVERLAY_PKGS"
 fi
 if [ "${CI_UBUNTU_DISTRO}" = "bionic" ]; then
   export CI_ROS1_DISTRO=melodic
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -46,9 +46,7 @@
 A list of packages - separated by spaces - which explicitly have to be built after sourcing the ROS 1 environment.
 All packages listed here have to be available from either the primary or supplemental repos file.
           </description>
-          <defaultValue>
-            ros1_bridge
-          </defaultValue>
+          <defaultValue>ros1_bridge</defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.StringParameterDefinition>
           <name>CI_TEST_ARGS</name>
--- a/job_templates/packaging_job.xml.em
+++ b/job_templates/packaging_job.xml.em
@@ -46,6 +46,9 @@
 A list of packages - separated by spaces - which explicitly have to be built after sourcing the ROS 1 environment.
 All packages listed here have to be available from either the primary or supplemental repos file.
           </description>
+          <defaultValue>
+            ros1_bridge
+          </defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.StringParameterDefinition>
           <name>CI_TEST_ARGS</name>
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -191,8 +191,8 @@ def get_args(sysargv=None):
         '--ros1-path', default=None,
         help="path of ROS 1 workspace to be sourced")
     parser.add_argument(
-        '--test-bridge', default=False, action='store_true',
-        help='test ros1_bridge')
+        '--mixed-ros-overlay-pkgs', nargs='+', default=[],
+        help='space separated list of packages to be built in an overlay workspace with ROS 1')
     parser.add_argument(
         '--colcon-mixin-url', default=None,
         help='A mixin index url to be included by colcon')
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -191,7 +191,7 @@ def get_args(sysargv=None):
         '--ros1-path', default=None,
         help="path of ROS 1 workspace to be sourced")
     parser.add_argument(
-        '--mixed-ros-overlay-pkgs', nargs='+', default=[],
+        '--mixed-ros-overlay-pkgs', nargs='*', default=[],
         help='space separated list of packages to be built in an overlay workspace with ROS 1')
     parser.add_argument(
         '--colcon-mixin-url', default=None,
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -24,24 +24,20 @@ from .util import info
 
 
 def build_and_test_and_package(args, job):
-    print('# BEGIN SUBSECTION: build')
-    # ignore ROS 1 bridge package for now
-    ros1_bridge_path = os.path.join(args.sourcespace, 'ros2', 'ros1_bridge')
-    info('ROS1 bridge path: %s' % ros1_bridge_path)
-    ros1_bridge_ignore_marker = None
-    if os.path.exists(ros1_bridge_path):
-        ros1_bridge_ignore_marker = os.path.join(ros1_bridge_path, 'COLCON_IGNORE')
-        info('ROS1 bridge path ignore file: %s' % ros1_bridge_ignore_marker)
-        with open(ros1_bridge_ignore_marker, 'w'):
-            pass
+    skip_pkgs = args.mixed_ros_overlay_pkgs + \
+        (['ros1_bridge'] if 'ros1_bridge' not in args.mixed_ros_overlay_pkgs else [])
+    print('Skipping the following packages from underlay:')
+    for skip_pkg in sorted(skip_pkgs):
+        print('-', skip_pkg)
 
-    # Now run build
+    print('# BEGIN SUBSECTION: build underlay packages')
     cmd = [
         args.colcon_script, 'build',
         '--base-paths', '"%s"' % args.sourcespace,
         '--build-base', '"%s"' % args.buildspace,
         '--install-base', '"%s"' % args.installspace,
     ] + (['--merge-install'] if not args.isolated else []) + \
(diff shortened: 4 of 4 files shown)


dirk-thomas:
Could this be merged with the `--test-bridge` argument and also reuse the build steps already being used for the bridge? Introducing yet another build step after the bridge looks unnecessary to me.

Karsten1987:
that would disable the case where I would only want to build/test my post bridge packages, am I understanding this correctly? One use case is that I only want to build and test the rosbag2 repo and don't really rely on testing the bridge.

dirk-thomas:
> that would disable the case where I would only want to build/test my post bridge packages, am I understanding this correctly?

I don't think so. Instead of having the `ros1_bridge` boolean checkbox there could be a text field of cross-ros-version package names. Where the checkbox is checked atm the list would be populated with the `ros1_bridge` package name. And instead if hard-coded skipping the `ros1_bridge` in the underlay and selecting it in the overlay the new list of package names would be used.

> One use case is that I only want to build and test the rosbag2 repo and don't really rely on testing the bridge.

For that use case you can simply pass the package names you want to use in the overlay and not include the bridge.

Karsten1987:
comments addressed here: https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/21/

dirk-thomas:
> It also looks like the patch doesn't have any part integrating the new argument into the Jenkins UI.

The current patch still doesn't contain any Jenkins UI changes?

Karsten1987:
Can you advise how to do so? I was assuming the jenkins jobs have to be
adjusted by hand. What do I have to do?

dirk-thomas:
> Can you advise how to do so? I was assuming the jenkins jobs have to be adjusted by hand. What do I have to do?

The XML of the Jenkins jobs is generated by the code and snippets in this repo (using the `create_jenkins_job.py` script). For whatever XML you manually changed in the test job you will find the corresponding snippet in this repo. When running the script without options it will output the to-be-changed-diff of the jobs.

job_templates/packaging_job.xml.em
```diff
@@ -152,8 +154,11 @@ if [ -z "${CI_ROS2_REPOS_URL+x}" ]; then
   CI_ROS2_REPOS_URL="@default_repos_url"
 fi
 export CI_ARGS="$CI_ARGS --repo-file-url $CI_ROS2_REPOS_URL"
-if [ "$CI_TEST_BRIDGE" = "true" ]; then
-  export CI_ARGS="$CI_ARGS --test-bridge"
+if [ -n "${CI_ROS2_SUPPLEMENTAL_REPOS_URL+x}" ]; then
+  export CI_ARGS="$CI_ARGS --supplemental-repo-file-url $CI_ROS2_SUPPLEMENTAL_REPOS_URL"
+fi
```
dirk-thomas: Left over from previous iterations?

Karsten1987:
I just changed the `packaging_job.xml.em` to the best of my knowledge :)
If you could tell me how to test this the best, I am happy to push this forward.

ros2_batch_job/packaging.py
```diff
@@ -75,48 +67,46 @@ def build_and_test_and_package(args, job):
...
+        ] + (['--packages-select', overlay_pkgs] if overlay_pkgs else []) +
+            (['--merge-install'] if not args.isolated else []) +
             (
                 ['--cmake-args', '-DCMAKE_BUILD_TYPE=' +
                     args.cmake_build_type]
                 if args.cmake_build_type else []
         ), env=env)
         print('# END SUBSECTION')
 
-        if args.test_bridge:
```
dirk-thomas: Doesn't this need a different condition now? `if args.mixed_ros_overlay_pkgs:`

Karsten1987: I believe this line should be sufficient:
https://github.com/ros2/ci/blob/87db40efa78e47c9bb23da499cc7e5ad4ade52c7/ros2_batch_job/packaging.py#L90

job_templates/packaging_job.xml.em
Karsten1987: I'd need that as long as e.g. `rosbags2` are not the `ros2.repos` file.

dirk-thomas: Currently this variable is not exposed anywhere in the Jenkins UI so you don't have a way to set it and the Python script doesn't have an argument `--supplemental-repo-file-url`. So as is it is unused code.

And you don't need that option. You can simply pass a custom `.repos` file including `rosbag2` through the existing option `CI_ROS2_REPOS_URL`.

ros2_batch_job/packaging.py
dirk-thomas: That implies that if you don't pass any names (which is the equivalent of not checking the previous checkbox) that all packages are being tested (which is not matching the behavior where nothing gets tested). Therefore I do think the condition is needed.

job_templates/packaging_job.xml.em
Karsten1987: I'll expose it through the jenkins ui then. Having a custom repos file only for rosbag doesn't make sense to me as I would have to constantly keep track of the original repos file and update my custom one accordingly.

ros2_batch_job/packaging.py
Karsten1987: I see. I just noticed that the packages are not being tested in the packaging job. I will change that accordingly.

job_templates/packaging_job.xml.em
dirk-thomas: > Having a custom repos file only for rosbag doesn't make sense to me as I would have to constantly keep track of the original repos file and update my custom one accordingly.

Why would a custom repos file be necessary in the first place? With these changes can't you simple add `rosbag2` to the `ros2.repos` file - the same as `ros1_bridge`? I thought that was the goal of this - to always test `rosbag2` as any other package.

ros2_batch_job/packaging.py
```diff
@@ -75,33 +67,33 @@ def build_and_test_and_package(args, job):
...
-            print('# BEGIN SUBSECTION: install Python packages for ros1_bridge test')
-            job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U', 'catkin_pkg', 'rospkg'], shell=True)
-            print('# END SUBSECTION')
+        print('# BEGIN SUBSECTION: install Python packages for ros1_bridge test')
+        job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U', 'catkin_pkg', 'rospkg'], shell=True)
+        print('# END SUBSECTION')
```
dirk-thomas: Should this block be further down in the condition `if args.mixed_ros_overlay_pkgs:`?

job_templates/packaging_job.xml.em
Karsten1987: That would be desired but for this we would have to build the rosbridge for every CI. Which we could once https://github.com/ros2/ros1_bridge/pull/164 landed.

dirk-thomas: It could still be limited to the packaging jobs - the same as it is right now. Just with `rosbag2` in addition to `ros1_bridge`.

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,19 @@
 
 
 def build_and_test_and_package(args, job):
+    overlay_pkgs = ' '.join(args.mixed_ros_overlay_pkgs)
+    skip_pkgs = overlay_pkgs + (' ros1_bridge' if not 'ros1_bridge' in overlay_pkgs else '')
+    print('skipping the following packages at first')
+    print(skip_pkgs)
```
dirk-thomas: Please move the print into the subsection. Also print the skipped packages one at a time and ordered for better readability:

```
for skip_pkg in sorted(skip_pkgs):
    print('-', skip_pkg)
```

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,19 @@
 
 
 def build_and_test_and_package(args, job):
+    overlay_pkgs = ' '.join(args.mixed_ros_overlay_pkgs)
+    skip_pkgs = overlay_pkgs + (' ros1_bridge' if not 'ros1_bridge' in overlay_pkgs else '')
+    print('skipping the following packages at first')
```
dirk-thomas: "at first" isn't very specific. Maybe: `Skipping the following packages from underlay:`.

ros2_batch_job/packaging.py
```diff
@@ -75,33 +67,33 @@ def build_and_test_and_package(args, job):
...
-            print('# BEGIN SUBSECTION: test ROS 1 bridge')
-            # Now run test only for the bridge
+            print('# BEGIN SUBSECTION: test overlay packages')
             ret_test = job.run([
                 args.colcon_script, 'test',
                 '--base-paths', '"%s"' % args.sourcespace,
                 '--build-base', '"%s"' % args.buildspace,
                 '--install-base', '"%s"' % args.installspace,
-                '--packages-select', 'ros1_bridge',
+                '--packages-select', '"%s"' & overlay_pkgs,
```
dirk-thomas: `&` ?

ros2_batch_job/packaging.py
```diff
@@ -56,52 +52,50 @@ def build_and_test_and_package(args, job):
...
+            env = dict(os.environ)
+            env['MAKEFLAGS'] = '-j1'
+            job.run([
+                args.colcon_script, 'build',
+                '--base-paths', '"%s"' % args.sourcespace,
+                '--build-base', '"%s"' % args.buildspace,
+                '--install-base', '"%s"' % args.installspace,
+                '--cmake-args', '-DBUILD_TESTING=ON', '--no-warn-unused-cli',
+                '--event-handlers', 'console_direct+',
+            ] + (['--packages-select', overlay_pkgs]) +
```
dirk-thomas: Each parameter needs to be a separate items in the list. Joining multiple with a space is wrong here. Needs to be:

`] + ['--packages-select'] + args.mixed_ros_overlay_pkgs +`

Same below.

ros2_batch_job/packaging.py
```diff
@@ -56,52 +52,50 @@ def build_and_test_and_package(args, job):
...
                 '--base-paths', '"%s"' % args.sourcespace,
                 '--build-base', '"%s"' % args.buildspace,
                 '--install-base', '"%s"' % args.installspace,
-                '--packages-select', 'ros1_bridge',
+            ] + (['--packages-select', overlay_pkgs]) +
                 # Ignore return codes that indicate test failures;
                 # they'll be picked up later in test reporting.
                 # '--ignore-return-codes',
-            ] + (['--merge-install'] if not args.isolated else []) +
+                (['--merge-install'] if not args.isolated else []) + \
```
dirk-thomas: No need for the trailing `\` within parenthesis.

job_templates/packaging_job.xml.em
```diff
@@ -32,11 +32,13 @@
     ignore_rmw_default=ignore_rmw_default,
     use_connext_debs_default=use_connext_debs_default,
 ))@
-        <hudson.model.BooleanParameterDefinition>
-          <name>CI_TEST_BRIDGE</name>
-          <description>By setting this to True, the tests for the ros1_bridge will be run.</description>
-          <defaultValue>@(test_bridge_default)</defaultValue>
```
dirk-thomas: Since the use of the variable `test_bridge_default` is being removed it probably needs to be removed elsewhere too.

job_templates/packaging_job.xml.em
```diff
@@ -32,11 +32,13 @@
...
-        </hudson.model.BooleanParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>CI_MIXED_ROS_OVERLAY_PKGS</name>
+          <description>
+A list of packages - separated by spaces - which explicitly have to be built after sourcing the ROS 1 environment.
+All packages listed here have to be available from either the primary or supplemental repos file.
+          </description>
+        </hudson.model.StringParameterDefinition>
```
dirk-thomas: This new parameter will probably need a default value so that the packaging jobs continue to test the bridge.

create_jenkins_job.py
```diff
@@ -182,7 +182,6 @@ def create_job(os_name, job_name, template_file, additional_dict):
                 'num_to_keep': 100,
             },
             'cmake_build_type': 'RelWithDebInfo',
-            'test_bridge_default': 'true',
```
Karsten1987: remove fourth location

ros2_batch_job/__main__.py
```diff
@@ -191,8 +191,8 @@ def get_args(sysargv=None):
         '--ros1-path', default=None,
         help="path of ROS 1 workspace to be sourced")
     parser.add_argument(
-        '--test-bridge', default=False, action='store_true',
-        help='test ros1_bridge')
+        '--mixed-ros-overlay-pkgs', nargs='+', default=[],
```
Karsten1987:
```suggestion
        '--mixed-ros-overlay-pkgs', nargs='*', default=[],
```

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,20 @@
...
-    ros1_bridge_ignore_marker = None
-    if os.path.exists(ros1_bridge_path):
-        ros1_bridge_ignore_marker = os.path.join(ros1_bridge_path, 'COLCON_IGNORE')
-        info('ROS1 bridge path ignore file: %s' % ros1_bridge_ignore_marker)
-        with open(ros1_bridge_ignore_marker, 'w'):
-            pass
-
-    # Now run build
+    skip_pkgs = args.mixed_ros_overlay_pkgs + \
+        (['ros1_bridge'] if 'ros1_bridge' not in args.mixed_ros_overlay_pkgs else [])
```
Karsten1987: remove ros1_bridge

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,20 @@
...
+    print('Skipping the following packages from underlay:')
+    for skip_pkg in sorted(skip_pkgs):
+        print('-', skip_pkg)
```
Karsten1987: after subsection

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,20 @@
...
+
+    print('# BEGIN SUBSECTION: build underlay packages')
     cmd = [
         args.colcon_script, 'build',
         '--base-paths', '"%s"' % args.sourcespace,
         '--build-base', '"%s"' % args.buildspace,
         '--install-base', '"%s"' % args.installspace,
     ] + (['--merge-install'] if not args.isolated else []) + \
+        (['--packages-skip', ' '.join(skip_pkgs)] if skip_pkgs else []) + \
```
Karsten1987: ' '.join(args.mixed_ros_overlay_pkgs)

ros2_batch_job/packaging.py
```diff
@@ -56,52 +52,50 @@ def build_and_test_and_package(args, job):
...
-            (
-                ['--cmake-args', '-DCMAKE_BUILD_TYPE=' +
-                    args.cmake_build_type]
-                if args.cmake_build_type else []
-        ), env=env)
-        print('# END SUBSECTION')
+        if args.mixed_ros_overlay_pkgs:
+            print('# BEGIN SUBSECTION: build overlay packages')
+
+            overlay_pkgs = ' '.join(args.mixed_ros_overlay_pkgs)
```
Karsten1987: unused

ros2_batch_job/packaging.py
```diff
@@ -56,52 +52,50 @@ def build_and_test_and_package(args, job):
...
-            '--event-handlers', 'console_direct+',
-        ] + (['--merge-install'] if not args.isolated else []) +
-            (
-                ['--cmake-args', '-DCMAKE_BUILD_TYPE=' +
-                    args.cmake_build_type]
-                if args.cmake_build_type else []
-        ), env=env)
-        print('# END SUBSECTION')
+        if args.mixed_ros_overlay_pkgs:
+            print('# BEGIN SUBSECTION: build overlay packages')
```
Karsten1987: move one up

ros2_batch_job/packaging.py
```diff
@@ -117,6 +111,10 @@ def build_and_test_and_package(args, job):
             )
             info("test-result returned: '{0}'".format(ret_test_results))
             print('# END SUBSECTION')
+        else:
+            print('# BEGIN SUBSECTION: build overlay packages')
```
Karsten1987: remove

ros2_batch_job/packaging.py
```diff
@@ -117,6 +111,10 @@ def build_and_test_and_package(args, job):
...
+            print('no overlay packages specified')
+            print('# END SUBSECTION')
```
Karsten1987: outside of if/else

ros2_batch_job/packaging.py
```diff
@@ -24,24 +24,19 @@
...
+        for skip_pkg in sorted(args.mixed_ros_overlay_pkgs):
+            print('-', skip_pkg)
+
     cmd = [
         args.colcon_script, 'build',
         '--base-paths', '"%s"' % args.sourcespace,
         '--build-base', '"%s"' % args.buildspace,
         '--install-base', '"%s"' % args.installspace,
     ] + (['--merge-install'] if not args.isolated else []) + \
+        (['--packages-skip', ' '.join(args.mixed_ros_overlay_pkgs)]) + \
```
dirk-thomas: This is still wrong. Each argument needs to passed separately - not a single argument with spaces.

Karsten1987: Are you sure? Looking at the current test build it seems to pass:

```
17:35:24 ==> . "/opt/ros/melodic/setup.sh" && . "/home/jenkins-agent/workspace/linux_packaging_post_rosbridge/ws/rti.sh" && /home/jenkins-agent/workspace/linux_packaging_post_rosbridge/venv/bin/colcon build --base-paths "src" --build-base "build" --install-base "install" --merge-install --packages-skip ros1_bridge ros1_rosbag_storage_vendor rosbag2_bag_v2_plugins --event-handlers console_cohesion+ console_package_list+ --cmake-args -DBUILD_TESTING=OFF --no-warn-unused-cli -DCMAKE_BUILD_TYPE=RelWithDebInfo -DINSTALL_EXAMPLES=OFF -DSECURITY=ON
17:35:24 ###############################################################################
17:35:24 
17:35:24          (c) Copyright, Real-Time Innovations, All rights reserved.            
17:35:24                                                                                
17:35:24                            RTI Connext DDS 5.3.1                       
17:35:24 
17:35:24 ###############################################################################
17:35:28 Topological order
17:35:28 - ament_cppcheck (ros.ament_python)
17:35:28 - ament_flake8 (ros.ament_python)
17:35:28 - ament_package (ros.ament_python)
17:35:28 - ament_pep8 (ros.ament_python)
[...]
17:35:28 - rosbag2_tests (ros.ament_cmake)
17:35:28 - rviz_default_plugins (ros.ament_cmake)
17:35:28 - rviz2 (ros.ament_cmake)
```

Neither the `ros1_bridge` nor the `rosbag2_bag_v2_plugins` are built.

https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/39/consoleFull

Karsten1987:
```suggestion
        ['--packages-skip'] + args.mixed_ros_overlay_pkgs + \
```

dirk-thomas:
With the test builds passing this LGTM.

Karsten1987:
thanks for iterating with me over this. Friendly asking you to deploy the new job configuration on the buildfarm.

create_jenkins_job.py
```diff
@@ -134,7 +134,6 @@ def main(argv=None):
     os_config_overrides = {
         'linux-centos': {
             'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp'},
-            'test_bridge_default': 'false',
```
dirk-thomas: The default value for CentOS was `false` and should therefore map to an empty string in `CI_MIXED_ROS_OVERLAY_PKGS`.

PR #250, opened 2019-02-28
Add jenkins parameter for a colcon mixin url.

nuclearsandwich:
The intent is to use this as an improved basis for #245 so colcon mixins can be used for future sanitizer jobs and local reproducibility.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -79,6 +79,7 @@ def main(argv=None):
             'rmw_opensplice_cpp'},
         'use_connext_debs_default': 'false',
         'use_isolated_default': 'true',
+        'colcon_mixin_url': 'https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml',
         'build_args_default': '--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON',
         'test_args_default': '--event-handlers console_direct+ --executor sequential --retest-until-pass 10',
         'compile_with_clang_default': 'false',
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -25,6 +25,7 @@
     use_isolated_default=use_isolated_default,
     ubuntu_distro=ubuntu_distro,
     cmake_build_type=cmake_build_type,
+    colcon_mixin_url=colcon_mixin_url,
     build_args_default=build_args_default,
     test_args_default=test_args_default,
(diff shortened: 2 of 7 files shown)


nuclearsandwich:
In testing on Mac, this configuration failed because adding a `colcon mixin add ...` is not idempotent.

nuclearsandwich:
> In testing on Mac, this configuration failed because adding a `colcon mixin add ...` is not idempotent.

This is resolved by removing the default mixin and squelching the exit status of `colcon mixin remove default` with an `|| true` in b22b977 and that approach appears to be working on all platforms.

bhatsach:
> > In testing on Mac, this configuration failed because adding a `colcon mixin add ...` is not idempotent.
> 
> This is resolved by removing the default mixin and squelching the exit status of `colcon mixin remove default` with an `|| true` in [b22b977](https://github.com/ros2/ci/commit/b22b977a5554c2bb80f09d3294faf3a3f13ab78b) and that approach appears to be working on all platforms.

Thanks for fixing this. I am guessing this is mainly required when you are running these scripts outside the buildfarm. Good thing with the buildfarm is that we run a docker for each job and hence default is not already set.

nuclearsandwich:
> I am guessing this is mainly required when you are running these scripts outside the buildfarm. Good thing with the buildfarm is that we run a docker for each job and hence default is not already set.

This is required because the Mac and Windows platforms don't benefit from isolated environments a la Docker. There's some application level isolation in the form of python virtualenvs and colcon workspaces but colcon's configuration is still global.

sloretz:
> This is resolved by removing the default mixin and squelching the exit status of `colcon mixin remove default` with an `|| true` in [b22b977](https://github.com/ros2/ci/commit/b22b977a5554c2bb80f09d3294faf3a3f13ab78b) and that approach appears to be working on all platforms.

@nuclearsandwich FYI it looks like this [`ci_windows` failure on icecube](https://ci.ros2.org/job/ci_windows/6137/console) might be caused by this PR.

```
14:53:43 ==> C:\Python37\Scripts\colcon.EXE mixin remove default || true
14:53:44 A repository with the name 'default' doesn't exist
14:53:44 'true' is not recognized as an internal or external command,
14:53:44 operable program or batch file.
14:53:44 <== 'C:\Python37\Scripts\colcon.EXE mixin remove default || true' exited with return code '1'
```

A [stackoverflow post](https://stackoverflow.com/questions/22046780/whats-the-windows-command-shell-equivalent-of-bashs-true-command) suggests `(exit 0)` or `VER>NUL` instead of `true`.

PR #251, opened 2019-02-28
use OpenSplice 6.9.190227 on Linux

dirk-thomas:
Together with ros2/rosidl_typesupport_opensplice#20 fixes ros2/rmw_opensplice#259.


dirk-thomas:
On Linux using the new Debian packages of OpenSplice (6.9.190227):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6297/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2715/)

dirk-thomas:
I am merging this without a review since we need to roll this change out since the apt repo doesn't contain the old version anymore.

PR #257, opened 2019-03-07
Install libc++ development libraries when building with clang.

nuclearsandwich:
In support of ros2/ros2#664 
FYI @emersonknapp. 

Further test runs for the libcxx work will want to use this branch (`libcxx`) for the CI_SCRIPTS_BRANCH and set the CI_COMPILE_WITH_CLANG job parameter to true.

For compactness I think it's reasonable to just always install clang and libc++ together but if there's a compelling reason to use a separate build argument and job parameter we can go that route too.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -62,7 +62,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y python3-pip
 RUN pip3 install -U setuptools pip virtualenv
 
 # Install clang if build arg is true
-RUN if test ${COMPILE_WITH_CLANG} = true; then apt-get update && apt-get install --no-install-recommends -y clang; fi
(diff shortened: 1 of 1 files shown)


emersonknapp:
@nuclearsandwich Thanks for this! Now that I know to find this here, I have tried both gcc and clang builds locally using the Docker container - it looks like these 2 packages were the only ones we needed to add 👍

nuclearsandwich:
On its own this won't change the parameters of our existing clang job. This change is a dependency of a forthcoming thread safety annotations job will require both clang and libcxx.

nuclearsandwich:
From discussion on Wednesday, building with clang alone isn't actually sufficient to leverage the Thread Safety Annotations (ros2/ros2#664) and we will be modifying the existing clang nightly to use libcxx in follow-up PRs.

I'd still like to get review for this PR as it will make it easier for myself and the AWS team to iterate on the clang+libcxx related changes to source packages.

PR #260, opened 2019-03-12
Update opensplice package version.

nuclearsandwich:
The latest opensplice package has a version extension indicating it's the second upstream release with the same version number. Because our Docker cache invalidation happens after this package is installed, the cached installation of the older package appears to still be in use. Updating the package here should be sufficient to invalidate the cache.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -68,7 +68,7 @@ RUN if test ${COMPILE_WITH_CLANG} = true; then apt-get update && apt-get install
 RUN apt-get update && apt-get install --no-install-recommends -y gcovr
 
 # Install the OpenSplice binary from the OSRF repositories.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The new r2 Debian package hasn't been imported into our apt repo afaik?

nuclearsandwich:
> The new r2 Debian package hasn't been imported into our apt repo afaik?

It hasn't 

> The new r2 Debian package hasn't been imported into our apt repo afaik?

It hasn't, and as long as it's crystal compatible and functional there's nothing blocking it from being imported.

The CI jobs get the package directly from the gazebo repository. Here's an example failing job https://ci.ros2.org/job/ci_linux/6381/console

dirk-thomas:
Hm, it shouldn't even be in the Gazebo repo... Probably @j-rivero triggered the wrong job when creating the new package for testing.

nuclearsandwich:
This change doesn't affect the job configuration so no deploy was needed. [Build Status](https://ci.ros2.org/job/ci_linux/6389/)

PR #267, opened 2019-03-22
Add nightly job running Address Sanitizer.

nuclearsandwich:
This updates and replaces #245 

Adds a nightly_linux_address_sanitizer job which builds and tests the `test_communication` package (same as the Thread sanitizer) using the asan-gcc mixin and the additional CMake argument `-DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON` since the address sanitizer and memory tools are currently mutually incompatible and there's no plan to change that on the roadmap.

[Build Status](https://ci.ros2.org/job/ci_linux/6509/)

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -185,6 +185,20 @@ def main(argv=None):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
+        # configure nightly job for testing with address sanitizer on linux
+        if os_name == 'linux':
+            asan_build_args = data['default_build_args'].replace('--cmake-args',
+                '--cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON') + \
+                ' --mixin asan-gcc --packages-up-to test_communication'
+
+            create_job(os_name, 'nightly_' + os_name + '_address_sanitizer', 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
+                'time_trigger_spec': PERIODIC_JOB_SPEC,
+                'mailer_recipients': DEFAULT_MAIL_RECIPIENTS + ' ros-contributions@amazon.com',
+                'build_args_default': asan_build_args,
+                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-select test_communication',
+            })
+
         # configure nightly job for compiling with clang+libcxx on linux
         if os_name == 'linux':
             create_job(os_name, 'nightly_' + os_name + '_clang_libcxx', 'ci_job.xml.em', {
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -187,7 +187,7 @@ def main(argv=None):
 
         # configure nightly job for testing with address sanitizer on linux
         if os_name == 'linux':
-            asan_build_args = data['default_build_args'].replace('--cmake-args',
+            asan_build_args = data['build_args_default'].replace('--cmake-args',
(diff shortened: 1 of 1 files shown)


create_jenkins_job.py
```diff
@@ -185,6 +185,20 @@ def create_job(os_name, job_name, template_file, additional_dict):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
+        # configure nightly job for testing with address sanitizer on linux
+        if os_name == 'linux':
+            asan_build_args = data['default_build_args'].replace('--cmake-args',
+                '--cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON') + \
+                ' --mixin asan-gcc --packages-up-to test_communication'
+
+            create_job(os_name, 'nightly_' + os_name + '_address_sanitizer', 'ci_job.xml.em', {
```
thomas-moulard: 'nightly_{}_address_sanitizer'.format(os_name) ?

nuclearsandwich:
This can be merged and deployed as soon as it receives review from a fellow ros2/team collaborator.

create_jenkins_job.py
```diff
@@ -185,6 +185,20 @@ def create_job(os_name, job_name, template_file, additional_dict):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
+        # configure nightly job for testing with address sanitizer on linux
+        if os_name == 'linux':
+            asan_build_args = data['build_args_default'].replace('--cmake-args',
+                '--cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON') + \
+                ' --mixin asan-gcc --packages-up-to test_communication'
```
dirk-thomas: This looks problematic. If the `--cmake-args` in `build_args_default` are followed by actual CMake arguments after the replace they will be behind `--packages-up-to test_communication`.

nuclearsandwich: If I understand you correctly that should be handled.

1. Say `data['build_args_default']` is the following string, which includes --cmake-args
```python
'--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON`'
```
2. `data['build_args_default'].replace('--cmake-args','--cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON')` returns a new string with the inserted cmake arguments prepended to the list that was present previously:
```python
'--event-handlers console_cohesion+ console_package_list+ --cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON -DINSTALL_EXAMPLES=OFF -DSECURITY=ON'
```
3. The new string has the remaining colcon arguments concatenated to the full list of cmake arguments yielding:
```python
'--event-handlers console_cohesion+ console_package_list+ --cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON -DINSTALL_EXAMPLES=OFF -DSECURITY=ON --mixin asan-gcc --packages-up-to test_communication'
```

dirk-thomas: I think I just missed the trailing `)` and didn't see that the next string starting with `--mixin` is independently concatenated. Sorry for the noise.

nuclearsandwich: No worries. It's a valid thing to check

PR #271, opened 2019-04-02
force reinstallation of numpy

dirk-thomas:
To ensure to get the debug version of `numpy` for debug builds.

Follow up of #268.

Aiming to address ros2/build_cop#176.


mjcarroll:
Are there any dependencies that numpy installs that would also need to be purged?

dirk-thomas:
I ran a build using this job to test a debug build. As intended it installed the debug version of `numpy`. A normal build not using this branch will now continue to use the already installed version and fail.

Therefore this needs to be merge right away (even if further iteration is needed to address the debug build problem).

dirk-thomas:
> Are there any dependencies that numpy installs that would also need to be purged?

Not that I know of.

PR #273, opened 2019-04-03
bump OpenSplice version to 6.9.190304

dirk-thomas:
Follow up of #251 and #260 aiming to address ros2/rmw_opensplice#259.

Waiting for the Debian package to become available in the apt repo...


dirk-thomas:
Merging this since all Debian packages are available now and otherwise the builds with OpenSplice will fail (since the oldered Debian package was removed from the apt repo).

PR #274, opened 2019-04-04
use debug Python wheels in packaging job

dirk-thomas:
The packaging job needs to use the debug `numpy` wheel otherwise it fails: [Build Status](https://ci.ros2.org/job/packaging_windows_debug/280/)

After: [Build Status](https://ci.ros2.org/job/ci_packaging_windows/97/) (using the job configuration change manually applied to the job)

:warning: Since this installed the debug wheels on the node which ran the build this should be merged quickly since the job will not reinstall the pip packages for future builds.


dirk-thomas:
While the build is still running it has passed way beyond the point of previous failures... Merging.

PR #278, opened 2019-04-16
Discard old CI builds.

nuclearsandwich:
In order to cut down on storage usage we should rotate out older manual CI builds.

This change only affects ci and ci_packaging jobs for now. I would like
to see if this has a large enough effect on curbing storage growth that we have time to consider archiving the nightly build logs and artifacts more efficiently outside of Jenkins before rotating them out as well.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -67,6 +67,7 @@ def main(argv=None):
     args = parser.parse_args(argv)
 
     data = {
+        'build_discard': dict(),
         'ci_scripts_repository': args.ci_scripts_repository,
         'ci_scripts_default_branch': args.ci_scripts_default_branch,
         'default_repos_url': DEFAULT_REPOS_URL,
@@ -146,6 +147,7 @@ def main(argv=None):
         job_data.update(additional_dict)
         job_data.update(os_config_overrides.get(os_name, {}))
         job_config = expand_template(template_file, job_data)
+        if job_name == 'test_ci_linux': print(job_config)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
     # configure os specific jobs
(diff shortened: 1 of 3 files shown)


create_jenkins_job.py
```diff
@@ -146,6 +147,7 @@ def create_job(os_name, job_name, template_file, additional_dict):
         job_data.update(additional_dict)
         job_data.update(os_config_overrides.get(os_name, {}))
         job_config = expand_template(template_file, job_data)
+        if job_name == 'test_ci_linux': print(job_config)
```
dirk-thomas: Left over debugging?

create_jenkins_job.py
```diff
@@ -159,15 +161,27 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
+                'days_to_keep': 1000,
+                'num_to_keep': 3000,
+            },
         })
 
         # configure a manual version of the packaging job
         create_job(os_name, 'ci_packaging_' + os_name, 'packaging_job.xml.em', {
+            'build_discard': {
+                'days_to_keep': 1000,
+                'num_to_keep': 100,
```
dirk-thomas: This is the only number which looks a bit short. How about keeping at least one year for now?

create_jenkins_job.py
```diff
@@ -67,6 +67,7 @@ def main(argv=None):
     args = parser.parse_args(argv)
 
     data = {
+        'build_discard': dict(),
```
dirk-thomas: Nitpick: use `{}` instead.

nuclearsandwich: Would have expected `{}` to create an empty set rather than an empty dict but TIL.

create_jenkins_job.py
nuclearsandwich: Only the Linux packaging job is currently over 100 and builds and only the windows one is close to 100 jobs. If we want this to have any immediate usefulness it can't be raised.

dirk-thomas: I didn't realize these are the `ci_packaging` jobs. They are only used to verify changes with a packaging build. We could even cut them down more: 180 days?

dirk-thomas:
How about limiting the actual `packaging` jobs?

create_jenkins_job.py
nuclearsandwich: No complaints from me for being stricter :+1:

nuclearsandwich:
> How about limiting the actual `packaging` jobs?

I would like to do that in a follow up PR after assessing how much runway this gives us and the effect it has on day-by-day storage growth.

nuclearsandwich:
Thanks for the style fixups.

PR #279, opened 2019-04-23
Modify ASAN/TSAN nightly jobs configuration

thomas-moulard:
This change modifies the ASAN and TSAN jobs to
only run until rcutils.

I chose rcutils because it is already
green for TSAN and will be green soon for ASAN.

Once both nightly jobs are green, we will have a way
to track ASAN/TSAN regressions.

As the team continues to "clean-up" new packages
higher in the stack, we will continue editing the
nightly jobs to expand progressively the scope of those
jobs to include "newly green" packages.

Signed-off-by: Thomas Moulard <tmoulard@amazon.com>

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -229,7 +229,7 @@ def main(argv=None):
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS + ' ros-contributions@amazon.com',
                 'build_args_default': asan_build_args,
-                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-up-to rcutils',
+                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-select rcpputils rcutils',
             })
 
         # configure nightly job for compiling with clang+libcxx on linux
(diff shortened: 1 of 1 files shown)


thomas-moulard:
@nuclearsandwich PTAL

create_jenkins_job.py
```diff
@@ -249,7 +249,7 @@ def create_job(os_name, job_name, template_file, additional_dict):
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS + ' ros-contributions@amazon.com',
                 'build_args_default': data['build_args_default'] + ' --mixin tsan --packages-up-to test_communication',
-                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-select test_communication',
+                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-up-to rcutils',
```
nuclearsandwich: What motivates the change to testing packages-up-to rather than packages select? I don't think we need to run asan tests on rcutils dependencies, which are mostly the ament packages and launch as a test dependency.

thomas-moulard: I see a few good reasons:

1. Keep ASAN and TSAN jobs aligned (reduce cognitive load)
1. Ensure that introducing a low-level package, it gets instrumented automatically. For instance, in your comment you forgot about `rcpputils` ;)

On a high-level note, my understanding is that we wanted to use `--packages-up-to` but reverted temporarily because doing that up to test communication was generating too much logs and made the JVM crash. Is my understanding wrong?

BTW I don't have a strong opinion for that, we can as well use a list of whitelisted packages if you prefer.

thomas-moulard: I pushed an updated whitelisting test jobs one by one, this seems a better method indeed.

nuclearsandwich:
@thomas-moulard I also modified what we build to only build up to what's being tested, there's a lot of code between the utils packages and test_communication and I'd rather save the CI time than build it all.

thomas-moulard:
@nuclearsandwich SGTM, feel free to merge this whenever you want. Once it's deployed, we'll focus on making this green first and then we'll expand the scope of those jobs.

PR #280, opened 2019-04-23
Use ==/!= to compare str, bytes, and int literals

cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/ros2/ci on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__
```
./ros2_batch_job/__main__.py:244:12: F632 use ==/!= to compare str, bytes, and int literals
        if len(gcda_files) is 0:
           ^
1     F632 use ==/!= to compare str, bytes, and int literals
1
```


dirk-thomas:
Thanks for the patch.

PR #283, opened 2019-04-24
Set default retention to 1000 days, 3000 builds.

nuclearsandwich:
We set this retention policy on ci_* jobs in #278.  This makes that the default for all generated jenkins jobs which will propagate the effect to nightlies and packaging jobs. The 180 days, 100 builds limit on ci_packaging jobs remains the same.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -67,7 +67,9 @@ def main(argv=None):
     args = parser.parse_args(argv)
 
     data = {
-        'build_discard': {},
+        'build_discard': {
+            'days_to_keep': 1000,
+            'num_to_keep': 3000},
         'ci_scripts_repository': args.ci_scripts_repository,
         'ci_scripts_default_branch': args.ci_scripts_default_branch,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
How will keeping 1000 packaging jobs be sufficient to clear up enough room.

The numbers you shared before where 1.1 TB usage with nightlies adding 2.75 GB every day? 1000 days would require 2.75 TB for the nightlies alone which is 2.5 times as much as at the moment.

nuclearsandwich:
> The numbers you shared before where 1.1 TB usage with nightlies adding 2.75 GB every day? 1000 days would require 2.75 TB for the nightlies alone which is 2.5 times as much as at the moment.

You plotted the math out further than I did, I hadn't actually looked at long term storage needs. I was looking at what would clear out enough storage right now. We're definitely adding more than we were >1000 days ago as our archives have gotten bigger.

nuclearsandwich:
b006f74 sets the packaging job retention at 370 days which should give us indefinite legroom. The nightly CI jobs which don't produce archives are still at the 1000 days retention period.

PR #291, opened 2019-05-10
Sets the nightly clang_libcxx job to use rcl_logging_noop

nburek:
When building with libc++ the rcl package will fail to link because it can't find the Log4cxx functions used by rcl_logging_log4cxx. 

This is causing the nightly CI build for linux_clang_libcxx to fail: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/45/
See conversation here: https://github.com/ros2/rcl_logging/pull/9#issuecomment-490628576

This change sets a build argument to use the rcl_logging_noop logger instead of the default one.

I had explored building log4cxx from source with clang and libc++ to resolve this, but ran into issues with getting the Apache Runtime Environment to build. Because of this, we think that it is not worth the extra work to get rcl_logging_log4cxx working just for the nightly build at this point. I've created a new Github issue to track enabling this use case in rcl_logging_log4cxx later (https://github.com/ros2/rcl_logging/issues/14).

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -236,12 +236,16 @@ def main(argv=None):
 
         # configure nightly job for compiling with clang+libcxx on linux
         if os_name == 'linux':
+            # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
+            clang_libcxx_build_args = data['build_args_default'].replace('--cmake-args',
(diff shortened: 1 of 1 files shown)


nburek:
I have been able to test this locally on a linux machine. Without the new cmake arg it fails when compiling with clang and libc++, but succeeds when the new cmake arg is added. 

This should be ready for review.

PR #294, opened 2019-05-15
Pass --retest-until-pass 10 to ASAN and TSAN jobs

thomas-moulard:
ASAN and TSAN jobs are flakey. They have been sometimes
failling and sometimes succeeding over the past one week.

The slow-down factor introduced by the code instrumentation
is probably increasing the likelihood that brittle test logic
will fail.

As a short-term workaround, this commit passes '--retest-until-pass 10'
which is currently used by other CI jobs.

The objective is to determine whether re-trying tests will
stabilize the jobs for now.

Signed-off-by: Thomas Moulard <tmoulard@amazon.com>

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -231,7 +231,9 @@ def main(argv=None):
                 'time_trigger_spec': PERIODIC_JOB_SPEC,
                 'mailer_recipients': DEFAULT_MAIL_RECIPIENTS + ' ros-contributions@amazon.com',
                 'build_args_default': asan_build_args,
-                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-select rcpputils rcutils',
+                'test_args_default': (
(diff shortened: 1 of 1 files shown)


thomas-moulard:
@prajakta-gokhale @xabxx @cev FYI

nuclearsandwich:
The original intent to omitting the retest options was to cut down on retesting guaranteed failures which adds considerable runtime to jobs. But as we've reduced the test scope it might be okay to retest again. 

Let's run some jobs and make sure that 1. It improves the results relative to last night's nightlies and 2. It doesn't drastically affect the runtime.

ASan [Build Status](https://ci.ros2.org/job/ci_linux/7095/)
TSan [Build Status](https://ci.ros2.org/job/ci_linux/7096/)

nuclearsandwich:
I have no argument with those results.

nuclearsandwich:
This is deployed and will run as configured for tonight's nightlies.

PR #293, opened 2019-05-13
Add support for armhf builds.

nuclearsandwich:
Configures the usual suite of nightly, manual CI, and packaging jobs for armhf.

Like the CentOS jobs these are excluded from the CI launcher.

PR #295, opened 2019-05-15
Fix container name arm packaging jobs.

nuclearsandwich:
Managed to get the container names backward with respect to the conditional order. This corrects it.


dirk-thomas:
For the record: follow up of #293.

PR #297, opened 2019-05-23
Blacklist rqt_graph on Windows debug.

clalancette:

This should fix https://github.com/ros2/build_cop/issues/208

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -130,6 +130,10 @@ def main(sysargv=None):
                 'tlsf',
                 'tlsf_cpp',
             ]
+        # There are no debug packages available for PyQt5 and PySide, so we
+        # can't build/run rqt_graph on Windows debug builds.
+        if sys.platform == 'win32' and args.cmake_build_type and args.cmake_build_type == "Debug":
+            blacklisted_package_names.append('rqt_graph')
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -130,6 +130,10 @@ def main(sysargv=None):
                 'tlsf',
                 'tlsf_cpp',
             ]
+        # There are no debug packages available for PyQt5 and PySide, so we
```
dirk-thomas: Why should this be conditional of the packaging flag? Shouldn't it be one level less indented?

ros2_batch_job/__main__.py
```diff
@@ -130,6 +130,10 @@ def main(sysargv=None):
...
+        # can't build/run rqt_graph on Windows debug builds.
+        if sys.platform == 'win32' and args.cmake_build_type and args.cmake_build_type == "Debug":
```
dirk-thomas: The second condition is redundant.

ros2_batch_job/__main__.py
clalancette: Fair point.  I was trying to keep it as restricted as possible, but I guess if we ever turn on "Debug" for packaging, it could hit there as well.  I'll move it out one level.

ros2_batch_job/__main__.py
clalancette: Good point, will remove.

PR #301, opened 2019-06-06
Blacklist rqt_py_common on Windows debug builds.

clalancette:

This should solve the nightly error https://ci.ros2.org/view/nightly/job/nightly_win_deb/1273/testReport/junit/(root)/rqt_py_common/test_test_rqt_ros_graph/

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -131,7 +131,7 @@ def main(sysargv=None):
                 'tlsf_cpp',
             ]
 
-    # There are no Windows debug packages available for PyQt5 and PySide, so
+    # There are no Windows debug packages available for PyQt5 and PySide2, so
     # python_qt_bindings can't be imported to run or test rqt_graph or
     # rqt_py_common.
     if sys.platform == 'win32' and args.cmake_build_type == 'Debug':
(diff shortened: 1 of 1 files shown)


clalancette:
CI: * Windows [Build Status](http://ci.ros2.org/job/ci_windows/7123/)

ros2_batch_job/__main__.py
```diff
@@ -132,9 +132,10 @@ def main(sysargv=None):
             ]
 
     # There are no debug packages available for PyQt5 and PySide, so we
-    # can't build/run rqt_graph on Windows debug builds.
+    # can't build/run rqt_graph or rqt_py_common on Windows debug builds.
```
dirk-thomas: It seems that the package built just fine. It only seems to be failing the tests since there is no debug version of PySide2 available?

clalancette: While the package will build (it is pure python), there is no hope that it will ever run without a debug version of `PyQt5.QtCore` and `PySide2` (they are required for both the tests and runtime, via an import of `python_qt_binding`).

dirk-thomas: Please update the comment then to state the actual reason why it is blacklisted.

clalancette: Done in 71c71ae

clalancette:
The failures on Windows are known, so merging this.

PR #304, opened 2019-06-11
mixed overlay pkgs as template parameter

Karsten1987:
Resolves issue address in here: https://github.com/ros2/ci/pull/249#discussion_r292674891

default parameter for `mixed_overlay_pkgs` is no longer hardcoded in the xml but specified via empy parameter.

/cc @dirk-thomas 

Signed-off-by: Karsten Knese <karsten@openrobotics.org>


dirk-thomas:
:boat: I reconfigured all jobs.

PR #310, opened 2019-06-25
Disable tests for qt_dotgraph on armhf linux builds

emersonknapp:
Part of https://github.com/ros2/ros2/issues/721

Its test dependency python3-pygraphviz is broken on that specific platform. This only affects the test suite and not necessarily the package itself.

pygraphviz 1.5 works fine when installed from pip on my armhf environment. From the rosdep documentation, native packages are preferred, and 1.5 is not available until the Ubuntu 19.04 apt repos. Rosdep doesn't have a way to select against architecture, to choose pip instead of apt for just this Ubuntu Bionic ARMHF case.

See https://ci.ros2.org/job/ci_linux-armhf/16/testReport/junit/ for failing tests on armhf

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -123,6 +123,8 @@ def main(argv=None):
             'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'},
             'build_args_default': data['build_args_default'].replace(
                 '--cmake-args', '--cmake-args -DCMAKE_CXX_FLAGS=-Wno-psabi -DCMAKE_C_FLAGS=-Wno-psabi -DDISABLE_SANITIZERS=ON'),
+            # bionic apt python3-pygraphviz (1.4~rc1) is broken on armhf, disabling only the tests since the package is only used by the tests
+            'test_args_default': '--packages-skip qt_dotgraph ' + data['test_args_default'],
         },
         'linux-centos': {
             'label_expression': 'linux',


emersonknapp:
@clalancette I think this is the right approach, but alternative suggestions welcome.

create_jenkins_job.py
```diff
@@ -123,6 +123,8 @@ def main(argv=None):
             'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'},
             'build_args_default': data['build_args_default'].replace(
                 '--cmake-args', '--cmake-args -DCMAKE_CXX_FLAGS=-Wno-psabi -DCMAKE_C_FLAGS=-Wno-psabi -DDISABLE_SANITIZERS=ON'),
+            # bionic apt python3-pygraphviz (1.4~rc1) is broken on armhf, disabling only the tests since the package is only used by the tests
```
clalancette: Is it really the case that this is only used in tests?  https://github.com/ros-visualization/qt_gui_core/blob/kinetic-devel/qt_dotgraph/src/qt_dotgraph/pygraphvizfactory.py#L33 suggests otherwise.  If that's the case, then we'll probably have to do something different here, since we don't want to package (for instance) something that can't possibly work.  Can you elaborate more on how python3-pygraphviz is broken on Bionic?

emersonknapp: Within this package, pygraphvizfactory is only imported by the tests, and the comment below what you posted says it's a "reference implementation" - but since it's part of the exposed API there is the possibility it's being used.

The issue is - using the following test file, and running `python3 test.py`, the program exits with no output as expected on x86_64 and aarch64

```
# test.py
#!/usr/bin/env python3

import pygraphviz

graph = pygraphviz.AGraph(directed=True, ranksep=0.2, rankdir='TB', rank='same', simplify=True)
graph.add_node('foo')
graph.add_node('bar')
graph.add_edge('foo', 'bar')
```


However, on arm32 (tried two ways - qemu on an x86_64 machine through Docker, and 32 bit Docker on an aarch64 EC2 instance), the following error happens

```
  File "/usr/lib/python3/dist-packages/pygraphviz/agraph.py", line 1561, in __new__
    nh = gv.agnode(graph.handle, n.encode(graph.encoding), _Action.find)
KeyError: 'agnode: no key'

During handling of the above exception, another exception occurred:

  File "test.py", line 8, in <module>
    graph.add_edge('foo', 'bar')
  File "/usr/lib/python3/dist-packages/pygraphviz/agraph.py", line 478, in add_edge
    e = Edge(self, eh=eh)
  File "/usr/lib/python3/dist-packages/pygraphviz/agraph.py", line 1625, in __new__
    s = Node(graph, nh=source)
  File "/usr/lib/python3/dist-packages/pygraphviz/agraph.py", line 1563, in __new__
    raise KeyError("Node %s not in graph." % n)
KeyError: 'Node None not in graph.'
```

If I upgrade to 1.5 via pip after installing the apt dependency `libgraphviz-dev`, this same snippet works on all of these platforms.

Maybe disabling the test is not the right idea, perhaps the dependency needs to be changed to use the latest from pip instead - but this process seems to always be questionable for the Windows build.

clalancette: I see what you mean about it working with pygraphviz 1.5 from pip, but the really weird thing is that it *also* works just fine if I install pygraphviz 1.4rc1 from pip (it does not work when installing from the debian package).  Also, it seems to work with the debian packages python3-pygraphviz 1.5 from disco, and python3-pygraphviz 1.4(!) from cosmic.  As far as I can tell, there are no differences between the cosmic and the bionic packages, other than it was rebuilt for Python 3.7.

Here's a summary:

| OS | pygraphviz version | Installed from debian or pip | Example Works? |
| ---- | ------------------------ | --------------------------------- | ---------- |
| Bionic | 1.4~rc1 | Debian | No |
| Cosmic | 1.4~rc1 | Debian | Yes |
| Disco | 1.5 | Debian | Yes |
| Bionic | 1.4~rc1 | pip | Yes |
| Bionic | 1.5 | pip | Yes |

And one more data point; if I pull down the debian sources (`apt-get source python3-pygraphviz ; apt-get build-dep python3-pygraphviz`), build with no modifications (`debuild -uc -us`), and then install (`dpkg -i python3-pygraphviz_1.4~rc1-1build2_armhf.deb`), it works just fine.

From all of this, I can only conclude that the Ubuntu Bionic debian package is somehow corrupt.  It may be worthwhile following up with Canonical about this, and seeing if they are willing to do a rebuild of the package; that would solve all of the issues here.  There are various things we can do to workaround it, but it would be better for everyone if we got this fixed in upstream Bionic.  @emersonknapp would you mind trying to do that?

emersonknapp: I'm on it - thanks for the extra tests, that's really informative

emersonknapp:
Keeping this up to date - opened bug against package https://bugs.launchpad.net/ubuntu/+source/python-pygraphviz/+bug/1834379 and asking on the #ubuntu-devel IRC

create_jenkins_job.py
nuclearsandwich: Once you've found or opened a bug report with Ubuntu please link it here.

emersonknapp: See main thread of this issue :)

nuclearsandwich: > See main thread of this PR :)
Ah thanks. I had the tab open already and the dynamic update failed :bow:

emersonknapp:
* ubuntu-devel suggested trying a no-change rebuild of the package in a PPA.
* Did that here https://launchpad.net/~emersonknapp/+archive/ubuntu/ppa/+build/17197601
* Installing from my PPA, the package passes the smoke test with no problem
* :+1:
* Ubuntu-devel response 
> "I *think* an SRU with debdiff, where the changelog includes "No change rebuild (LP: 1834379)" would be a reasonable next step
* SRU == https://wiki.ubuntu.com/StableReleaseUpdates

emersonknapp:
The bug has been updated with all requested information, sounds like it will take at least a week to go through. Options I see
1. wait and accept that the armhf build is going to be yellow for that much longer - this is not a core feature of an armhf use case anyways
2. put our own no-change build up on the ros ppa and use that
3. disable the test or remove this "reference implementation"

clalancette:
> The bug has been updated with all requested information, sounds like it will take at least a week to go through.

Thanks for pushing this through Ubuntu; it's really helpful!

If it really is going to be about a week, I say we wait.  If it drags on much longer than that, then we can explore the other options.

kyrofa:
Sorry for the delay folks, stable release updates (SRUs) are always a question of finding who to needle and how much needling they require :wink: . @emersonknapp's fix has finally been accepted into bionic-proposed, where it has been tested and verified. Protocol dictates that it now much sit for 7 days until it can be released into -updates to ensure it doesn't break anything that depends upon it. Given that it's just a rebuild, that seems unlikely. Once 7 days are up I'll do a little more needling and this should be unblocked.

kyrofa:
Alright, the fix is contained in version 1.4~rc1-1build2.1, which has now been released into bionic-updates/universe. This should now be unblocked.

emersonknapp:
Running build to verify qt_dotgraph tests, then will close this PR [Build Status](https://ci.ros2.org/job/ci_linux-armhf/22/)

(Note: arm build agents backed up, might not start for a while)

emersonknapp:
The qt_dotgraph tests are now passing with the new version of the package available in bionic-updates

clalancette:
> The qt_dotgraph tests are now passing with the new version of the package available in bionic-updates

Woohoo!  Thanks for following through with this.

PR #317, opened 2019-07-10
Make opencv-python a pip dependency for Windows.

clalancette:
That way it will get properly installed in both debug
and non-debug versions.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -136,10 +136,14 @@ def main(sysargv=None):
                 'tlsf_cpp',
             ]
 
-    # There are no Windows debug packages available for PyQt5 and PySide2, so
-    # python_qt_bindings can't be imported to run or test rqt_graph or
-    # rqt_py_common.
     if sys.platform == 'win32' and args.cmake_build_type == 'Debug':
+        # There are no Windows debug packages available for opencv-python, so
+        # the image_tools_py test and runtime have no hope of succeeding.
+        blacklisted_package_names.append('image_tools_py')
+
+        # There are no Windows debug packages available for PyQt5 and PySide2, so
+        # python_qt_bindings can't be imported to run or test rqt_graph or
+        # rqt_py_common.
         blacklisted_package_names.append('rqt_graph')
         blacklisted_package_names.append('rqt_py_common')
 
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -136,10 +136,14 @@ def main(sysargv=None):
                 'tlsf_cpp',
             ]
 
-    # There are no Windows debug packages available for PyQt5 and PySide2, so
-    # python_qt_bindings can't be imported to run or test rqt_graph or
-    # rqt_py_common.
     if sys.platform == 'win32' and args.cmake_build_type == 'Debug':
+        # There are no Windows debug packages available for opencv-python, so
```
mjcarroll: Can't win here...

ros2_batch_job/__main__.py
```diff
@@ -79,6 +79,10 @@
         'cryptography',
         'lxml',
     ]
+if sys.platform == 'win32' and not args.cmake_build_type == 'Debug':
+    pip_dependencies += [
+        'opencv-python',
```
nuclearsandwich: Will opencv-python from pypi work with our opencv binary?

mjcarroll: A follow up question would be why doesn't our opencv binary include opencv-python?

mjcarroll: That could potentially fix the debug builds as well.

clalancette: > Will opencv-python from pypi work with our opencv binary?

Yes and no.  opencv-python is its own compiled version of the library, and doesn't use our version.  Because of this, there is a potential of version drift between them (we currently package OpenCV 3.4, while opencv-python from pip is currently 4.10).

> A follow up question would be why doesn't our opencv binary include opencv-python?

Yeah, maybe that is the right way to go.  I guess we'd have to figure out how to make that compile happen, and then we'd also have to figure out how to tell users to unpack properly to `C:\python37` and/or setup `PYTHONPATH`.  @mjcarroll , up for a challenge? :).

> That could potentially fix the debug builds as well.

Only if you can include both the debug and release `.pyd` files in the same tarball.  Can you?

nuclearsandwich: > Only if you can include both the debug and release `.pyd` files in the same tarball. Can you?

I believe there's some dark magic in the CI scripts to uninstall and reinstall python extensions on Windows so the correct Debug or Release library is used.

clalancette:
Giving up on this one, we'll decide whether to do this another day.

PR #318, opened 2019-07-16
Revert "Pin pytest version until a fix is released in colcon."

nuclearsandwich:
Reverts ros2/ci#311

When the changes in https://github.com/colcon/colcon-core/pull/200 are released this will no longer be needed.


dirk-thomas:
Passing CI build with only `ament_cppcheck`: [Build Status](https://ci.ros2.org/job/ci_linux/7591/)

PR #320, opened 2019-07-23
enable ccache stats on macOS

dirk-thomas:
To demonstrate the benefits for ros2/build_cop#219.

macOS build with `ccache` stats: [Build Status](https://ci.ros2.org/job/ci_osx/6246/)


dirk-thomas:
The intent is to merge this patch and have the statistics available the same as for Linux nodes.

PR #322, opened 2019-07-31
update OpenSplice Debian version to 6.9.190705+osrf1-1~$UBUNTU_DISTRO

dirk-thomas:
Needed after http://build.ros2.org/job/import_upstream/161/ since current builds are failing: https://ci.ros2.org/job/ci_packaging_linux/222/


dirk-thomas:
The new build has passed the point of installing the OpenSplice deb successfully: https://ci.ros2.org/job/ci_packaging_linux/223/

PR #323, opened 2019-07-31
kill existing colcon when stating CI on Windows to avoid file locks

wjwwood:
Workaround for https://github.com/ros2/build_cop/issues/161 on Windows. It would be nice to understand why this is happening and if possible make colcon more robust to this kind of situation, assuming there is anything we could do differently in colcon or in our batch ci script.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -146,6 +146,8 @@ def main(sysargv=None):
             'qt_dotgraph',
         ]
 
+    # TODO(wjwwood): remove this when a better solution is found, as
+    #   this is just a work around for https://github.com/ros2/build_cop/issues/161
     # If on Windows, kill any still running `colcon` processes to avoid
     # problems when trying to delete files from pip or the workspace during
     # this job.
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -18,6 +18,7 @@ import platform
 from shutil import which
 import subprocess
 import sys
+import time
 
 # Make sure we're using Python3
 assert sys.version.startswith('3'), "This script is only meant to work with Python3"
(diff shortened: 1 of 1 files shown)


clalancette:
For what it is worth, this may improve the situation, but I'm pretty sure it won't solve it completely.  When I've killed off colcon by hand sometimes, there still seem to be rogue Python.exe processes that are still running, and that also cause subsequent builds to fail.

Actually, thinking about it more, that may be a clue.  Those Python processes may be leftover tests that don't properly exit, and colcon may be waiting around for them to complete.  Not sure how to verify that.

wjwwood:
In my limited experiments when I killed colcon it caused the whole tree to exit (from subprocesses up to the "batch" ci script). But for sure it may not be sufficient. Hopefully this will be incrementally better and we can continue to make it more intelligent.

wjwwood:
Here's a job on `icecube` failing due to a zombie colcon process:

- [Build Status](https://ci.ros2.org/job/test_ci_windows/56/)

And it getting past that point with this branch:

- [Build Status](https://ci.ros2.org/job/test_ci_windows/57/)

Obviously we could do better (kill more/different processes) or adapt the "subprocess reaper" script from build.ros.org, but I'd propose we merge this (with a positive review) until then.

ros2_batch_job/__main__.py
```diff
@@ -144,6 +145,14 @@ def main(sysargv=None):
             'image_tools_py',
             'qt_dotgraph',
         ]
+
+    # If on Windows, kill any still running `colcon` processes to avoid
+    # problems when trying to delete files from pip or the workspace during
+    # this job.
```
jacobperron: Is it worth mentioning that this is a workaround for https://github.com/ros2/build_cop/issues/161 and maybe add a TODO to remove once the real issue is resolved?

wjwwood: Yeah, I was gonna mention it on the that issue (rather than close it) and then make a new issue to track the removal, but I can also add a TODO.

wjwwood: 33ff469d4d3c6fdb43a5115708c477ba0586aecc

wjwwood:
@cottsay can I leave this to you to decide when/if to merge it?

cottsay:
> @cottsay can I leave this to you to decide when/if to merge it?

Yes, I'll take a look and merge it when I have a sec.

PR #325, opened 2019-08-05
install Python package mypy

dirk-thomas:
Required for ament/ament_lint#154.


dirk-thomas:
> Do you think we should add to any of the Eloquent installation instructions?

Absolutely: ros2/ros2_documentation#311

PR #326, opened 2019-08-14
Run update-ccache-symlinks after build-essential is installed.

nuclearsandwich:
According to man(8) update-ccache-symlinks we shouldn't have to worry
about running this by hand because it should run whenever a compiler is
installed. However I noticed that the symlinks for plain old cc were
missing in our ci_batch_job image.

When trying to sort this out locally it appears that not all compiler packages correctly trigger the update. Whether you install build-essential together with ccache or install ccache first the result is symlinks for the following in `/usr/lib/ccache/`:

```
c89-gcc  c99-gcc  g++  g++-7  gcc  gcc-7  x86_64-linux-gnu-g++  x86_64-linux-gnu-g++-7  x86_64-linux-gnu-gcc  x86_64-linux-gnu-gcc-7
```

Installing ccache after produces a complete set. But I think it makes more sense to just forcibly invoke the updater here so we call apt fewer times.

I've been debugging these on a test jenkins instance so the jobs aren't linkable but here are before and after excerpts of osrf_testing_tools_cpp's cmake configuration:

Before:

```
--- output: osrf_testing_tools_cpp
Not searching for unused variables given on the command line.
-- The C compiler identification is GNU 7.4.0
-- The CXX compiler identification is GNU 7.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
```

After:

```
--- output: osrf_testing_tools_cpp
Not searching for unused variables given on the command line.
-- The C compiler identification is GNU 7.4.0
-- The CXX compiler identification is GNU 7.4.0
-- Check for working C compiler: /usr/lib/ccache/cc
-- Check for working C compiler: /usr/lib/ccache/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib/ccache/c++
-- Check for working CXX compiler: /usr/lib/ccache/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
```

And the before and after ccache stats for the latter job:

```
cache directory                     /home/rosbuild/.ccache
primary config                      /home/rosbuild/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Wed Aug 14 18:23:40 2019
cache hit (direct)                     0
cache hit (preprocessed)               0
cache miss                             0
cache hit rate                      0.00 %
preprocessor error                     3
cleanups performed                     0
files in cache                         0
cache size                           0.0 kB
max cache size                       5.0 GB
```

```
cache directory                     /home/rosbuild/.ccache
primary config                      /home/rosbuild/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Wed Aug 14 18:23:40 2019
cache hit (direct)                     7
cache hit (preprocessed)               0
cache miss                           202
cache hit rate                      3.35 %
called for link                      116
compile failed                         1
preprocessor error                     5
cleanups performed                     0
files in cache                       400
cache size                          21.7 MB
max cache size                       5.0 GB
```

although these were many misses this is a far larger amount of cache attempts overall than we've been seeing on the life farm and is therefore good news.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -35,6 +35,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 RUN if test ${UBUNTU_DISTRO} != xenial; then apt-get update && apt-get install --no-install-recommends -y python3-lark-parser python3-opencv; fi
+RUN update-ccache-symlinks
 
 # Install build and test dependencies of ROS 2 packages.
 RUN apt-get update && apt-get install --no-install-recommends -y \
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -35,7 +35,6 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 RUN if test ${UBUNTU_DISTRO} != xenial; then apt-get update && apt-get install --no-install-recommends -y python3-lark-parser python3-opencv; fi
-RUN update-ccache-symlinks
 
 # Install build and test dependencies of ROS 2 packages.
 RUN apt-get update && apt-get install --no-install-recommends -y \
@@ -172,6 +171,12 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     mesa-utils \
     xvfb \
     matchbox-window-manager
+
+# After all packages are installed, update ccache symlinks (see ros2/ci#326).
+# This command is supposed to be invoked whenever a new compiler is installed
+# but that isn't happening. So we invoke it here to make sure all compilers are
+# picked up.
+RUN update-ccache-symlinks
 
 ENV DISPLAY=:99
 


nuclearsandwich:
> Any reason not to run it after all of the debs have been installed? It might not be necessary right now, but the nature of the operation is to account for compilers which weren't present when it was last run...

Nothing firm, I defaulted to trying to keep the workaround localized near the line which "causes" the problem. I'm fine moving it, probably with a comment.

nuclearsandwich:
Interestingly I can reproduce this misbehavior in docker containers all the way up to debian:bullseye and ubuntu:eoan and back to debian:stretch but it's not a problem in debian:jessie or ubuntu:xenial. With that information it's likely that our Bionic builds on CI have never actually used ccache fully.

nuclearsandwich:
I think this means that a followup to this PR will be to ID which package isn't triggering the update and find or open a Debian bug for that behavior.

nuclearsandwich:
> Any reason not to run it after all of the debs have been installed? It might not be necessary right now, but the nature of the operation is to account for compilers which weren't present when it was last run...


d37bfa7 moves the update below all apt-get installs.

nuclearsandwich:
CI that builds everything (juust in case using CCache happens to bork anything) but runs basically no tests.

:penguin: [Build Status](https://ci.ros2.org/job/ci_linux/7887/)
:penguin: :muscle: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3940/)
:penguin: :muscle: :business_suit_levitating: [Build Status](https://ci.ros2.org/job/ci_linux-armhf/57/)

dirk-thomas:
How about installing `ccache` in a separate `RUN` line after the compilers and add a note why it is separate?

nuclearsandwich:
> How about installing `ccache` in a separate `RUN` line after the compilers and add a note why it is separate?

I like this approach less because it adds another `apt-get update && apt-install` roundtrip. ccache is already there. This command could be invoked via triggering  dpkg-reconfigure but it makes the most sense to me to invoke exactly what we need. Especially for a workaround to an issue that should be pushed upstream.

nuclearsandwich:
@cottsay assuming you're still happy about this PR and assuming you're still online when the CI finishes you're welcome to merge this as I will likely be offline for the day when they come back in.

dirk-thomas:
> I like this approach less because it adds another apt-get update && apt-install roundtrip. ccache is already there.

This patch also adds an additional `RUN` line. While it doesn't do `apt` calls it is so far at the bottom of the Dockerfile that is get invalidated / redone very frequently where as a separate `ccache` installation will be cached as long as all the other installs like `build-essentials`.

PR #340, opened 2019-09-25
Add support for Cyclone DDS

ruffsl:
Context: https://github.com/ros2/ros2/issues/786


dirk-thomas:
Thanks for the contribution.

PR #352, opened 2019-11-01
Exclude rmw_cyclonedds_cpp from the extra rmw release nightly.

nuclearsandwich:
While this is the right nightly to eventually add rmw_cyclonedds_cpp it
seems that the rmw was introduced with test failures assuming it was not
covered by a nightly job.

This makes that assumption true until those test failures are addressed.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -314,7 +314,8 @@ def main(argv=None):
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
             'ignore_rmw_default': {
                 'rmw_connext_cpp',
-                'rmw_connext_dynamic_cpp'},
+                'rmw_connext_dynamic_cpp',
+                'rmw_cyclonedds_cpp',}
         })
 
         # configure nightly triggered job


dirk-thomas:
Since this is the only job testing CycloneDDS on non-Linux platforms I would rather keep it enabled to provide the developers feedback what is not working.

I assume the job isn't green without CycloneDDS either?

nuclearsandwich:
> I assume the job isn't green without CycloneDDS either?

They've been flaky since their creation but the Linux, Linux-arm64, macOS, and Windows jobs all had recent green builds before adding cycloneDDS.

The CycloneDDS packages were merged under the incorrect assumption that the known issues would not affect nightlies. If we now want the issues to affect nightlies that is a departure from what I read above.

dirk-thomas:
#787 contains specific suggestions what could be tried. Until that has happened and it doesn't work and there is no further input from upstream we could remove CycloneDDS - but not before.

nuclearsandwich:
Closing as stale.

PR #356, opened 2019-11-22
strip mention of (ROS 1) underlay workspace from setup.sh

dirk-thomas:
Since we don't want the (ROS 1) underlay be sourced automatically for the user.

* Linux (FastRTPS only): [Build Status](https://ci.ros2.org/job/ci_packaging_linux/273/)

### DIFF
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -131,7 +131,7 @@ def build_and_test_and_package(args, job):
                 lines[start_index:end_index] = []
                 with open(setup_file, 'wb') as h:
                     for line in lines:
-                        h.write(line + '\n')
+                        h.write(line + b'\n')
 
     # Only on Linux and OSX Python scripts have a shebang line
     if args.os in ['linux', 'osx']:
--- a/ros2_batch_job/packaging.py
+++ b/ros2_batch_job/packaging.py
@@ -119,15 +119,19 @@ def build_and_test_and_package(args, job):
             with open(setup_file, 'rb') as h:
                 content = h.read()
             lines = content.splitlines()
-            start_index = lines.index(b'# source chained prefixes')
-            end_index = lines.index(b'# source this prefix')
-            # remove lines starting with start_index until before end_index
-            print('- removing %d lines from %s' %
-                  (end_index - start_index, setup_file))
-            lines[start_index:end_index] = []
-            with open(setup_file, 'wb') as h:
-                for line in lines:
-                    h.write(line + '\n')
+            try:
+                start_index = lines.index(b'# source chained prefixes')
+            except ValueError:
+                pass
+            else:
+                end_index = lines.index(b'# source this prefix')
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This time updating all setup files: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/274/)

ros2_batch_job/packaging.py
```diff
@@ -111,6 +111,24 @@ def build_and_test_and_package(args, job):
...
 
+    # Only on Linux update the prefix level setup files
+    # to exclude the chained ROS 1 workspace used for the ros1_bridge
+    if args.os == 'linux':
+        print('# BEGIN SUBSECTION: remove ROS 1 underlay')
+        for setup_file in glob.glob(os.path.join(args.installspace, 'setup.*')):
+            with open(setup_file, 'rb') as h:
+                content = h.read()
+            lines = content.splitlines()
+            start_index = lines.index(b'# source chained prefixes')
```
jacobperron: Is there a case where the packaging job is triggered in a way that the underlay doesn't exist? Then this line would fail.

dirk-thomas: These searched for lines are always present in the setup file - even if no underlay is used. See https://github.com/colcon/colcon-core/blob/f742e9261bc1f75e78469cbc48e2963ceb82844f/colcon_core/shell/template/prefix_chain.sh.em#L33-L42

jacobperron: My workspace locally doesn't have the `# source chained prefixes` line.

https://github.com/colcon/colcon-core/blob/f742e9261bc1f75e78469cbc48e2963ceb82844f/colcon_core/shell/template/prefix_chain.sh.em#L31

dirk-thomas: Good point. Updated in 96328ef1713b517f5b04a0a29f387e7b471b6092.

ros2_batch_job/packaging.py
```diff
@@ -111,6 +111,24 @@ def build_and_test_and_package(args, job):
...
+            end_index = lines.index(b'# source this prefix')
+            # remove lines starting with start_index until before end_index
+            print('- removing %d lines from %s' %
+                  (end_index - start_index, setup_file))
+            lines[start_index:end_index] = []
+            with open(setup_file, 'wb') as h:
+                for line in lines:
+                    h.write(line + '\n')
```
nuclearsandwich: Simulating this locally I got an error concatenating str and bytes.

```suggestion
                    h.write(line + b'\n')
```

dirk-thomas: Indeed :angry:

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/275/)

nuclearsandwich:
I downloaded the archive from the above job and tested the bash, sh, and zsh. All of which I could source successfully.

PR #360, opened 2019-12-20
Add Linux bullet dependency

ahcorde:
Added bullet dependency

Related with this issue: https://github.com/ros2/geometry2/pull/205


mjcarroll:
Do you intend to follow-up with the other platforms?

ahcorde:
Yes, for Windows the package is here https://github.com/ros2/choco-packages/pull/16

In Mac OS you can install it using brew (but I think this should be done manually at least in the CI machines)

Documentation is updated here https://github.com/ros2/ros2_documentation/pull/481

PR #361, opened 2019-12-21
Dockerfile to create a dev env on Windows and run ros2_batch scr…

brawner:
Adding dockerfile and install scripts to be able to eventually run builds and tests in a docker container for windows.


brawner:
It looks like downloading OpenSSL might require regular link updates or hosting on a dedicated server. They bumped the build from "t" to "u". The Ros2 install instructions are nice because they link to the downloads page not the download itself.

Broken link
https://slproweb.com/download/Win64OpenSSL-1_0_2t.exe

Working link
https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe

brawner:
Compiler checker issue has been resolved by cloning ci code into docker image instead of mounting a volume from the host computer.

Master branch of fastrtps does not compile under the latest VS tools (VS 16,24). A [PR is pending](https://github.com/eProsima/Fast-RTPS/pull/932). For now, I've uploaded two versions. One for VS 2017 and one for VS 2019.  I've checked the 2017 version for build up to rclcpp for now.

brawner:
The Fastrtps PR has now been merged, so MSVC 2019 is now working. I've switched from vs_community to vs_buildtools, which works with a few more explicit packages. I've also removed .Net SDK as a separate installer because it can be installed through the Buildtools installer.

I also worked out the isolated_network issue. For some reason on windows `docker network create` requires a `-d` option. If "nat" is not suitable, a couple of others worked.

Cleaned up the PR a bit too and moved dockerfiles into the same directory.

MSVC 2019 up to rclcpp and tests
https://citest.ros2.org/job/ci_windows/68/

brawner:
This PR now depends on https://github.com/ros2/ci/pull/370

brawner:
This PR was separated into two. The goal of this PR is to be merge-able without affecting current ci.ros2.org build jobs.

brawner:
Removing WIP title, I think this is ready for review

brawner:
After updating my desktop to 1909, and updating the EC2 instance to server core 1909, it appears that issues regarding building debug builds in a mounted directory can be resolved by running the docker container with "--isolation=process". Therefore, I'm no longer basing this PR on #375

Example:
[Build Status](https://citest.ros2.org/job/nightly_win_deb/15/)

linux_docker_resources/.dockerignore
```diff
@@ -0,0 +1 @@
+*/.git
```
nuclearsandwich: What does this change relate to?

ros2_batch_job/__main__.py
```diff
@@ -521,15 +521,20 @@ def run(args, build_function, blacklisted_package_names=None):
...
+            pip_cmd.append('--force-reinstall')
+        job.run(pip_cmd + ['pip', 'setuptools',], shell=True)
         # Print setuptools version
         job.run(['"%s"' % job.python, '-c', '"import setuptools; print(setuptools.__version__)"'],
                 shell=True)
         # Print the pip version
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
-        pip_packages = list(pip_dependencies)
+        pip_packages = ['pip', 'setuptools', ] + list(pip_dependencies)
```
nuclearsandwich: I believe we were intentionally updating pip and setuptools before updating other packages so that the remaining packages were installed using the latest version of these utilities.

windows_docker_resources/Dockerfile.msvc2019
```diff
@@ -0,0 +1,151 @@
...
+
+# xmllint files
+ADD https://www.zlatkovic.com/pub/libxml/64bit/libxml2-2.9.3-win32-x86_64.7z C:\TEMP\libxml2.7z
+ADD https://www.zlatkovic.com/pub/libxml/64bit/zlib-1.2.8-win32-x86_64.7z C:\TEMP\zlib.7z
+ADD https://www.zlatkovic.com/pub/libxml/64bit/iconv-1.14-win32-x86_64.7z C:\TEMP\iconv.7z
+
+# Restore the default Windows shell for correct batch processing.
+SHELL ["cmd", "/S", "/C"]
+
+# 3010 is an acceptable exit code, but it will confuse docker
```
nuclearsandwich: Can you elaborate on _Why_ 3010 is an acceptable return code?

windows_docker_resources/Dockerfile.msvc2019
```diff
@@ -0,0 +1,151 @@
...
+  TargetDir=C:\Python37 `
+  PrependPath=1 `
+  Include_debug=1 `
+  Include_symbols=1
+
+# Install Chocolatey by powershell script
+RUN powershell -noexit "Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))"
+
+# choco installs
+RUN choco install -y cmake curl git KB3035131 KB3033929 KB2919442 vcredist2013 vcredist140 cppcheck
```
nuclearsandwich: Can we document what these KB payloads are and why they are needed?

windows_docker_resources/Dockerfile.msvc2019
```diff
@@ -0,0 +1,151 @@
...
+RUN choco install -y -s C:\TEMP asio cunit eigen tinyxml-usestl tinyxml2 log4cxx
+RUN choco install -y patch
```
nuclearsandwich: Why is patch installed separately from vcredist* and cppcheck above?

linux_docker_resources/.dockerignore
brawner: I'm not sure why it was added. Removed

ros2_batch_job/__main__.py
brawner: That's a good point, and they still are in line 530. This workaround came from the issue that --force-reinstall on the pip_dependencies list caused setuptools to be reinstalled as well. But for some reason, the pip wheel was not found if they were done separately. This was the only work around I could find.

windows_docker_resources/Dockerfile.msvc2019
brawner: Added a comment. Basically, it just means that the install was successful, but a restart is required.

windows_docker_resources/Dockerfile.msvc2019
brawner: choco installs are sometimes flakey, and these KB's were needed at one point. They don't look like they are needed anymore with 1909. Removed

windows_docker_resources/Dockerfile.msvc2019
brawner: It didn't work for some reason in the past, but seems to work fine when combined now...

ros2_batch_job/__main__.py
brawner: It looks like the error I was running into no longer occurs.

brawner: Reverting this section

PR #362, opened 2019-12-24
[Question] Remove top level COLCON_IGNORE file before archiving?

mikaelarguedas:
The presence of this file prevents catkin_pkg from finding these packages if `ROS_PACKAGE_PATH` points to the root of the extracted archive.
Removing this also an install from archive to be closer to an install from debs.

Wiht this PR, we suggest to remove it as it doesn't seem to serve the purpose it does in a full workspace with a `src` directory.

More details at https://github.com/osrf/docker_images/pull/345


dirk-thomas:
Thanks for the patch.

PR #367, opened 2020-01-13
bump Jenkins plugin versions

dirk-thomas:
Just matching the plugin versions deployed on ci.ros2.org.


mjcarroll:
I can confirm that updated versions match ci.ros2.org

dirk-thomas:
@nuclearsandwich can you please review this when you have a minute. Thanks.

PR #370, opened 2020-01-16
Add CL arguments for source/build/install space

brawner:
This adds command-line arguments to run_ros2_batch.py for customized source/build/install space directories, so they can be passed to different subsections appropriately.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -258,9 +258,11 @@ def get_args(sysargv=None):
     args = parser.parse_args(argv)
     args.build_args = build_args
     args.test_args = test_args
-    for name in ('sourcespace', 'buildspace', 'installspace'):
-        if name in args.white_space_in and getattr(args, name) != parser.get_default(name):
-            raise Exception('Argument {} and "--white-space-in" cannot both be used'.format(name))
+
+    if args.white_space_in is not None:
+        for name in ('sourcespace', 'buildspace', 'installspace'):
+            if name in args.white_space_in and getattr(args, name) != parser.get_default(name):
+                raise Exception('Argument {} and "--white-space-in" cannot both be used'.format(name))
     return args
 
 
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -183,7 +183,7 @@ def get_args(sysargv=None):
              "doesn't exist fall back to the default branch (default: latest "
              'release)')
     parser.add_argument(
-        '--white-space-in', nargs='*', default=None,
+        '--white-space-in', nargs='*', default=[],
         choices=['sourcespace', 'buildspace', 'installspace', 'workspace'],
         help="which folder structures in which white space should be added")
     parser.add_argument(
@@ -259,10 +259,9 @@ def get_args(sysargv=None):
     args.build_args = build_args
     args.test_args = test_args
 
-    if args.white_space_in is not None:
-        for name in ('sourcespace', 'buildspace', 'installspace'):
-            if name in args.white_space_in and getattr(args, name) != parser.get_default(name):
-                raise Exception('Argument {} and "--white-space-in" cannot both be used'.format(name))
+    for name in ('sourcespace', 'buildspace', 'installspace'):
+        if name in args.white_space_in and getattr(args, name) != parser.get_default(name):
+            raise Exception('Argument {} and "--white-space-in" cannot both be used'.format(name))
     return args
 
 
@@ -427,7 +426,6 @@ def run(args, build_function, blacklisted_package_names=None):
 
     job = None
 
-    args.white_space_in = args.white_space_in or []
     args.workspace = 'work space' if 'workspace' in args.white_space_in else 'ws'
     # Add white space to path. Logic above should prevent conflict of manually specified path and '--white-space-in'
     for name in ('sourcespace', 'buildspace', 'installspace'):
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -239,6 +239,15 @@ def get_args(sysargv=None):
     parser.add_argument(
         '--visual-studio-version', default=None, required=(os.name == 'nt'),
         help='select the Visual Studio version')
+    parser.add_argument(
+        '--source-space', default='src', dest='sourcespace',
+        help='source directory path')
+    parser.add_argument(
+        '--build-space', default='build', dest='buildspace',
+        help='build directory path')
+    parser.add_argument(
+        '--install-space', default='install', dest='installspace',
+        help='install directory path')
 
     argv = sysargv[1:] if sysargv is not None else sys.argv[1:]
     argv, build_args = extract_argument_group(argv, '--build-args')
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -408,9 +417,11 @@ def run(args, build_function, blacklisted_package_names=None):
 
     args.white_space_in = args.white_space_in or []
     args.workspace = 'work space' if 'workspace' in args.white_space_in else 'ws'
-    args.sourcespace = 'source space' if 'sourcespace' in args.white_space_in else 'src'
-    args.buildspace = 'build space' if 'buildspace' in args.white_space_in else 'build'
-    args.installspace = 'install space' if 'installspace' in args.white_space_in else 'install'
+    for dir_name in [args.source_space, args.build_space, args.install_space]:
+        dir_name.rstrip('/' + os.sep)
```
cottsay: `rstrip` isn't an in-place change - it returns the modified value.

ros2_batch_job/__main__.py
```diff
@@ -239,6 +239,15 @@ def get_args(sysargv=None):
     parser.add_argument(
         '--visual-studio-version', default=None, required=(os.name == 'nt'),
         help='select the Visual Studio version')
+    parser.add_argument(
+        '--source-space', default='src',
```
dirk-thomas: Please use `dest='sourcespace'` to avoid having to renamed the used `args`. That will make the patch much more concise.

Same for the other two options.

ros2_batch_job/__main__.py
brawner: Gah. Added a nested function instead

ros2_batch_job/__main__.py
brawner: Done

ros2_batch_job/__main__.py
```diff
@@ -407,10 +416,14 @@ def run(args, build_function, blacklisted_package_names=None):
...
+        if space_name in args.white_space_in:
+            return path.rstrip('/' + os.sep) + ' space'
+        return path
     args.workspace = 'work space' if 'workspace' in args.white_space_in else 'ws'
-    args.sourcespace = 'source space' if 'sourcespace' in args.white_space_in else 'src'
-    args.buildspace = 'build space' if 'buildspace' in args.white_space_in else 'build'
-    args.installspace = 'install space' if 'installspace' in args.white_space_in else 'install'
+    args.sourcespace = add_white_space_to_path(args.sourcespace, 'sourcespace')
+    args.buildspace = add_white_space_to_path(args.buildspace, 'buildspace')
+    args.installspace = add_white_space_to_path(args.installspace, 'installspace')
```
dirk-thomas: How about simplifying the logic to something like this:

```
for name in ('sourcespace', 'buildspace', 'installspace'):
    if name in args.white_space_in:
        value = getattr(args, name)
        assert value.endswith('space')
        value = value[:-5] + ' ' + value[-5:]
        setattr(args, name, value)
```

brawner: Wouldn't that assertion fail for the default values of source/build/install space? I was trying to make it so that neither white_space_in, nor the source/build/install-space arguments had precedence over one another.

So the following cases would still work:
'build' -> 'build space'
'/path/to/build' -> '/path/to/build space'
'/path/to/build/' -> '/path/to/build space'

Or on windows:
'C:\\path\\to\\build\\' -> 'C:\\path\\to\\build\\'
'C:/path/to/build/' -> 'C:/path/to/build space'

dirk-thomas: The assertion was just an example. Feel free to use any other approach to determine the modified `value`. My suggestion was mostly about the overall logic.

brawner:
I made the requested changes, but now I'm rethinking the conflict of --white-space-in and --source/build/install-space arguments. Would it be better to put in logic that prevents a path from being set, but also modified through --white-space-in? Seems like any use case that wants to overwrite the path, would not want that path to change internally.

dirk-thomas:
> Would it be better to put in logic that prevents a path from being set, but also modified through --white-space-in? Seems like any use case that wants to overwrite the path, would not want that path to change internally.

That sounds even better to me too. :+1:

brawner:
Added logic to `get_args()` to prevent `--white-space-in` and `--source/build/install-space` arguments.

ros2_batch_job/__main__.py
```diff
@@ -249,6 +258,9 @@ def get_args(sysargv=None):
     args = parser.parse_args(argv)
     args.build_args = build_args
     args.test_args = test_args
+    for name in ('sourcespace', 'buildspace', 'installspace'):
+        if name in args.white_space_in and getattr(args, name) != parser.get_default(name):
```
dirk-thomas: Technically this isn't a sufficient check if the user passed an argument explicitly. It could very well be passed explicitly but have the same value as the default.

That being said I think it is good enough for this use case.

brawner: Fair point. I could use `default=argparse.SUPPRESS` to catch when it is or isn't provided, but then we would be going back to adding fields to args after its been parsed and putting the defaults outside of the argument parser. This was the best balance between expected behavior and maintainability that I could think of.

cottsay:
What if we just defaulted the three new arguments to `None`? Then you'd be able to reliably tell if they were specified or not.

Also, later on, you could maintain the existing behavior of `buildspace` vs `build space` instead of the currently proposed logic which can result in `buildspace space`.

ros2_batch_job/__main__.py
dirk-thomas: Just for context: this is what I have done in colcon for the exact same problem: colcon/colcon-core#288

But as I said this can stay as is.

brawner:
Squashed and rebased on master for testing. First build failed

[Build Status](https://ci.ros2.org/job/ci_linux/9106/)

brawner:
This failed because args.white_space_in is defaulted to None. Should I set the default to [], since line 430 does that anyway? Or should I add the check that I just committed?

dirk-thomas:
> Should I set the default to []

That sounds like the best option to me - no idea why it should default to `None`.

brawner:
[Build Status](https://ci.ros2.org/job/ci_linux/9109/)
[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4843/)
[Build Status](https://ci.ros2.org/job/ci_osx/7444/)
[Build Status](https://ci.ros2.org/job/ci_windows/9039/)

brawner:
I just uploaded a commit that changes the argument parser logic so that source/build/install space defaults are set to None. I'm not sure this is better this way though because other parts of the logic get more complex. @cottsay, which one did you have a preference for?

ros2_batch_job/__main__.py
```diff
@@ -413,11 +438,12 @@ def run(args, build_function, blacklisted_package_names=None):
 
     job = None
 
-    args.white_space_in = args.white_space_in or []
     args.workspace = 'work space' if 'workspace' in args.white_space_in else 'ws'
-    args.sourcespace = 'source space' if 'sourcespace' in args.white_space_in else 'src'
-    args.buildspace = 'build space' if 'buildspace' in args.white_space_in else 'build'
-    args.installspace = 'install space' if 'installspace' in args.white_space_in else 'install'
+    # Add white space to path. Logic above should prevent conflict of manually specified path and '--white-space-in'
```
cottsay: I think we can remove all of this logic now, right?

brawner: Correct, I removed stuff

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9148/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4869/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9073/)

PR #374, opened 2020-01-24
Use subprocess_reaper.py instead of trying to kill all colcon processes.

nuclearsandwich:
Killing all colcon.exe processes wasn't sufficient to resolve this issue
fully since some of the dangling processes were python3 nodes or the
ros2cli daemon, or python3 pytest runners rather than colcon processes.

Switching to subprocess_reaper.py (borrowed from ros_buildfarm) allowed several test jobs to be aborted and rebuilt without interference. Previously, aborting jobs on Windows was all but guaranteed to cause dangling processes resulting in failure of the next job on the host.


nuclearsandwich:
[Build Status](https://ci.ros2.org/job/test_ci_windows/123/)

ros2_batch_job/__main__.py
```diff
@@ -150,13 +150,12 @@ def main(sysargv=None):
...
-    # If on Windows, kill any still running `colcon` processes to avoid
-    # problems when trying to delete files from pip or the workspace during
-    # this job.
+    # If on Windows, run the subprocess reaper to kill any still running job
+    # processes to avoid problems when trying to delete files from pip or the
+    # workspace during this job.
     if sys.platform == 'win32':
-        os.system('taskkill /f /im colcon.exe')
+        import subprocess
+        subprocess.Popen(['python.exe', 'subprocess_reaper.py', str(os.getpid())])
```
wjwwood: Would `sys.executable` make more sense than just `python.exe`? I actually don't know.

nuclearsandwich: Doesn't seem unreasonable to me. I'll make the change.

dirk-thomas:
I don't think copying the code here is a good idea. I would rather like to see the `ros_buildfarm` package being used here than continue to maintain that logic in multiple places.

dirk-thomas:
Instead of installing a released version of `ros_buildfarm` installing the latest state from the git repo would allow us to iterate quicker and not require new releases in case we want to tweak any of that logic.

nuclearsandwich:
> Instead of installing a released version of `ros_buildfarm` installing the latest state from the git repo would allow us to iterate quicker and not require new releases in case we want to tweak any of that logic.

I arrived at the same conclusion since the updated version of the script isn't released yet.

ros2_batch_job/__main__.py
```diff
@@ -150,13 +150,12 @@ def main(sysargv=None):
...
-    # If on Windows, kill any still running `colcon` processes to avoid
-    # problems when trying to delete files from pip or the workspace during
-    # this job.
+    # If on Windows, run the subprocess reaper to kill any still running job
+    # processes to avoid problems when trying to delete files from pip or the
+    # workspace during this job.
     if sys.platform == 'win32':
-        os.system('taskkill /f /im colcon.exe')
+        subprocess_reaper_path = which('subprocess_reaper.py')
+        subprocess.Popen([sys.executable, subprocess_reaper_path, str(os.getpid())])
```
dirk-thomas: I think the process should be detached to prevent it being killed before it can do its job once this script exits, like here: https://github.com/ros2/ros2cli/blob/c00dec0a72c049d3a4a8a80f1324ea24dc8373c6/ros2cli/ros2cli/node/daemon.py#L51-L67

Also piping the output of the script into a file would be good for debugging: like https://github.com/ros-infrastructure/ros_buildfarm/blob/a322b0793986fa3b83feb794a0daca8c781fc4c2/ros_buildfarm/templates/ci/ci_job.xml.em#L151

nuclearsandwich: > I think the process should be detached to prevent it being killed before it can do its job once this script exits, like here: [ros2/ros2cli:ros2cli/ros2cli/node/daemon.py@`c00dec0`#L51-L67](https://github.com/ros2/ros2cli/blob/c00dec0a72c049d3a4a8a80f1324ea24dc8373c6/ros2cli/ros2cli/node/daemon.py#L51-L67)

The subprocess_reaper script [ignores SIGTERM](https://github.com/ros-infrastructure/ros_buildfarm/blob/a322b0793986fa3b83feb794a0daca8c781fc4c2/scripts/subprocess_reaper.py#L172) but looking at the log file it seems that Windows is sending SIGINT or similar instead causing a `KeyboardInterrupt`. I would like to keep it attached but to do so I need to make it un-^C-able on Windows. 


> Also piping the output of the script into a file would be good for debugging: like [ros-infrastructure/ros_buildfarm:ros_buildfarm/templates/ci/ci_job.xml.em@`a322b07`#L151](https://github.com/ros-infrastructure/ros_buildfarm/blob/a322b0793986fa3b83feb794a0daca8c781fc4c2/ros_buildfarm/templates/ci/ci_job.xml.em#L151)

Done.

nuclearsandwich:
Closing as stale.

Issue #365, opened 2020-01-11
Transition Windows/Mac CI farms to cloud virtualization


brawner:
This issue tracks the progress of transitioning the Windows CI/Mac CI machines to cloud virtualization. The goal of the project is to create a more reproducible, deployable and scalable Jenkins CI farm. Because Microsoft has embraced containerization in the last couple of years, it is feasible to run Windows CI instances inside docker containers running on VMs from a cloud provider. On the mac side, MacStadium offers potential for virtualization with mac instances.

Window cloud virtualization tasks:
- [x]  Demonstrate a CI build by running run_ros2_batch.py from a docker container 
- [x]  Deploy a Jenkin's build job to an EC2 instance
    https://citest.ros2.org/job/ci_windows/87/
- [x]  Merge windows_docker_resources PR (https://github.com/ros2/ci/pull/361) (1/31/2019)
- [x]  Cloud Jenkins agents running side-by-side with current CI farm (1/31/2019)
- [x]  Windows configuration management of EC2 VMs with Chef (1/31/2019)
- [x]  Retire bare metal CI Windows servers (3/1/2019)

Mac Cloud Virtualization tasks (TBD)
- [ ]  Investigate MacStadium's cloud virtualization offering
- [ ]  Demostrate CI build by running run_ros2_batch.py from virtual mac instance
- [ ]  Deploy a Jenkin's build job to a MacStadium instance
- [ ]  Create and Merge Mac CI PR
- [ ]  Cloud Jenkins agents running side-by-side with current CI farm
- [ ]  Mac configuration management of MacStadium VMs with Orka
- [ ]  Retire bare metal CI Mac servers

brawner:
I've run into an issue attempting to build in a directory mounted with `docker run -v`. When building, cmake runs a compiler check on a simple program in debug mode and with the `/Zi` build flag. This creates a pdb file with debug symbols. Any references to the debug symbols are requested through mspdbsrv.exe. I suspect this has to do with file handle access across the mounted directory, between the containerized OS and the host OS. Building in a directory that's not in a mounted location, but copying build/install results back to the mounted directory seems to work.

Example of output failure.
```
cl /c /Zi /W3 /WX- /diagnostics:classic /Od /Ob0 /D WIN32 /D _WINDOWS /D "CMAKE_INTDIR=\"Debug\"" /D _MBCS /Gm- /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Z
[...]

PR #375, opened 2020-01-27
Add CL arguments for building in a temporary directory

brawner:
This PR adds CL arguments for building inside a temporary directory. When building inside a docker container, CMake's compiler checker runs various debug builds that fail if built inside a mounted directory. The gist of the issue is that the debug symbol files (.PDB) are served by a mspdbsrv.dll, and this has troubles when trying to access files in the mounted directory from the docker container.

Overriding --build-args/test-args when launching run_ros2_batch.py is insufficient because they are not passed to `colcon test-result`.

This depends on #370, and is depended on by WIP PR #361.

New commit: https://github.com/ros2/ci/commit/cd417a5d9d29a5281a31e34cd22c5158aa57cf75

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -777,8 +777,7 @@ def run(args, build_function, blacklisted_package_names=None):
             print('Moving buildspace "{}" failed with error.\n{}'.format(args.buildspace, repr(e)), file=sys.stderr)
 
         try:
-            if os.path.exists(temp_workspace):
-                shutil.rmtree(temp_workspace, ignore_errors=True)
+            shutil.rmtree(temp_workspace, ignore_errors=True)
         except PermissionError as e:
             print('Cleanup of temporary build dir "{}" failed with error.\n{}'.format(temp_workspace, repr(e)), file=sys.stderr)
         print('# END SUBSECTION')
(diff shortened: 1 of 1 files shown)


brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9149/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4870/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7472/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9074/)

dirk-thomas:
> When building inside a docker container, CMake's compiler checker runs various debug builds that fail if built inside a mounted directory. The gist of the issue is that the debug symbol files (.PDB) are served by a mspdbsrv.dll, and this has troubles when trying to access files in the mounted directory from the docker container.

Can you point to any references for this problem?

On Jenkins we explicitly exclude the workspace from Defender to avoid concurrent access. I am not sure if that is / should be the case here.

brawner:
This is where I have documented the issue, https://github.com/ros2/ci/issues/365#issuecomment-573858026

Some other references:
https://github.com/docker/for-win/issues/829#issuecomment-397149736
https://stackoverflow.com/questions/56877220/how-do-i-set-compile-options-for-cmake-msvc-build-in-docker

One of the references mentions adding `--isolation=process`, which I think I tried, but I forgot the result of that test.

brawner:
This issue was resolved by switching to 1909 and using `docker run --isolation=process`

PR #380, opened 2020-01-31
remove OpenSplice from nightly job

dirk-thomas:
Follow up of https://discourse.ros.org/t/opensplice-move-to-eclipse-cyclone-dds-in-ros-2/12370/


nuclearsandwich:
The configuration change was not in effect when I ran create_jenkins_job.py for other changes. Based on the config history it doesn't appear that this change was previously applied. Nightlies up until today's were still including opensplice but will not any longer.

PR #381, opened 2020-02-03
Restrict Qt5 LTS versions to known compatible versions (windows docker builds)

brawner:
Due to a recent update to Qt5, there is a yet-to-be discovered breaking change between 5.12.7 to 5.12.6. However, 5.12.6 is no longer available to be installed (5.12.5 is available however). This adds a parameter to the qt-installer script to improve the regex search of compatible LTS versions as well as restricting the current docker build to 5.12.1-6.

Failed build:
[Build Status](https://ci.ros2.org/job/ci_windows-container/12/)


brawner:
Build with this code fix
[Build Status](https://ci.ros2.org/job/ci_windows-container/29/)

nuclearsandwich:
@brawner we can close this now that we're using the latest Qt release right? An alternative would be turning the PR into a commented out example of how to pin to a specific release of Qt if we ever need to do so.

brawner:
By using the latest patch version 5.14.1, the build does not fail for whatever reason. Closing this PR.

PR #387, opened 2020-02-20
Pinning cppcheck to 1.89 for windows container builds

brawner:
To reduce failed cppchecks due to version 1.90, this pins the docker container to version 1.89.

Failing example:
[Build Status](https://ci.ros2.org/job/ci_windows-container/165/)

Working:
[Build Status](https://ci.ros2.org/job/ci_windows-container/175/)

### DIFF
--- a/windows_docker_resources/Dockerfile.msvc2019
+++ b/windows_docker_resources/Dockerfile.msvc2019
@@ -51,7 +51,8 @@ RUN C:\TEMP\python-37.exe /quiet `
 RUN powershell -noexit "Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))"
 
 # choco installs
-RUN choco install -y cmake curl git vcredist2013 vcredist140 cppcheck patch
+RUN choco install -y cmake curl git vcredist2013 vcredist140 patch
+RUN choco install -y cppcheck --version 1.89
 RUN choco install -y -s C:\TEMP asio cunit eigen tinyxml-usestl tinyxml2 log4cxx
 
 RUN C:\TEMP\Win64OpenSSL.exe /VERYSILENT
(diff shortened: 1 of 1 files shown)


brawner:
Most of the cppcheck failures were due to missing include directories in the cppcheck. Those were addressed in:
https://github.com/ros2/rclcpp/pull/1003
https://github.com/ros2/system_tests/pull/400
https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/140

There remains one issue with this [if statement](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/clock.cpp#L142) which appears to be an open bug, with cppcheck. https://trac.cppcheck.net/ticket/9537

The following code also throws the same error, but otherwise compiles fine.
```
JumpHandler::SharedPtr ptr(handler.release(), [weak_clock](JumpHandler * handler) {
    if (0) { }
  });
```

tfoote:
Permalink since the if statement referenced above has moved slightly: 

https://github.com/ros2/rclcpp/blob/f30329fbec79cb699f41a37cc27c0c475de3edc1/rclcpp/src/rclcpp/clock.cpp#L157

chapulina:
This is running CI for branches targetting Foxy, right? Wouldn't it be better to detect cppcheck failures sooner rather than later? (i.e. keep the latest version)

In fact, it looks like all the known issues have been addressed already? Except for `rclcpp.cppcheck.error: internalAstError (src\rclcpp\clock.cpp:142)`, which is an upstream bug, so I think it could be suppressed inline?

brawner:
This PR only affects Windows, which does a `choco upgrade all`, and thus gets the latest version. It's only meant as a temporary fix until cppcheck fixes their bug. I would also be in favor of inline suppression. It looks like in order to make that happen, `--inline-suppr` has to be used as part of that command. Does anyone know if that's already being done and inline suppression is possible?

chapulina:
It looks like `rclcpp` already has some suppressions, so my guess is that they're having an effect.

```
rclcpp$ grep -r "cppcheck-supp" .
./rclcpp/include/rclcpp/duration.hpp:  // cppcheck-suppress noExplicitConstructor
./rclcpp/include/rclcpp/duration.hpp:  // cppcheck-suppress noExplicitConstructor
./rclcpp/include/rclcpp/duration.hpp:  // cppcheck-suppress operatorEq // this is a false positive from cppcheck
./rclcpp/include/rclcpp/qos.hpp:  // cppcheck-suppress noExplicitConstructor
```

brawner:
This false positive may have fixed itself. I don't see it coming up on rclcpp anymore on the windows builds.

https://ci.ros2.org/view/nightly/job/nightly_win_rel/1490/testReport/rclcpp/cppcheck/

I was going to close it anyway in favor of suppressing the warning, but I guess that's not needed.

PR #388, opened 2020-02-21
remove Xenial support

dirk-thomas:
Fixing an issue with Focal / Python 3.8:

```
  File "run_ros2_batch.py", line 32, in <module>
    sys.exit(main())
  File "/home/jenkins-agent/workspace/ci_linux/ros2_batch_job/__main__.py", line 153, in main
    if sys.platform.lower().startswith('linux') and platform.linux_distribution()[2] == 'xenial':
AttributeError: module 'platform' has no attribute 'linux_distribution'
```

Before: [Build Status](https://ci.ros2.org/job/ci_linux/9408/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/9409/)


dirk-thomas:
> I'm mildly curious whether the undefined attribute is an  artifact of Focal's pre-release status or if it's something we can  expect in the released version as well.

The function has simply been removed in Python 3.8 after being deprecated since Python 3.5: https://docs.python.org/2/library/platform.html#platform.linux_distribution

PR #390, opened 2020-02-26
Only run Mac packaging jobs on hosts running mojave.

nuclearsandwich:
mini3 hasn't migrated as we needed to sort out the issues with mini1 and
mini2 but now that both Lore and mini2 (and hopefully soon mini1) are operational we should prevent new packaging jobs from running on hosts that aren't running Mojave.


nuclearsandwich:
There are multiple issues with this change as-is. I am working on the resolutions for them now.

PR #392, opened 2020-02-27
Support to skip credentials configuration in Jenkins jobs

j-rivero:
A new `--avoid-credentials` option to skip credentials setup in Jenkins for private repositories. Useful for testing buildfarms with no need to access to the `rticonnext-src` private repo. 

Changes:
  * Skip submodule checkout (maybe too aggressive but did not find a bettter solution)
  * Skip the ssh-agent jenkins config

Not related to the new parameter but needed to keep docker `ADD` happy:
  * Create fake license.dat in presence of `CI_USE_CONNEXT_DEBS` var

I'm testing the changes here: [Build Status](https://citest2.ros2.org/job/ci_linux/16/)

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -45,7 +45,7 @@
     <userRemoteConfigs>
       <hudson.plugins.git.UserRemoteConfig>
         <url>@(ci_scripts_repository)</url>
-        <credentialsId>1c2004f6-2e00-425d-a421-2e1ba62ca7a7</credentialsId>
+        <credentialsId>@[if not avoid_credentials]1c2004f6-2e00-425d-a421-2e1ba62ca7a7@[end if]</credentialsId>
       </hudson.plugins.git.UserRemoteConfig>
     </userRemoteConfigs>
     <branches>
@@ -60,7 +60,7 @@
     <submoduleCfg class="list"/>
     <extensions>
       <hudson.plugins.git.extensions.impl.SubmoduleOption>
-        <disableSubmodules>false</disableSubmodules>
+        <disableSubmodules>@[if avoid_credentials]true@[else]false@[end if]</disableSubmodules>
         <recursiveSubmodules>true</recursiveSubmodules>
         <trackingSubmodules>false</trackingSubmodules>
         <reference/>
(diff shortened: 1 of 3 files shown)


job_templates/ci_job.xml.em
```diff
@@ -45,7 +45,7 @@
     <userRemoteConfigs>
       <hudson.plugins.git.UserRemoteConfig>
         <url>@(ci_scripts_repository)</url>
-        <credentialsId>1c2004f6-2e00-425d-a421-2e1ba62ca7a7</credentialsId>
+        <credentialsId>@[if not avoid_credentials]1c2004f6-2e00-425d-a421-2e1ba62ca7a7@[end if]</credentialsId>
```
dirk-thomas: Nitpick: you can avoid the bracket syntax here: `@('1c2004f6-2e00-425d-a421-2e1ba62ca7a7' if not avoid_credentials else '')`.

job_templates/ci_job.xml.em
```diff
@@ -152,6 +152,8 @@ fi
 if [ "$CI_USE_CONNEXT_DEBS" = "true" ]; then
   export CI_ARGS="$CI_ARGS --connext-debs"
   export DOCKER_BUILD_ARGS="$DOCKER_BUILD_ARGS --build-arg INSTALL_CONNEXT_DEBS=$CI_USE_CONNEXT_DEBS"
+  # need to create a fake license so the dockerfile does not fail when using ADD. It won't be used.
+  echo "fake" > linux_docker_resources/rticonnextdds-license/rti_license.dat
```
dirk-thomas: Should this be made conditional on `avoid_credentials`?

job_templates/ci_job.xml.em
```diff
@@ -60,7 +60,7 @@
     <submoduleCfg class="list"/>
     <extensions>
       <hudson.plugins.git.extensions.impl.SubmoduleOption>
-        <disableSubmodules>false</disableSubmodules>
+        <disableSubmodules>@[if avoid_credentials]true@[else]false@[end if]</disableSubmodules>
```
dirk-thomas: I could see this becoming a problem if any other repo requires a submodule in the future...

dirk-thomas: Nitpick: you can avoid the bracket syntax here: `@('true' if avoid_credentials else 'false')`.

claireyywang:
@j-rivero friendly ping

j-rivero:
> @j-rivero friendly ping

I'm going to close the PR until we found a way of avoiding to checkout the submodule. Note that this is done by Jenkins before running any code, in the SCM checkout step, so our options to handle it are reduced since the credentials should be in place before running the scripts.

PR #393, opened 2020-02-28
Update Jenkins plugin xunit: from 1.102 to 2.3.8

j-rivero:
The bump of of xunit from 1.x to 2.x series will break the current builds since the format used by pytest in the builds is not compatible with the checks imposed in the 2.x series of the plugin. To fix this I'm passing `-o junit_family=xunit2` to `colcon test --pytest-args` that brings new XML format to the result.

It fixes the problem in my testing buildfarm using `ci_linux` job on `ament_copyright` package.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -377,11 +377,13 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: test')
 
+    # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     test_cmd = [
         args.colcon_script, 'test',
         '--base-paths', '"%s"' % args.sourcespace,
         '--build-base', '"%s"' % args.buildspace,
         '--install-base', '"%s"' % args.installspace,
+        '--pytest-args', '-o junit_family=xunit2',
     ]
     if not args.isolated:
         test_cmd.append('--merge-install')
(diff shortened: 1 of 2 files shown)


ros2_batch_job/__main__.py
```diff
@@ -377,11 +377,13 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: test')
 
+    # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     test_cmd = [
         args.colcon_script, 'test',
         '--base-paths', '"%s"' % args.sourcespace,
         '--build-base', '"%s"' % args.buildspace,
         '--install-base', '"%s"' % args.installspace,
+        '--pytest-args', '-o junit_family=xunit2',
```
dirk-thomas: How does this behave when the user chooses to pass custom `--pytest-args`?

ros2_batch_job/__main__.py
```diff
@@ -377,6 +377,11 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: test')
 
+    # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
+    pytest_ini = open("pytest.ini", "w")
```
dirk-thomas: Please use a context manager: `with open(...) as h:`.

Nitpick: also use single quotes.

ros2_batch_job/__main__.py
j-rivero: During my tests colcon ended up with two `--pytest-args` if one was supplied as user input parameter. This does not work aggregating options so it was not a valid solution. I changed the code to use a [ini file with the configuration](https://docs.pytest.org/en/latest/deprecations.html#junit-family-default-value-change-to-xunit2). Tests are happy.

ros2_batch_job/__main__.py
j-rivero: thanks, 0bb3e24

PR #396, opened 2020-03-02
enable CycloneDDS as a default RMW

dirk-thomas:
In order to fulfill tier-1 requirement.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -78,7 +78,6 @@ def main(argv=None):
         'mailer_recipients': '',
         'ignore_rmw_default': {
             'rmw_connext_dynamic_cpp',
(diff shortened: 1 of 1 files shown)


brawner:
Not sure if these issues are showstoppers for enabling default cyclonedds, but I had found a couple of cyclonedds specific issues that affect multiple platforms.

https://github.com/ros2/ros2cli/issues/456
https://github.com/ros2/ros2cli/issues/457

dirk-thomas:
> Not sure if these issues are showstoppers for enabling default cyclonedds,

Those should for sure be addressed but as long as they don't break the build / prevent us from getting test results / flood the result with numerous test failures I don't think they are showstoppers.

PR #397, opened 2020-03-03
Remove bare metal windows configs and set containers as default

brawner:
This PR sets windows-containers as the only windows configuration. Unless they are deleted from the ci.ros2.org, the win jobs will remain because this PR renames all windows jobs to use 'windows' in their job names.

This will need to be checked before being committed to ci.ros2.org.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -108,12 +108,12 @@ def main(argv=None):
             'ci_scripts_repository': args.ci_scripts_repository.replace(
                 'git@github.com:', 'https://github.com/'),
         },
-        'windows': {
+        'windows-metal': {
             'label_expression': 'windows',
             'shell_type': 'BatchFile',
             'use_isolated_default': 'false',
         },
-        'windows-container': {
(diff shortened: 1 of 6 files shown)


nuclearsandwich:
If we decide to remove all the name shortening logic we should also manually rename the existing jobs to keep their history.

dirk-thomas:
I though the idea was to *not* change the naming and break all existing links?

nuclearsandwich:
> I though the idea was to _not_ change the naming and break all existing links?

Right. This. I knew there was a reason renaming wouldn't be sufficient.

brawner:
Ok, so we need to leave ci_windows as is then? Same thing with the packaging jobs?

brawner:
Ok this has been updated. Everything referencing the 'windows' os-config is now 'windows-metal', and 'windows-container' is the new windows.

I've also disabled nightly and packaging jobs for windows-metal.

nuclearsandwich:
This is deployed and the `-container` job variants have been disabled.

PR #385, opened 2020-02-13
Use venv module for virtualenv creation.

nuclearsandwich:
Alternative to #384.

The venv module was new in Python 3.3 and contains provides standard
library support for virtualenv creation.

Recent changes in virtualenv 20.0 have made it desireable to consider a
stable virtual environment creation module.

PR #399, opened 2020-03-04
Pin virtualenv 16.7.9 to fix Python bin installation.

nuclearsandwich:
virtualenv>=20 and venv both set `sys.base_prefix` to a value outside the virtual environment whereas virtualenv 16.7.9 does not.
This triggers virtualenv detection in setuptools/distutils[1] which causes them to ignore our configured install-scripts options.

Since this issue is causing test failures and issues with our distributed binary archives (ros2/build_farmer#266), I think it's worth pinning a version of virtualenv that meets our needs while we explore alternatives.

[1]: https://github.com/pypa/setuptools/blob/a5dec2f14e3414e4ee5dd146bff9c289d573de9a/setuptools/dist.py#L573-L579

This PR affects a revert of #385 but not because venv itself was unsuitable. We see the same issue using current versions of virtualenv which was totally rewritten for the 20.x series.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -33,7 +33,7 @@ RUN echo "deb http://packages.osrfoundation.org/gazebo/ubuntu `lsb_release -cs` 
 RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 
 # Install some development tools.
-RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool python3-venv
+RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9530/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5239/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7785/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9455/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/276/)

nuclearsandwich:
> Grudging approval.

I know we've been chatting offline, but for the record this is grudging because the solution is a pin to an old version not because there's anything unsound with the PR itself?

mjcarroll:
> I know we've been chatting offline, but for the record this is grudging because the solution is a pin to an old version not because there's anything unsound with the PR itself?

Yes, sorry, nothing unsound about the PR.  I'm mainly just grumpy because the behavior changed so substantially in a near-silent update.

PR #406, opened 2020-03-11
Patch pytest.ini files present in ROS packages to include xunit2 format

j-rivero:
The [change to xunit 2.x](https://github.com/ros2/ci/pull/393) Jenkins plugin is not working in situations where ROS packages ship their own pytest.ini file (i.e. ament_package) since it takes precedence over the `pytest.ini` created in the scripts.

I've tested the changes in [this build](https://ci.ros2.org/job/ci_linux/9595/console). Note that failures are coming from cppcheck (which is a different bug) not about pytest.xml.

Alternatively (or in addition) another way of fixing Jenkins would be to include the `junit_family` directive in the pytest.ini file of each ROS package that has this file in it: ament_package, ament_pyflakes, launch_testing_ros,  launch_testing.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -381,7 +381,7 @@ def build_and_test(args, job):
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
     # check if packages have a pytest.ini file and add the xunit2
-    # format2 if not present
+    # format if it is not present
     from pathlib import Path
     for path in Path('.').rglob('pytest.ini'):
         with open(path, "r+") as pytest:
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -381,6 +381,9 @@ def build_and_test(args, job):
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
 
+    with open('build/pytest.ini', 'w') as ini_file:
+        ini_file.write('[pytest]\njunit_family=xunit2')
+
     test_cmd = [
         args.colcon_script, 'test',
         '--base-paths', '"%s"' % args.sourcespace,
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import argparse
+import configparser
 import os
+from pathlib import Path
 import platform
 from shutil import which
 import subprocess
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -380,6 +380,14 @@ def build_and_test(args, job):
     # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
+    # check if packages have a pytest.ini file and add the xunit2
+    # format if it is not present
+    from pathlib import Path
+    for path in Path('.').rglob('pytest.ini'):
+        with open(path, "r+") as pytest:
+            if 'xunit2' not in pytest.read():
```
dirk-thomas: This check if fragile. The string could be only in a comment.

ros2_batch_job/__main__.py
```diff
@@ -380,6 +380,14 @@ def build_and_test(args, job):
...
+                print("xunit2 patch applied to " + str(path.resolve()))
+                pytest.write('\njunit_family=xunit2')
```
dirk-thomas: This won't work if the last group the existing file is not `[pytest]`.

ros2_batch_job/__main__.py
j-rivero: changed it in a6ce403 to use `ConfigParser`

ros2_batch_job/__main__.py
j-rivero: changed it in a6ce403 to use ConfigParser to write the ini file

j-rivero:
I changed the manipulation of the `pytest.ini` file to use `ConfigParser`. Tested in this build https://ci.ros2.org/job/ci_linux-aarch64/5297

ros2_batch_job/__main__.py
```diff
@@ -380,6 +380,23 @@ def build_and_test(args, job):
     # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
+    # check if packages have a pytest.ini file and add the xunit2
+    # format if it is not present
+    from pathlib import Path
+    from configparser import ConfigParser, NoOptionError
```
dirk-thomas: Nitpick: move imports to the top of the file, one import per line (I suggest to `import configparser` and use its classes fully qualified).

ros2_batch_job/__main__.py
```diff
@@ -380,6 +380,23 @@ def build_and_test(args, job):
...
+    for path in Path('.').rglob('pytest.ini'):
+        config = ConfigParser()
+        config.read(str(path.resolve()))
```
dirk-thomas: Why need to call `.resolve()`? A relative path should work just fine.

ros2_batch_job/__main__.py
```diff
@@ -380,6 +380,23 @@ def build_and_test(args, job):
...
+        try:
+            # only if xunit2 is set continue the loop with the file unpatched
+            if config.get('pytest', 'junit_family') == 'xunit2':
+                continue
+        except NoOptionError:
+            pass
+        print("xunit2 patch applied to " + str(path.resolve()))
```
dirk-thomas: Nitpick: single quotes.

Again, why `.resolve()`? A relative path should be much more readable.

ros2_batch_job/__main__.py
j-rivero: ready 5e1884c

ros2_batch_job/__main__.py
j-rivero: you are right, changed 5e1884c

ros2_batch_job/__main__.py
j-rivero: 5e1884c

j-rivero:
new changes tested here [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5300/)

PR #410, opened 2020-03-17
Pin BuildTools version to 16.4.5

brawner:
There is a current Connext issue that surfaced with MS BuildTools 16.5. I'm currently investigating it, but pinning BuildTools to 16.4.5 seems to resolve the issue.

Example failure
[Build Status](https://ci.ros2.org/job/ci_windows/9565/)
Success
[Build Status](https://ci.ros2.org/job/ci_windows/9566/)


dirk-thomas:
Please add a comment above that line to mention what that URL actually fetches.

brawner:
Looks like the test passed. I opened up an issue to track this
https://github.com/ros2/ci/issues/411

PR #414, opened 2020-03-19
Default to Ubuntu Focal for Linux jobs

jacobperron:
Along with Ubuntu Focal, we should use Noetic as the ROS 1 distribution.

Scanning the repo, these seem like the only places that need to be changed; let me know if I missed something. Also, let me know if there's any action that needs to be taken after this PR is merged for changes to take affect.


jacobperron:
Focal CI:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/9705/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5363/)

dirk-thomas:
This change broke the packaging jobs: [Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/1795/)

dirk-thomas:
Another regression: ros/pluginlib#180.

dirk-thomas:
Also all nightly jobs on `armhf` failed.

PR #415, opened 2020-03-20
Force xunit2 for pytest 6.0 when legacy mode is enabled

j-rivero:
following https://github.com/ros2/ci/pull/406#pullrequestreview-372986480
pytest 6.x enables by default the xunit2 format (the one needed by ci.ros2.org Jenkins). With this version we won't need to force the pytest.ini patching in ROS packages. However, upstream could declare the use of the `legacy` mode not to use xunit2 format. This PR detects these cases and patch them accordingly if `legacy` is being used.

**Tested:** [Build Status](https://ci.ros2.org/job/ci_linux/9707/) 
(Tests errors seems unrelated to me, xunit2 patching is done right as far as I can tell).

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -335,10 +335,6 @@ def process_coverage(args, job):
     return 0
 
 
-def check_xunit2_junit_family_value(config, value):
-    return config.get('pytest', 'junit_family', fallback='') == value
-
-
 def build_and_test(args, job):
     compile_with_clang = args.compile_with_clang and args.os == 'linux'
 
@@ -448,6 +444,10 @@ def build_and_test(args, job):
     # Uncomment this line to failing tests a failrue of this command.
     # return 0 if ret_test == 0 and ret_testr == 0 else 1
     return 0
+
+
+def check_xunit2_junit_family_value(config, value):
+    return config.get('pytest', 'junit_family', fallback='') == value
 
 
 def run(args, build_function, blacklisted_package_names=None):
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -335,23 +335,28 @@ def process_coverage(args, job):
     return 0
 
 
+def check_xunit2_junit_family_value(config, value):
+    try:
+        if config.get('pytest', 'junit_family') == value:
+            return True
+    except configparser.NoOptionError:
+        return False
+    else:
+        return False
+
+
 def force_xunit2_in_pytest_ini_files():
+    xunit_6_or_greater = StrictVersion(pytest.__version__) >= StrictVersion('6.0.0')
     for path in Path('.').rglob('pytest.ini'):
         config = configparser.ConfigParser()
         config.read(str(path))
-        if StrictVersion(pytest.__version__) < StrictVersion('6.0.0'):
-            try:
-                # only if xunit2 is set continue the loop with the file unpatched
-                if config.get('pytest', 'junit_family') == 'xunit2':
-                    continue
-            except configparser.NoOptionError:
-                pass
+        if xunit_6_or_greater:
+            # only need to correct explicit legacy option if exists
+            if not check_xunit2_junit_family_value(config, 'legacy'):
+                continue
         else:
-            try:
-                # only need to correct explicit legacy existing option
-                if config.get('pytest', 'junit_family') != 'legacy':
-                    continue
-            except configparser.NoOptionError:
+            # in xunit < 6 need to enforce xunit2 if not set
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -333,6 +335,30 @@ def process_coverage(args, job):
...
+                    continue
+            except configparser.NoOptionError:
+                pass
+        else:
+            try:
+                # only need to correct explicit legacy existing option
+                if config.get('pytest', 'junit_family') != 'legacy':
+                    continue
+            except configparser.NoOptionError:
+                continue
```
dirk-thomas: I am wondering if the duplicate logic in the `if` and `else` block can't be avoided...

ros2_batch_job/__main__.py
```diff
@@ -333,6 +335,30 @@ def process_coverage(args, job):
     return 0
 
 
+def force_xunit2_in_pytest_ini_files():
+    for path in Path('.').rglob('pytest.ini'):
+        config = configparser.ConfigParser()
+        config.read(str(path))
+        if StrictVersion(pytest.__version__) < StrictVersion('6.0.0'):
```
dirk-thomas: This condition should probably only be evaluated once rather than repeatedly during the loop.

dirk-thomas:
@j-rivero Friendly ping.

dirk-thomas:
@j-rivero Friendly ping.

dirk-thomas:
@j-rivero Friendly ping.

ros2_batch_job/__main__.py
j-rivero: the code was awful, yes. I made some changes in 057c1f7 . Hopefully it can be read more easily.

ros2_batch_job/__main__.py
j-rivero: should be fixed by 057c1f7

j-rivero:
Another round of testing for package up to rclcpp [Build Status](https://ci.ros2.org/job/ci_linux/10522/)

ros2_batch_job/__main__.py
```diff
@@ -333,6 +335,35 @@ def process_coverage(args, job):
...
+        config.read(str(path))
+        if xunit_6_or_greater:
+            # only need to correct explicit legacy option if exists
+            if not check_xunit2_junit_family_value(config, 'legacy'):
+                continue
+        else:
+            # in xunit < 6 need to enforce xunit2 if not set
+            if check_xunit2_junit_family_value(config, 'xunit2'):
+                continue
+        print('xunit2 patch applied to ' + str(path))
```
dirk-thomas, line 361: Please update the message to be more specific and quote the path, like:

    print("Patch '%s' to override 'pytest.junit_family=xunit2'" % path)

ros2_batch_job/__main__.py
```diff
@@ -382,21 +413,9 @@ def build_and_test(args, job):
...
-                continue
-        except configparser.NoOptionError:
-            pass
-        print('xunit2 patch applied to ' + str(path))
-        config.set('pytest', 'junit_family', 'xunit2')
-        with open(path, 'w+') as configfile:
-            config.write(configfile)
+    # check if packages have a pytest.ini file that will override xunit2 setup
+    # and patch configuration if needed to keep the xunit2 configuration
+    force_xunit2_in_pytest_ini_files()
```
dirk-thomas, line 418: Why move the code into a function when it is only used in this single location? It makes the diff much more difficult to read.

ros2_batch_job/__main__.py
j-rivero: ack, d0cf651

ros2_batch_job/__main__.py
j-rivero: It tried to reduce the number of code lines in the main code path. Changed in 4e70a79

ros2_batch_job/__main__.py
```diff
@@ -333,6 +335,35 @@ def process_coverage(args, job):
...
 
 
+def check_xunit2_junit_family_value(config, value):
+    try:
+        if config.get('pytest', 'junit_family') == value:
+            return True
+    except configparser.NoOptionError:
+        return False
+    else:
+        return False
```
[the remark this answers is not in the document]
j-rivero: much better, bce4422

j-rivero:
New round of CI [Build Status](https://ci.ros2.org/job/ci_linux/11029/)

dirk-thomas:
(I just made some minor non-functional edits.)

j-rivero:
Final test run https://ci.ros2.org/job/ci_linux/11030/console

PR #416, opened 2020-03-20
Default to error on pytest warning

jacobperron:
By passing the colcon test argument '--pytest-args -W error', any Python warnings produced by pytests will result in a test failure.
Previously, warnings were silently ignored.

Example of build with the newly proposed test argument, testing `ros2node`, which has a known Python warning: [Build Status](https://ci.ros2.org/job/ci_linux/9730/)


jacobperron:
I don't think this is the right solution, since it causes builds to be red on warnings.

PR #419, opened 2020-03-21
remove explicit Python 2 rospkg installation

dirk-thomas:
Follow up of #414.


linux_docker_resources/Dockerfile
```diff
@@ -134,7 +134,6 @@ RUN if test \( ${BRIDGE} = true -o ${INSTALL_TURTLEBOT2_DEMO_DEPS} = true \) ; t
 
 # Install build and test dependencies of ros1_bridge.
 RUN if test ${BRIDGE} = true; then apt-get update && apt-get install --no-install-recommends -y \
-    python-rospkg \
```
dirk-thomas: No need to state this package explicitly since `roslaunch` is pulling it in and will get the correct Python version of it: https://github.com/ros/ros_comm/blob/ac9f88c59a676ca6895e13445fc7d71f398ebe1f/tools/roslaunch/package.xml#L31-L32.

PR #421, opened 2020-03-24
non-EOL openssl on windows

mikaelarguedas:
OpenSSL 1.0.2 is now EOL https://www.openssl.org/policies/releasestrat.html

1.1.1 is the new LTS, also used on macOS and Ubuntu Focal.

This will likely fail CI due to https://github.com/eProsima/Fast-RTPS/issues/1087.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -546,7 +546,7 @@ def run(args, build_function, blacklisted_package_names=None):
             if args.cmake_build_type == 'Debug':
                 pip_packages += [
                     'https://github.com/ros2/ros2/releases/download/cryptography-archives/cffi-1.12.3-cp37-cp37dm-win_amd64.whl',  # required by cryptography
-                    'https://github.com/ros2/ros2/releases/download/cryptography-archives/cryptography-2.7-cp37-cp37dm-win_amd64.whl',
+                    'https://github.com/ros2/ros2/releases/download/cryptography-archives/cryptography-2.7-cp37-cp37dm-win_amd64-openssl111.whl',
                     'https://github.com/ros2/ros2/releases/download/lxml-archives/lxml-4.3.2-cp37-cp37dm-win_amd64.whl',
                     'https://github.com/ros2/ros2/releases/download/netifaces-archives/netifaces-0.10.9-cp37-cp37dm-win_amd64.whl',
                     'https://github.com/ros2/ros2/releases/download/numpy-archives/numpy-1.16.2-cp37-cp37dm-win_amd64.whl',
--- a/windows_docker_resources/Dockerfile.foxy
+++ b/windows_docker_resources/Dockerfile.foxy
@@ -22,8 +22,6 @@ FROM mcr.microsoft.com/windows:$WINDOWS_RELEASE_VERSION
 # Regularly updated installers are next to their associated code.
 ADD https://github.com/ADLINK-IST/opensplice/releases/download/OSPL_V6_9_190925OSS_RELEASE/PXXX-VortexOpenSplice-6.9.190925OSS-HDE-x86_64.win-vs2019-installer.zip C:\TEMP\OpenSplice.zip
 
-# OpenSSL
-ADD https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe C:\TEMP\Win64OpenSSL.exe
 
 # OpenCV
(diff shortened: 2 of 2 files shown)


brawner:
1.1.1 is also listed as experimental on the hosting page. Maybe use 1.1.0?

https://slproweb.com/download/Win64OpenSSL-1_1_0L.exe

mikaelarguedas:
1.1.0 is EOL as well

I believe the "experimental" refers to the MSI installer, not the version of openSSL

mikaelarguedas:
@brawner I ran a job with this branch's [first commit](https://github.com/ros2/ci/pull/421/commits/57517d5b42507f42e834b691a406d335612feec9) trying to use openssl 1.1.1e but apparently it was already replaced by 1.1.1f.
The surprising part is that even if the docker build failed, [the job](https://ci.ros2.org/job/ci_windows/9770) continued without complaining and used an older image.
I would have expected the job to fail if the docker build fails.

Is there a reason not to fail on docker build failure?

nuclearsandwich:
One thing that strikes me about this PR is that it not only changes the OpenSSL version but also the package source. Do you have more info about the OpenSSL chocolatey package. Is that going to track the LTS release or the standard release channel?

mikaelarguedas:
Yeah this is an experiment(still in progress), as the URL broke in the middle of testing this PR it made me think it was worth trying to make this less fragile and less cached by installing via choco.
I have not been able to test the switch to choco yet because the docker build was silently failing for other reasons so the jobs ran without this change. 

Knowing chocolatey, it will most likely track the latest stable (similarly to homebrew). So when a non-LTS version is released it will likely bump to that.
It may also be that chocolatey doesnt install the Debug libraries and would need to be revert back to hard-coded URLs.

This will have two advantages:
- catching breakage earlier, (right now in sros2 we're just finding out now that some implementations don't support OpenSSL LTS because it was not tested/reported before). Catching this breakage will also allow to automatically find out which versions to pin in the installation instructions instead of waiting for user breakage report.
- avoid silent breakage when the installer URL is deleted in favor a more recent version

But this needs further testing in both release and debug mode to make sure it can be used.

sloretz:
Testing this [with 1 minor tweak to the path](https://github.com/ros2/ci/commit/f80951b483890ec6d593da0f9bb5879cfdf27698) and with https://github.com/osrf/homebrew-ros2/pull/8 which has been installed onto mini1. Just `test_security` since those are the tests that have been failing.

 [Build Status](https://ci.ros2.org/job/test_ci_osx/315/)

mikaelarguedas:
This branch worked successfully with Fast-RTPS in Release mode: https://github.com/ros2/system_tests/pull/415#issuecomment-623931513 and failed in Debug mode. Potentially due to outdated wheels.

This has not worked for Connext. It appears that fixing up the name of the `RTI_OPENSSL_LIBS` is necessary for the `PATH` to be modified at runtime for the security tests but is not sufficient for a successful loading of the security plugins: https://ci.ros2.org/job/ci_windows/10424/.
The fact that the paths passed via environment variable hard-code "release" is also a point of concern for debug mode in the future:
```
ENV RTI_OPENSSL_BIN C:\connext\openssl-1.0.2n\x64Win64VS2017\release\bin
ENV RTI_OPENSSL_LIBS C:\connext\openssl-1.0.2n\x64Win64VS2017\release\lib
```

|          |              FastRTPS             |        Connext           |
|:--------:|:---------------------------------:|:------------------------:|
| Release | :heavy_check_mark: | :x: |
| Debug | :x: | :x: |

jacobperron:
@mikaelarguedas I think this ticket can be closed since https://github.com/ros2/ci/pull/454/ has been merged. Feel free to comment and reopen if I'm mistaken.

mikaelarguedas:
@jacobperron The tests are still failing / disabled on WIndows. Do you have another place to track it ?
Did you give a shot at https://github.com/ros2/ci/pull/454#issuecomment-635807294 ?

jacobperron:
Good call. I've opened https://github.com/ros2/ci/issues/488 to track the failures.

> Did you give a shot at #454 (comment) ?

I did not. The referenced branch doesn't seem to exist anymore. Can you provide an updated reference or PR?

mikaelarguedas:
> Can you provide an updated reference or PR?

Took me a bit to get back to it but see https://github.com/ros2/ci/pull/490 and https://github.com/ros2/system_tests/pull/439

PR #432, opened 2020-04-14
Get Python 3.8.2 in Windows Dockerfile for Foxy

jacobperron:
This is an attempt to switch to Python 3.8.2 for our Windows CI.
This is the version of Python we propose for Foxy https://github.com/ros-infrastructure/rep/pull/217

This is a draft PR for two reasons:

1. It doesn't actually work. There errors coming from `rclpy` of the nature:

        E   ImportError: DLL load failed while importing _rclpy: The specified module could not be found.
   
   It's not clear to me what DLL's are failing to load. Here's an example build: https://ci.ros2.org/job/ci_windows/10081

2. Should we have an option in our CI config to continue building with Python 3.7.6?
    I think the answer is "yes", so that we can continue building for Dashing and Eloquent.
    I haven't looked closely, but I think it should be easy to pass in an ARG to the Dockerfile for the Python version.

Feedback on one or both of the above points is appreciated.

### DIFF
--- a/windows_docker_resources/Dockerfile.foxy
+++ b/windows_docker_resources/Dockerfile.foxy
@@ -28,8 +28,8 @@ ADD https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe C:\TEMP\Win64OpenSSL.e
 # OpenCV
 ADD https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip C:\TEMP\opencv.zip
 
-# Python 3.7
-ADD https://www.python.org/ftp/python/3.7.6/python-3.7.6-amd64.exe C:\TEMP\python-37.exe
+# Python 3.8
+ADD https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe C:\TEMP\python-38.exe
 
 # Custom choco packages
 ADD https://github.com/ros2/choco-packages/releases/download/2020-02-24/asio.1.12.1.nupkg C:\TEMP\asio.nupkg
@@ -48,8 +48,8 @@ ADD https://www.zlatkovic.com/pub/libxml/64bit/iconv-1.14-win32-x86_64.7z C:\TEM
 # Restore the default Windows shell for correct batch processing.
 SHELL ["cmd", "/S", "/C"]
 
-RUN C:\TEMP\python-37.exe /quiet `
-  TargetDir=C:\Python37 `
+RUN C:\TEMP\python-38.exe /quiet `
+  TargetDir=C:\Python38 `
   PrependPath=1 `
   Include_debug=1 `
   Include_symbols=1


brawner:
I don't have immediate input on issue 1. You're not seeing that error in focal?

For part 2, it might be better to support separate dockerfiles for different releases. We started doing that with a VS2017/2019 dockerfile split, but after crystal went eol, we dropped 2017. Perhaps instead we have dashing/eloquent/foxy splits of the dockerfiles?

jacobperron:
> You're not seeing that error in focal?

Our nightlies are using Focal and we don't see the error.

> Perhaps instead we have dashing/eloquent/foxy splits of the dockerfiles?

Sounds fine to me.  I guess there may be significant overlap, but it's probably the easiest thing to do.

nuclearsandwich:
I have rebased this PR onto master which has now incorporated #450 and #451 so that this change now targets the Foxy-specific Windows Dockerfile.

nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_windows/10626/)

Trying a build just up to rclpy to see what the damage is there.

jacobperron:
So, it turns out in Python 3.8 DLLs are loaded differently.

> DLL dependencies for extension modules and DLLs loaded with ctypes on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with add_dll_directory() are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for add_dll_directory() and if it exists, use it to add your DLLs directory while loading your library

https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

We're appending package bin directories (containing DLLs) to `PATH`, but Python no longer uses `PATH` for DLL resolution.

After messing around with the new [`os.add_dll_directory`](https://docs.python.org/3/library/os.html#os.add_dll_directory) API, I was able to call `rclpy.init` successfully. I had to call `add_dll_directory` for every recursive DLL dependency of `_rclpy.pyd`.

Here's some example code that uses the new `add_dll_directory()` function, which seems to work:

```python
import os

import rclpy

dll_dir_handles = []

# We could probably just use PATH here
ament_prefix_path = os.environ['AMENT_PREFIX_PATH'].split(';')
for prefix_path in ament_prefix_path:
    # Assuming all DLLs are in 'bin'
    prefix_path_bin = os.path.join(prefix_path, 'bin')
    if os.path.exists(prefix_path_bin):
        # New in Python 3.8: 'add_dll_directory()'
        dll_dir_handles.append(os.add_dll_directory(prefix_path_bin))

rclpy.init()

for handle in dll_dir_handles:
    handle.close()
```

I think the next step is to figure out a good way to integrate the call to `add_dll_library()` in `rclpy`.

jacobperron:
Testing `rclpy` and `demo_nodes_py` with https://github.com/ros2/rclpy/pull/558: [Build Status](https://ci.ros2.org/job/ci_windows/10651/)

jacobperron:
Looks like the rclpy patch didn't fix things; I'll try out some more things locally.

jacobperron:
I also had to patch rosidl_generator_py (https://github.com/ros2/rosidl_python/pull/113), the build passes (`rclpy` and `demo_nods_py`) with Python 3.8: [Build Status](https://ci.ros2.org/job/ci_windows/10656/)

jacobperron:
Full CI with all connected PRs: [Build Status](https://ci.ros2.org/job/ci_windows/10662/)

dirk-thomas:
Instead of all PRs implementing the same non-trivial logic why not provide a context manager implementing the logic which is then used in all these places?

nuclearsandwich:
> Instead of all PRs implementing the same non-trivial logic why not provide a context manager implementing the logic which is then used in all these places?

And thus begins the creation tale of rpyutils. :wink:

jacobperron:
Since the logic relies on the directories being retrieved from PATH and [PATH is set by the template provided by `ament_package`](https://github.com/ament/ament_package/blob/82a6d0ef05818268f06eb83866a76da26fff1b2e/ament_package/template/environment_hook/path.bat#L4), a common implementation could live in `ament_package`. Thoughts?

jacobperron:
> And thus begins the creation tale of rpyutils.

voila https://github.com/ros2/rpyutils/pull/2

jacobperron:
Full CI will all connected PRs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10797/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6182/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8790/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10686/)

PRs:

- https://github.com/ros2/ros2/pull/920
- https://github.com/ros2/rpyutils/pull/2
- https://github.com/ros2/rclpy/pull/558
- https://github.com/ros2/rosidl_python/pull/113
- https://github.com/ros2/geometry2/pull/266
- https://github.com/ros2/rosbag2/pull/411

windows_docker_resources/Dockerfile.foxy
```diff
@@ -28,8 +28,8 @@ ADD https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe C:\TEMP\Win64OpenSSL.e
 # OpenCV
 ADD https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip C:\TEMP\opencv.zip
 
-# Python 3.7
-ADD https://www.python.org/ftp/python/3.7.6/python-3.7.6-amd64.exe C:\TEMP\python-37.exe
+# Python 3.8
+ADD https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe C:\TEMP\python-38.exe
```
nuclearsandwich, line 32: 3.8.3 is out now. Do you feel lucky?

jacobperron:
I think the test failures and warnings are already occurring in the latest nightlies. I'm going to merge this and the connected PRs now and open up a second PR to tryout Python 3.8.3.

windows_docker_resources/Dockerfile.foxy
jacobperron: :game_die: https://github.com/ros2/ci/pull/452

PR #434, opened 2020-04-16
skip tests with a marker / label 'xfail', add nightlies for xfail-only

dirk-thomas:
Related to ros2/ros2#900.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -376,7 +376,7 @@ def main(argv=None):
             'cmake_build_type': 'None',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
             'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
-            'test_args_default': data['test_args_default'].replace('--ctest-args -LE xfail', '--ctest-args -L xfail').replace("--pytest-args -m 'not xfail'", '--pytest-args -m xfail')
+            'test_args_default': data['test_args_default'].replace('--ctest-args -LE xfail', '--ctest-args -L xfail').replace("--pytest-args -m 'not xfail'", '--pytest-args -m xfail --runxfail')
         })
 
         # configure turtlebot jobs on Linux only for now
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -84,7 +84,7 @@ def main(argv=None):
         'use_isolated_default': 'true',
         'colcon_mixin_url': 'https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml',
         'build_args_default': '--event-handlers console_cohesion+ console_package_list+ --cmake-args -DINSTALL_EXAMPLES=OFF -DSECURITY=ON',
-        'test_args_default': '--event-handlers console_direct+ --executor sequential --retest-until-pass 2',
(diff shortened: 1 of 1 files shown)


create_jenkins_job.py
```diff
@@ -279,7 +277,7 @@ def create_job(os_name, job_name, template_file, additional_dict):
                 # We do not want to test more than this as we observe issues with the clang libcxx standard library
                 # we don't plan to tackle for now. The important part of this nightly is to make sure the code compiles
                 # without emitting thread-safety related warnings.
-                'test_args_default': '--event-handlers console_direct+ --executor sequential --packages-select rcutils',
+                'test_args_default': data['test_args_default'].replace(' --retest-until-pass 2', '') + ' --packages-select rcutils'
```
nuclearsandwich, line 280: Is omitting the retest arg related to the rest of the PR somehow?

create_jenkins_job.py
```diff
@@ -363,7 +359,7 @@ def create_job(os_name, job_name, template_file, additional_dict):
         if os_name == 'windows':
             job_name = job_name[:15]
         test_args_default = os_configs.get(os_name, data).get('test_args_default', data['test_args_default'])
-        test_args_default = test_args_default.replace('--retest-until-pass', '--retest-until-fail') + " --ctest-args -LE linter --pytest-args -m 'not linter'"
+        test_args_default = test_args_default.replace('--retest-until-pass', '--retest-until-fail').replace('--ctest-args -LE xfail', "--ctest-args -LE '(linter|xfail)'").replace("--pytest-args -m 'not xfail'", "--pytest-args -m 'not linter and not xfail'")
```
nuclearsandwich, line 362: Although there's no code style guidelines the replacements here might be worth splitting into multiple lines for human readability.

create_jenkins_job.py
nuclearsandwich: Oh I see. Because we're reusing the defaults we have to tweak them to match the previous test args. Rescinded.

dirk-thomas: The patch doesn't change the actual behavior (beside skipping `xfail` of course). Before all args where repeated and the retest arg omitted as described in the comment above. After it uses the previous definition of the args and manipulate it accordingly to avoid duplication.

create_jenkins_job.py
dirk-thomas: Done in fdc53f6b4855d7ba524cd202847f00356bcf12ef.

chapulina:
I marked a test with `xfail` on https://github.com/ros2/rclcpp/pull/1109 but it wasn't skipped. Should this label have an effect on `ci_linux` jobs?

https://ci.ros2.org/job/ci_linux/10657/testReport/

dirk-thomas:
> I marked a test with `xfail` on https://github.com/ros2/rclcpp/pull/1109 but it wasn't skipped. Should this label have an effect on `ci_linux` jobs?

Yes, CI jobs do skip tests marked with that label. You have overwritten the CTest args in the job parameters with something else and with that reenabled the test.

chapulina:
> You have overwritten the CTest args in the job parameters with something else and with that reenabled the test.

Thanks! I copied that from another job, triggered a new one without that :crossed_fingers:

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-05-21/14247/1

PR #441, opened 2020-04-30
add isolated parameter to packaging jobs defaulting to False

dirk-thomas:
This will allow building custom archives with isolation which will help in debugging.

### DIFF
--- a/job_templates/snippet/property_parameter-definition.xml.em
+++ b/job_templates/snippet/property_parameter-definition.xml.em
@@ -10,6 +10,7 @@
     cmake_build_type=cmake_build_type,
     build_args_default=build_args_default,
     os_name=os_name,
+    use_isolated_default=use_isolated_default,
 ))@
         <hudson.model.BooleanParameterDefinition>
           <name>CI_USE_WHITESPACE_IN_PATHS</name>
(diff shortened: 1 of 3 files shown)


job_templates/snippet/property_parameter-definition.xml.em
```diff
@@ -23,11 +24,6 @@ This tests the robustness to whitespace being within the different paths.</descr
     ignore_rmw_default=ignore_rmw_default,
     use_connext_debs_default=use_connext_debs_default,
 ))@
-        <hudson.model.BooleanParameterDefinition>
-          <name>CI_ISOLATED</name>
-          <description>By setting this to True, the build will use the --isolated option.</description>
-          <defaultValue>@(use_isolated_default)</defaultValue>
```
sloretz, line 29: Where did `use_isolated_default` come from before this PR added it to the `property_parameter-definition_common` snippet?

dirk-thomas: The Python script passes it as an argument to the template evaluation and the main template passes it down to the template snippets.

sloretz: Seems like `packaging_job.xml.em` needs the logic to add the `--isolated` flag like the `ci_job.xml.em` does

https://github.com/ros2/ci/blob/9ce633a9411e964e2c54c2940da4e566388efbd0/job_templates/ci_job.xml.em#L163-L165

sloretz: I would have thought `--isolated` was the default and `--merge-install` would have to be given. Are we using a colcon.pkg somewhere that changes that?

sloretz: > I would have thought --isolated was the default and --merge-install would have to be given. Are we using a colcon.pkg somewhere that changes that?

Ohhh I see this is an argument to a different script that defaults to false.

https://github.com/ros2/ci/blob/e667c841f50e32f5273c6adf1d4118ec14e59e33/ros2_batch_job/__main__.py#L200-L202

IIUC since the packaging job didn't pass `--isolated` before, that script defaulted to an isolated build. In that case, it definitely looks like `packaging_job.xml.em` needs the logic to read `$CI_ISOLATED` and set `$CI_ARGS` then.

dirk-thomas: Good catch: d3cc42eaa5a292d0079edbdbe9224e9ad8f6415d.

dirk-thomas:
:ship: deployed.

PR #443, opened 2020-05-01
Install PyQt5 and numpy from PyPI

sloretz:
Needed on OSX machines for rosidl_generator_py https://github.com/ros2/rosidl_python/blob/5aae255ed3cd4e357a62b3e19920a9df4f840f48/rosidl_generator_py/package.xml#L25

PR #436, opened 2020-04-18
Use OpenSSL 1.0.2 for Connext on MacOS

mikaelarguedas:
Without this, Connext cannot use the security plugins: https://github.com/ros2/build_farmer/issues/269

Mini2: [Build Status](http://ci.ros2.org/job/ci_osx/8382/)
Lore: [Build Status](http://ci.ros2.org/job/ci_osx/8383/)

Mini3: [Build Status](http://ci.ros2.org/job/ci_osx/8384/)
Not sure what's going on on mini3, It shows the exact same error as the other ones without this patch applied (e.g. nightly on lore https://ci.ros2.org/view/nightly/job/nightly_osx_release/1620/).
it doesn't have the same OpenSSL version as the other ones (1.0.2s vs 1.0.2t), maybe that's part of the reason ? 

Tested with https://github.com/ros2/ci/pull/421

PR #444, opened 2020-05-01
Platform on OSX may start with macos

sloretz:
Alternative to #443. This is exploring using Python 3.8 as a solution to the problem described here https://github.com/ros2/ros2_documentation/pull/641#issuecomment-622611497

Platform name using python 3.8: `macOS-10.14.6-x86_64-i386-64bit`

[Build Status](https://ci.ros2.org/view/All/job/test_ci_osx/319/)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -459,7 +459,7 @@ def run(args, build_function, blacklisted_package_names=None):
         args.os = 'linux'
         from .linux_batch import LinuxBatchJob
         job = LinuxBatchJob(args)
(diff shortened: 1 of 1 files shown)


sloretz:
CI up to `qt_binding_cpp` looked ok, trying all of CI [Build Status](https://ci.ros2.org/job/test_ci_osx/320/)

sloretz:
CI looks ok, build is red because of test failures, but test failures are because of not having #436

@jacobperron mind reviewing? This is required to use upgrade the OSX machines since our dependencies from homebrew upgraded to require Python 3.8

@chapulina FYI as the current buildfarmer

PR #448, opened 2020-05-07
Remove stray quote from powershell one-liner.

nuclearsandwich:
This didn't show up in tests since the job configuration was deployed by
hand to avoid affecting production builds and the stray quote wasn't
propagated.


nuclearsandwich:
Test CI finished the Docker build and was able to build rcutils. [Build Status](https://ci.ros2.org/job/test_ci_windows/154/)

PR #451, opened 2020-05-13
Split windows dockerfiles for different ROS distros.

nuclearsandwich:
Right now both Dockerfiles are identical but this split will allow us to start using Python 3.8 (#432) and OpenSSL 1.1.1 (#421) in Foxy without dropping support for building Dashing and Eloquent.

Test builds with both Dockerfiles just building rcutils (just validating parameter as the files are currently identical).
Foxy [Build Status](https://ci.ros2.org/job/test_ci_windows/155/)
Dashing [Build Status](https://ci.ros2.org/job/test_ci_windows/156/)


dirk-thomas:
Nitpick: it would have been nice to maintain the history of Dockerfile for Foxy.

PR #453, opened 2020-05-20
Update debug wheels used on Windows.

nuclearsandwich:
This updates the Debug wheels for Python 3.8.

In order to support builds of Dashing and Eloquent we'll need to come
back to this and add support for using a different set of wheels for the
older python version but this hopefully gets us the Windows debug nightlies back.

PR #454, opened 2020-05-20
Update Foxy OpenSSL to 1.1.1g on Windows.

nuclearsandwich:
This is an alternative to #421 which installs (currently the same) OpenSSL binary via chocolatey. Since the OpenSSL version needs to be compiled in to our debug wheel of cryptography for sros2 I would prefer to opt into updates rather than get them by surprise via chocolatey.

This OpenSSL binary distribution also has a different default install
location: C:\Program Files\OpenSSL-Win64 I didn't see a cli option in the installer to change the target location but one may exist.

### DIFF
--- a/windows_docker_resources/Dockerfile.foxy
+++ b/windows_docker_resources/Dockerfile.foxy
@@ -117,7 +117,7 @@ ENV RTI_LICENSE_FILE C:\connext\rti_license.dat
 COPY rticonnextdds-src\openssl-1.0.2n-target-x64Win64VS2017.zip C:\TEMP\connext\
 RUN 7z.exe x C:\TEMP\connext\openssl-1.0.2n-target-x64Win64VS2017.zip -aoa -oC:\connext\
 ENV RTI_OPENSSL_BIN C:\connext\openssl-1.0.2n\x64Win64VS2017\release\bin
-ENV RTI_OPENSSL_LIB C:\connext\openssl-1.0.2n\x64Win64VS2017\release\lib
+ENV RTI_OPENSSL_LIBS C:\connext\openssl-1.0.2n\x64Win64VS2017\release\lib
 
 COPY rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe.??? C:\TEMP\connext\
 RUN copy /b C:\TEMP\connext\rti_connext_dds-5.3.1-pro-host-x64Win64.exe.??? C:\TEMP\connext\rti_connext_dds-5.3.1-pro-host-x64Win64.exe
--- a/windows_docker_resources/Dockerfile.foxy
+++ b/windows_docker_resources/Dockerfile.foxy
@@ -23,7 +23,7 @@ FROM mcr.microsoft.com/windows:$WINDOWS_RELEASE_VERSION
 ADD https://github.com/ADLINK-IST/opensplice/releases/download/OSPL_V6_9_190925OSS_RELEASE/PXXX-VortexOpenSplice-6.9.190925OSS-HDE-x86_64.win-vs2019-installer.zip C:\TEMP\OpenSplice.zip
 
 # OpenSSL
-ADD https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe C:\TEMP\Win64OpenSSL.exe
+ADD https://slproweb.com/download/Win64OpenSSL-1_1_1g.exe C:\TEMP\Win64OpenSSL.exe
 
 # OpenCV
 ADD https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip C:\TEMP\opencv.zip
@@ -78,11 +78,11 @@ RUN 7z.exe x C:\TEMP\opencv.zip -aoa -oC:\
 RUN 7z.exe x C:\TEMP\OpenSplice.zip -aoa -oC:\opensplice
 
 # Environment setup
-ENV OPENSSL_CONF C:\OpenSSL-Win64\bin\openssl.cfg
+ENV OPENSSL_CONF "C:\Program Files\OpenSSL-Win64\bin\openssl.cfg"
 ENV OpenCV_DIR C:\opencv
 ENV OSPL_HOME C:\opensplice\HDE\x86_64.win64
 # You can't use ENV to append to the PATH https://stackoverflow.com/questions/42092932/appending-to-path-in-a-windows-docker-container
-RUN setx PATH "%PATH%;C:\Program Files\Git\cmd;C:\Program Files\CMake\bin;C:\OpenSSL-Win64\bin\;C:\xmllint\bin;"C:\opencv\x64\vc16\bin"
+RUN setx PATH "%PATH%;C:\Program Files\Git\cmd;C:\Program Files\CMake\bin;C:\Program Files\OpenSSL-Win64\bin\;C:\xmllint\bin;"C:\opencv\x64\vc16\bin"
 
 RUN powershell -Command Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1
 RUN mkdir C:\ws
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
CI building up to sros2 and test_security and testing sros2 and test_security.

Windows [Build Status](https://ci.ros2.org/job/ci_windows/10732/)
Windows Debug [Build Status](https://ci.ros2.org/job/ci_windows/10733/) (will hopefully resolve issues introduced by #453 since the cryptography wheel was built against this OpenSSL binary distribution.

nuclearsandwich:
This PR currently exchanges security failures using rmw_connext_cpp in all Windows configurations for working builds with the new Cryptography wheel in Windows Debug.
I'm currently trying to resolve the rmw_connext_cpp issue. I'm not *yet* ready advocate for regressing these tests to fix the build but I am considering it.

nuclearsandwich:
When looking at it, the Windows dockerfiles always had support for pointing connext to its own OpenSSL installation but there was a disagreement between the environment variable name in the initial and final iterations of the PR. [Build Status](https://ci.ros2.org/job/ci_windows/10738/) is running with a corrected environment variable name and hopefully it's :green_circle:.

nuclearsandwich:
Well the latest CI run didn't change the outcome. For some reason the RTI_OPENSSL_LIBS directory is being put on the PATH rather than the RTI_OPENSSL_BIN directory (which on Windows includes both executables and ddls). I'm out of gas for tonight so we'll lose another debug nightly and I'll be back at it in the morning.

windows_docker_resources/Dockerfile.foxy
```diff
@@ -78,11 +78,11 @@ RUN 7z.exe x C:\TEMP\opencv.zip -aoa -oC:\
 RUN 7z.exe x C:\TEMP\OpenSplice.zip -aoa -oC:\opensplice
 
 # Environment setup
-ENV OPENSSL_CONF C:\OpenSSL-Win64\bin\openssl.cfg
+ENV OPENSSL_CONF "C:\Program Files\OpenSSL-Win64\bin\openssl.cfg"
```
jacobperron, line 81: Strange, on my native Windows machine, the OpenSSL installer defaults to ` C:\OpenSSL-Win64` as the install root.

nuclearsandwich: Do you have an earlier version of OpenSSL installed? It might be looking at the previous installation path?

jacobperron: I re-located the old installation of OpenSSL; but it's possible there was still some user-data floating around.

jacobperron:
I'm stuck on this as well. Locally, the Connext security tests pass for me when setting `RTI_OPENSSL_LIBS` and `RTI_OPENSSL_BIN` accordingly. It seems there an issue with our CI configuration.

Based on the errors, it sounds like Connext isn't installed properly:

```
12:41:51 21: [test_publisher-1] >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
12:41:51 21: [test_publisher-1] This error state is being overwritten:
12:41:51 21: [test_publisher-1] 
12:41:51 21: [test_publisher-1]   'failed to create participant, at C:\ci\ws\src\ros2\rmw_connext\rmw_connext_shared_cpp\src\node.cpp:281, at C:\ci\ws\src\ros2\rcl\rcl\src\rcl\node.c:277'
12:41:51 21: [test_publisher-1] 
12:41:51 21: [test_publisher-1] with this new error message:
12:41:51 21: [test_publisher-1] 
12:41:51 21: [test_publisher-1]   'rcl node's rmw handle is invalid, at C:\ci\ws\src\ros2\rcl\rcl\src\rcl\node.c:429'
```

But I ran `rcl` tests with this CI branch and it appears to be working (modulo a test failure): https://ci.ros2.org/job/ci_windows/10788/testReport/

@mikaelarguedas any ideas?

jacobperron:
>  Locally, the Connext security tests pass for me when setting RTI_OPENSSL_LIBS and RTI_OPENSSL_BIN accordingly

It seems that the Connext tests pass locally for me even if `RTI_OPENSSL_LIBS` and `RTI_OPENSSL_LIBS` are unset. 🤔

mikaelarguedas:
> It seems that the Connext tests pass locally for me even if RTI_OPENSSL_LIBS and RTI_OPENSSL_LIBS are unset. thinking

Could you still be using Openssl 1.0.2 like the current CI does so you dont need to point to RTI's OpenSSL?

---

This s a bit unfortunate that this doesnt build on the work of https://github.com/ros2/ci/pull/421 that already has context, a bunch of tickets linked to it and acknowledges the work done on this topic over the last 2 months... several of the questions in this PR are answered on the other PR...

> the Windows dockerfiles always had support for pointing connext to its own OpenSSL installation but there was a disagreement between the environment variable name in the initial and final iterations of the PR

I believe the "always had support" was mostly a copy-n-paste from the Linux dockerfiles but have never been tested, variable names were not matching and variable contents were never taken into account https://github.com/ros2/system_tests/pull/409. So this is not known the have ever worked in the past that I know of.
Necessity of updating env var name and impact on the PATH highlighted in https://github.com/ros2/ci/pull/421#issuecomment-628251401

> Well the latest CI run didn't change the outcome. For some reason the RTI_OPENSSL_LIBS directory is being put on the PATH rather than the RTI_OPENSSL_BIN directory (which on Windows includes both executables and ddls).

I don't think that's true. A good way to confirm it is to see that the first element of the `PATH` of the tests in the [CI job linked above](https://ci.ros2.org/job/ci_windows/10738/) is `C:/connext/openssl-1.0.2n/x64Win64VS2017/release/bin`
However it's never been attempted to use that extracted OpenSSL on Windows and I don't know the content of the folders `C:\connext\openssl-1.0.2n\x64Win64VS2017\release\lib` and `C:\connext\openssl-1.0.2n\x64Win64VS2017\release\bin` which is why I couldn't push https://github.com/ros2/ci/pull/421 further.
2 things that come to mind:
- these paths have VS2017 in it, could it be that we need a version built with a more recent compiler/standard libraries ?
- these paths have "release" hard-coded in them, this looks like it may not work out of the box for the debug case...

alternative idea:
- if OpenSSL 1.1.1 and OpenSSL 1.0.2 install in different locations, could we install them side by side and just change the environment variables to point to the 1.0.2 version ?

jacobperron:
> Could you still be using Openssl 1.0.2 like the current CI does so you dont need to point to RTI's OpenSSL?

I have OpenSSL 1.1.1g installed. And then extracted RTI's OpenSSL into a separate directory. :man_shrugging: 

> these paths have VS2017 in it, could it be that we need a version built with a more recent compiler/standard libraries ?

AFAIK, VS2017 and VS2019 are binary compatible.

> these paths have "release" hard-coded in them, this looks like it may not work out of the box for the debug case...

Maybe not, but I guess we should first figure out why tests are not passing in a release build.

> if OpenSSL 1.1.1 and OpenSSL 1.0.2 install in different locations, could we install them side by side and just change the environment variables to point to the 1.0.2 version ?

I think you're suggesting making this change in test_security, right? We can give it a try.

mikaelarguedas:
> I have OpenSSL 1.1.1g installed. And then extracted RTI's OpenSSL into a separate directory. :man_shrugging: 

That's encouraging ! now just need to play spot the difference with CI

> > if OpenSSL 1.1.1 and OpenSSL 1.0.2 install in different locations, could we install them side by side and just change the environment variables to point to the 1.0.2 version ?
> I think you're suggesting making this change in test_security, right? We can give it a try.

What I meant is to make this dockerfile install both openssl 1.1.1 and openssl 1.0.2 from the website and set the RTI_OPENSSL_* variables to point to the OpenSSL 1.0.2 installation folder (without needing a change in test_security) and see if in that case the Connext security plugins can be loaded successfully

jacobperron:
I've just opened this PR that disables the Connext security tests on Windows: https://github.com/ros2/system_tests/pull/433

I think it would be good to land that and this PR so that we unblock Windows debug builds and at get OpenSSL 1.1.1 for the security tests with other RMWs.

mikaelarguedas:
> I've just opened this PR that disables the Connext security tests on Windows: ros2/system_tests#433

> I think it would be good to land that and this PR so that we unblock Windows debug builds and get OpenSSL 1.1.1 for the security tests with other RMWs.

Sure, that's still a step in the right direction :+1: 

---

I gave a shot to the alternative of installing both openssl 1.0.2 and 1.1.1 from the website (using this branch https://github.com/ros2/ci/compare/recent-openssl) and ran jobs with FastRTPS and Connext in release and debug and they came back green:
Release: [Build Status](https://ci.ros2.org/job/ci_windows/10817/)
Debug: [Build Status](https://ci.ros2.org/job/ci_windows/10818/)


The Fast-RTPS build step is for sure using openssl 1.1.1g, I'm not certain for the tests but it seems likely

PR #457, opened 2020-05-21
Filter build/ and install/ paths from coverage report

j-rivero:
Reduce the number of entries in the coverage report by excluding install/ and build/ workspace directories. I've used the `--exclude` option in gcovr.

Before:
https://ci.ros2.org/job/ci_linux_coverage/78/cobertura/

After:
https://ci.ros2.org/job/ci_linux_coverage/116/cobertura/

I think that ideally the report should only show the packages selected (somehow) by the user when triggering the job. Not sure if that is possible with the set of current Jenkins input parameters that we think on using for running coverage or we would need to add a custom parameter for that. 

//cc @chapulina

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -308,6 +308,8 @@ def process_coverage(args, job):
             '-k',
             '-r', os.path.abspath('.'),
             '--xml', '--output=' + outfile,
+            '-e' + os.path.join(os.path.abspath('.'), args.installspace),
+            '-e' + os.path.join(os.path.abspath('.'), args.buildspace),
             '-g']
         print(cmd)
         subprocess.run(cmd, check=True)


dirk-thomas:
Don't that exclude any kind of generated code?

chapulina:
> Don't that exclude any kind of generated code?

Good point. What do you suggest we do in this case? The goal is to reduce the duplication that's currently there.

j-rivero:
> Good point. What do you suggest we do in this case?

Yep, it is a valid point. Digging it a bit deeper, one example:

* `class loader` There are 7 headers in the source code: https://github.com/ros/class_loader/tree/ros2/include/class_loader. There are [4 in the install/ workspace](https://ci.ros2.org/job/ci_linux_coverage/78/cobertura/install_class_loader_include_class_loader/) reported in by the coverage report and[ 2 in the src/](https://ci.ros2.org/job/ci_linux_coverage/78/cobertura/src_ros_class_loader_include_class_loader/) also present in `install/` but with different levels of coverage. 

My hypothesis is: generated code goes into `install/` and `build/`. The `build/` rates come from the tests present in the same package (unit tests mainly). The `install/` rates (which are different) and come from tests in other packages that use the headers or code in it (integration/system tests).

Since file paths are not the same, try to combine files can be tricky but feasible.

j-rivero:
follow in #464

PR #461, opened 2020-05-28
A solution of Suppress progress bar in CI

### LINKED ISSUE #431: Suppress progress bar in CI
In CI, pip shows the progress bar https://ci.ros2.org/job/ci_linux/lastStableBuild/consoleText

This creates log noise and should probably be disabled (maybe with `--progress-bar off` or `--quiet`) This can be set with pip config https://pip.pypa.io/en/stable/reference/pip_config/

```
  Downloading mypy-0.770-cp38-cp38-manylinux1_x86_64.whl (22.6 MB)
[?25l
[K     |                                | 10 kB 40.1 MB/s eta 0:00:01
[K     |                                | 20 kB 45.6 MB/s eta 0:00:01
[K     |                                | 30 kB 55.6 MB/s eta 0:00:01
[K     |                                | 40 kB 27.8 MB/s eta 0:00:01
[K     |                                | 51 kB 31.9 MB/s eta 0:00:01
[K     |                                | 61 kB 36.1 MB/s eta 0:00:01
[K     |                                | 71 kB 39.1 MB/s eta 0:00:01
[K     |▏                               | 81 kB 41.0 MB/s eta 0:00:01
[K     |▏                               | 92 kB 43.5 MB/s eta 0:00:01
[K     |▏                               | 102 kB 36.1 MB/s eta 0:00:01
[K     |▏                               | 112 kB 36.1 MB/s eta 0:00:01
[K     |▏                               | 122 kB 36.1 MB/s eta 0:00:01
[K     |▏                               | 133 kB 36.1 MB/s eta 0:00:01
[K     |▏                               | 143 kB 36.1 MB/s eta 0:00:01
[K     |▏                               | 153 kB 36.1 MB/s eta 0:00:01
[K     |▎                               | 163 kB 36.1 MB/s eta 0:00:01
```

sunbo57123:
Context: https://github.com/ros2/ci/issues/431 
add `--process-bar off` to pip command can really disable process bars.
Result in my Jenkins is
job:ci_linux 
env: Ubuntu 18 / amd64
Fixes #431
image

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -569,7 +569,7 @@ def run(args, build_function, blacklisted_package_names=None):
             job.run(
                 ['"%s"' % job.python, '-m', 'pip', 'uninstall', '-y'] +
                 ['cryptography', 'lxml', 'numpy'], shell=True)
-        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U', '--progress-bar off']
+        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U']
         if args.do_venv or sys.platform == 'win32':
             # Force reinstall so all dependencies are in virtual environment
             # On Windows since we switch between the debug and non-debug
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -569,7 +569,7 @@ def run(args, build_function, blacklisted_package_names=None):
             job.run(
                 ['"%s"' % job.python, '-m', 'pip', 'uninstall', '-y'] +
                 ['cryptography', 'lxml', 'numpy'], shell=True)
-        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U']
+        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U', '--progress-bar off']
         if args.do_venv or sys.platform == 'win32':
             # Force reinstall so all dependencies are in virtual environment
             # On Windows since we switch between the debug and non-debug
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -185,6 +185,7 @@ class MyProtocol(AsyncSubprocessProtocol):
             self.progress.append(data)
             if len(self.progress) == 10:
                 sys.stdout.write(self.progress[-1].decode('utf-8', 'replace').replace(os.linesep, '\n'))
+                self.progress = []
         else:
             sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
 
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -197,11 +197,14 @@ class PipProtocol(AsyncSubprocessProtocol):
         # '[?25l' and the end sequence of '[?25h'.
         if b'[?25l' in data:
             self.progress_bar = True
-        if b'[?25h' in data:
-            for display_line in self.progress_data.split("\r"):
+            sys.stdout.write(data.decode('utf-8', 'replace').replace('\r', '')+'\n')
+        elif b'[?25h' in data:
+            for display_line in self.progress_data.replace("\n","").split("\r"):
                 sys.stdout.write(display_line+'\n' if display_line else '')
             self.progress_bar = False
-        if self.progress_bar:
+            self.progress_data = ''
+            sys.stdout.write(data.decode('utf-8', 'replace').replace('\r', ''))
+        elif self.progress_bar:
             self.progress_data += data.decode('utf-8', 'replace')
             if len(self.progress_data) >= 1000:
                 sys.stdout.write(self.progress_data.split("\r")[-1]+'\n')
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -188,7 +188,7 @@ class PipProtocol(AsyncSubprocessProtocol):
         self.cmd = cmd
         self.exit_on_error = exit_on_error
         self.progress_bar = False
-        self.progress_data = []
+        self.progress_data = ''
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
@@ -198,14 +198,14 @@ class PipProtocol(AsyncSubprocessProtocol):
         if b'[?25l' in data:
             self.progress_bar = True
         if b'[?25h' in data:
-            for display_bar in self.progress_data:
-                sys.stdout.write(display_bar.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+            for display_line in self.progress_data.split("\r"):
+                sys.stdout.write(display_line+'\n' if display_line else '')
             self.progress_bar = False
         if self.progress_bar:
-            self.progress_data.append(data)
-            if len(self.progress_data) == 10:
-                sys.stdout.write(self.progress_data[-1].decode('utf-8', 'replace').replace(os.linesep, '\n'))
-                self.progress_data = []
+            self.progress_data += data.decode('utf-8', 'replace')
+            if len(self.progress_data) >= 1000:
+                sys.stdout.write(self.progress_data.split("\r")[-1]+'\n')
+                self.progress_data = ''
         else:
             sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
 
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -215,7 +215,10 @@ def run_with_prefix(prefix, cmd, exit_on_error=True, **kwargs):
 
 def _run(cmd, exit_on_error=True, **kwargs):
     def create_protocol(*args, **kwargs):
-        return MyProtocol(cmd, exit_on_error, *args, **kwargs)
+        if "pip" in cmd and "install" in cmd:
+            return PipProtocol(cmd, exit_on_error, *args, **kwargs)
+        else:
+            return MyProtocol(cmd, exit_on_error, *args, **kwargs)
 
     @asyncio.coroutine
     def run_coroutine(future):
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -167,25 +167,25 @@ def warn(*args, **kwargs):
 
 class MyProtocol(AsyncSubprocessProtocol):
     def __init__(self, cmd, exit_on_error, *args, **kwargs):
-        self.cmd = cmd  
+        self.cmd = cmd
         self.exit_on_error = exit_on_error
         self.progress_bar = False
-        self.progress = []
+        self.progress_data = []
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
+        # '[?25l' and '[?25h' can be considered as a sign of begining and end of pip progress bar
         if b'[?25l' in data:
             self.progress_bar = True
         if b'[?25h' in data:
-            if self.progress:
-                for display in self.progress:
-                    sys.stdout.write(display.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+            for display_bar in self.progress_data:
+                sys.stdout.write(display_bar.decode('utf-8', 'replace').replace(os.linesep, '\n'))
             self.progress_bar = False
         if self.progress_bar:
-            self.progress.append(data)
-            if len(self.progress) == 10:
-                sys.stdout.write(self.progress[-1].decode('utf-8', 'replace').replace(os.linesep, '\n'))
-                self.progress = []
+            self.progress_data.append(data)
+            if len(self.progress_data) == 10:
+                sys.stdout.write(self.progress_data[-1].decode('utf-8', 'replace').replace(os.linesep, '\n'))
+                self.progress_data = []
         else:
             sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
 
@@ -197,7 +197,6 @@ class MyProtocol(AsyncSubprocessProtocol):
         if self.exit_on_error and returncode != 0:
             log("@{rf}@!<==@| '{0}' exited with return code '{1}'",
                 fargs=(" ".join(self.cmd), returncode))
-
 
 
 def run(cmd, exit_on_error=True, **kwargs):
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -174,7 +174,9 @@ class MyProtocol(AsyncSubprocessProtocol):
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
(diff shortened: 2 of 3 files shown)


ros2_batch_job/__main__.py
```diff
@@ -569,7 +569,7 @@ def run(args, build_function, blacklisted_package_names=None):
             job.run(
                 ['"%s"' % job.python, '-m', 'pip', 'uninstall', '-y'] +
                 ['cryptography', 'lxml', 'numpy'], shell=True)
-        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U']
+        pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-U', '--progress-bar off']
```
clalancette, line 572: While this does make it quieter, it has two issues:

1.  It doesn't work on Ubuntu 18.04.  While this isn't supported on Foxy, we still need to be able to run this for Eloquent and Dashing, so we need to make this conditional on the Ubuntu version.
1.  We don't want to *totally* disable the progress bar, as that doesn't give us any indication that pip hung on a download.  I might suggest switching this to `ascii` instead (though I don't know how that looks in the Jenkins console output).

sunbo57123: hi,
Thanks for your reply.
```
It doesn't work on Ubuntu 18.04. While this isn't supported on Foxy, we still need to be able to run this for Eloquent and Dashing, so we need to make this conditional on the Ubuntu version.
```
In fact, I am running CI jobs on an AMD server withUbuntu 18, and the configuration comes from the master branch, so there may be some configuration differences between the official server and my server. 
````
We don't want to totally disable the progress bar, as that doesn't give us any indication that pip hung on a download. I might suggest switching this to ascii instead (though I don't know how that looks in the Jenkins console output).
````
I tried to change the stdout of the pip progress bar to ASCII through "--progress-bar ascii", and the log is still very verbose. The change looks like the block has changed to "#", and the number of lines in the progress bar has not decreased. 

The solution may be to change the printing speed, progress bar line would be printed every 0.5 second.

Comment (clalancette) [reply ros2_batch_job/__main__.py:572]: > In fact, I am running CI jobs on an AMD server with Ubuntu 18.04, and the configuration comes from the master branch, so there may be some configuration differences between the official server and my server.

Well, since this is a Dockerfile, by default you'd use the distribution specified at the top.  By default, that is 20.04, but we also have a switch to 18.04.  There, the `--progress-bar` option doesn't exist, and throws an error.

> The solution may be to change the printing speed, progress bar line would be printed every 0.5 second.

Sure, if there is a way to slow down how often it prints, that works too.  I think I would still stick with the ascii output, since there aren't any "special" characters there.

**Comment by sunbo57123** [comment]
Hello, 
I added a filter to the standard output decoding part. The execution part comes from another project: osrf, so I think this part should not be changed.

This solution works in my environment, so if you have time, could you check whether it works on the official environment.

btw, there is a similar problem in this discussion:
https://discuss.circleci.com/t/build-output-is-too-large-due-to-capturing-pip-install-progress/21029/4

File: ros2_batch_job/util.py [hunk 2]
```diff
@@ -167,12 +167,28 @@ def warn(*args, **kwargs):
...
+        self.progress_bar = False
+        self.progress = []
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
-        sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+        if b'[?25l' in data:
+            self.progress_bar = True
+        if b'[?25h' in data:
+            if self.progress:
```
clalancette, line 180: This `if` statement isn't required; the loop below will execute 0 times if there is no data in `self.progress`.

ros2_batch_job/util.py
```diff
@@ -167,12 +167,28 @@ def warn(*args, **kwargs):
 
 class MyProtocol(AsyncSubprocessProtocol):
     def __init__(self, cmd, exit_on_error, *args, **kwargs):
-        self.cmd = cmd
+        self.cmd = cmd  
         self.exit_on_error = exit_on_error
+        self.progress_bar = False
+        self.progress = []
```
clalancette, line 173: I'll suggest renaming this something more like `self.progress_data`; that will make it a little clearer on what it is for.

ros2_batch_job/util.py
```diff
@@ -167,12 +167,28 @@ def warn(*args, **kwargs):
...
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
-        sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+        if b'[?25l' in data:
```
clalancette, line 177: Please put a comment at the top of this logic explaining why we are doing this.  That will help for anybody who comes along later and wonders why this code is here.

ros2_batch_job/util.py
```diff
@@ -167,12 +167,28 @@ def warn(*args, **kwargs):
...
+            self.progress_bar = True
+        if b'[?25h' in data:
+            if self.progress:
+                for display in self.progress:
+                    sys.stdout.write(display.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+            self.progress_bar = False
+        if self.progress_bar:
+            self.progress.append(data)
+            if len(self.progress) == 10:
```
clalancette, line 186: Do you have a screenshot of what this ends up looking like?  Since we are now only printing every 10th piece of output, I feel like this is going to look pretty weird in the end, but I'm not sure.

ros2_batch_job/util.py
```diff
@@ -183,6 +199,7 @@ def on_process_exited(self, returncode):
                 fargs=(" ".join(self.cmd), returncode))
 
 
+
```
clalancette, line 202: Please remove this extraneous whitespace change.

ros2_batch_job/util.py
```diff
@@ -167,12 +167,28 @@ def warn(*args, **kwargs):
 
 class MyProtocol(AsyncSubprocessProtocol):
     def __init__(self, cmd, exit_on_error, *args, **kwargs):
-        self.cmd = cmd
+        self.cmd = cmd
```
clalancette, line 170: Please remove this extraneous whitespace change.

ros2_batch_job/util.py
sunbo57123: image
This is a part of pip install process. If it is less than 10, the last few lines will be printed.

clalancette: OK, yeah.  It is slightly weird in that the end `|` moves around, but I think that is an acceptable tradeoff here.  Thanks!

ros2_batch_job/util.py
```diff
@@ -169,10 +169,26 @@ class MyProtocol(AsyncSubprocessProtocol):
     def __init__(self, cmd, exit_on_error, *args, **kwargs):
         self.cmd = cmd
         self.exit_on_error = exit_on_error
+        self.progress_bar = False
+        self.progress_data = []
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
-        sys.stdout.write(data.decode('utf-8', 'replace').replace(os.linesep, '\n'))
+        # '[?25l' and '[?25h' can be considered as a sign of begining and end of pip progress bar
```
clalancette, line 177: I'm looking for something describing why we are doing this (though knowing the mechanism a bit can be helpful).  Something more like:

```suggestion
        # The 'pip install' progress bar output can be pretty noisy in CI output.  Print only every 10th
        # output to reduce the noisiness.  We detect the pip progress bar by the beginning sequence of
        # '[?25l' and the end sequence of '[?25h'.
```

sunbo57123:
> The problem with this code is that it gets executed for every command that is run, even ones that will never have the markers we are looking for. This will make the commands slower, but more importantly may inadvertently start executing for commands that we don't want them for.
> 
> So I'm going to suggest we do something different. Instead of modifying the `MyProtocol` class, let's instead copy the whole class with your new code into one called `PipProtocol`. Then we can modify `run_with_prefix` (which is the only place the `pip` command is called from) to use `PipProtocol` instead of `MyProtocol`. That way, we'll only run this detection code for when we need it.
> 
> Does that make sense?

Thanks for your suggestions, making pip protocol in a separate class sounds great.

Could it be achieved by filtering `cmd` in `_run`?
If "pip" and "install" are included in the cmd, the protocol will switch to "PipProtocol".

ros2_batch_job/util.py
sunbo57123: Much thanks for this better explanation!

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11143/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6435/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9090/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11062/)
* Linux packaging [Build Status](http://ci.ros2.org/job/ci_packaging_linux/380/)
* Windows packaging [Build Status](http://ci.ros2.org/job/ci_packaging_windows/158/)

clalancette:
@sunbo57123 Looking at the output from the CI jobs, it looks like this is definitely an improvement.  Particularly for large packages it cuts down on the output quite significantly.  So that's good.

However, I noticed that on a very large package (like mypy), some of the lines aren't output properly.  See https://ci.ros2.org/job/ci_linux/11143/consoleFull, and search for mypy.  Can you look into why that might be?

sunbo57123:
hi @clalancette 
I have found the reason for this exception, the data which is considered as one piece of progress bar may contain multiple progress bars, so one data is not equal to one progress bar. These abnormal data cause this unusual display. This exception occurs randomly, may be related to python flush operation. 
I have found one solution by changing the log analysis way, it may work correctly.

clalancette:
Great, thanks for looking into it.  Here's another CI run to see what it looks like (just Linux):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11156/)

sunbo57123:
> Great, thanks for looking into it. Here's another CI run to see what it looks like (just Linux):
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/11156/)

Much thanks for your help! (while... This seems like the old version, could you please check it?)

clalancette:
> Much thanks for your help! (while... This seems like the old version, could you please check it?)

Yeah, I made a mistake when running CI.  I'm running it again.

clalancette:
@sunbo57123 So this is looking better, but not quite all the way.  If you look closely at mypy again, you'll notice two things:

1.  Right at the beginning of installing mypy, it is missing the "Download mypy-0.780-cp38-cp38-manylinux1_x86_64.whl".
1.  Right at the end there are some extra whitespace.

The second one isn't that important, but the first one would be really good to fix.  Thanks!

sunbo57123:
> @sunbo57123 So this is looking better, but not quite all the way. If you look closely at mypy again, you'll notice two things:
> 
> 1. Right at the beginning of installing mypy, it is missing the "Download mypy-0.780-cp38-cp38-manylinux1_x86_64.whl".
> 2. Right at the end there are some extra whitespace.
> 
> The second one isn't that important, but the first one would be really good to fix. Thanks!

hi @clalancette ,
These 2 problems are fixed [my result](http://119.8.102.15:8080/job/ci_linux/64/consoleFull), Thank you for your careful review, I mistakenly ignored this part.

clalancette:
All right, here is another try at CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11165/)

clalancette:
That looks good to me!  I'm going to do one more review of the code.

ros2_batch_job/util.py
```diff
@@ -183,6 +183,45 @@ def on_process_exited(self, returncode):
...
+
+    def on_stdout_received(self, data):
+        # The 'pip install' progress bar output can be pretty noisy in CI output.  Print only every 10th
+        # output to reduce the noisiness.  We detect the pip progress bar by the beginning sequence of
+        # '[?25l' and the end sequence of '[?25h'.
+        if b'[?25l' in data:
+            self.progress_bar = True
+            sys.stdout.write(data.decode('utf-8', 'replace').replace('\r', '')+'\n')
+        elif b'[?25h' in data:
+            for display_line in self.progress_data.replace("\n","").split("\r"):
```
clalancette, line 202:
```suggestion
            for display_line in self.progress_data.replace('\n', '').split('\r'):
```

ros2_batch_job/util.py
```diff
@@ -183,6 +183,45 @@ def on_process_exited(self, returncode):
...
+                sys.stdout.write(display_line+'\n' if display_line else '')
```
clalancette, line 203:
```suggestion
                sys.stdout.write(display_line + '\n' if display_line else '')
```

ros2_batch_job/util.py
```diff
@@ -183,6 +183,45 @@ def on_process_exited(self, returncode):
...
+            self.progress_bar = False
+            self.progress_data = ''
+            sys.stdout.write(data.decode('utf-8', 'replace').replace('\r', ''))
+        elif self.progress_bar:
+            self.progress_data += data.decode('utf-8', 'replace')
+            if len(self.progress_data) >= 1000:
+                sys.stdout.write(self.progress_data.split("\r")[-1]+'\n')
```
clalancette, line 210:
```suggestion
                sys.stdout.write(self.progress_data.split('\r')[-1] + '\n')
```

ros2_batch_job/util.py
```diff
@@ -183,6 +183,45 @@ def on_process_exited(self, returncode):
...
+class PipProtocol(AsyncSubprocessProtocol):
+    def __init__(self, cmd, exit_on_error, *args, **kwargs):
+        self.cmd = cmd
+        self.exit_on_error = exit_on_error
+        self.progress_bar = False
+        self.progress_data = ''
+        AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
+
+    def on_stdout_received(self, data):
+        # The 'pip install' progress bar output can be pretty noisy in CI output.  Print only every 10th
```
clalancette, line 195: This comment is now slightly wrong.  Now we print only once we've gotten 1000 characters into the string, and when we do print, we just print the last line that came in.  So please update the comment to reflect that.

sunbo57123:
> A few small things to fix. Once these are fixed, I'll do another CI run on all platforms, then I think this will be good to go.

Problems are fixed. Looking forward to another CI running.

clalancette:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11171/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6448/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9105/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11080/)

clalancette:
The builds are yellow, but that's because of tests that are known to have problems.  For this particular PR, I think things are looking really good.  I'm gonig to go ahead and merge this; thanks @sunbo57123 for the work!

PR #462, opened 2020-06-03
Pin VS BuildTools to 16.5.5 in order to avoid broken VC++ download.

brawner:
The daily Dockerfile invalidation updated VS BuildTools to 16.6 which is apparently missing the visual studio package download for Visual C++ (as reported [here](https://developercommunity.visualstudio.com/content/problem/1060913/1661-a-package-is-not-found-for-download-no-more-6.html)). The result is that the python wheel installation of netifaces fails although without a C++ compiler the rest of the build isn't likely to pass if we worked around that.

By pinning back to the earlier Build Tools version we keep the old package which is still available and thus can continue building. We should be able to lift this pin when an updated Build Tools is available which includes the missing VC++ package.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/windows_docker_resources/Dockerfile.dashing
+++ b/windows_docker_resources/Dockerfile.dashing
@@ -138,6 +138,7 @@ COPY rticonnextdds-src\rti_security_plugins-5.3.1-target-x64Win64VS2017.rtipkg C
 RUN ""%ProgramFiles%\rti_connext_dds-5.3.1\bin\rtipkginstall.bat" C:\TEMP\connext\rti_security_plugins-5.3.1-target-x64Win64VS2017.rtipkg"
 
 # Visual Studio Build Tools and .Net SDK`
+# Pinning to 16.5.5 as building the netifaces wheel fails to find VC++
 ADD https://download.visualstudio.microsoft.com/download/pr/68d6b204-9df0-4fcc-abcc-08ee0eff9cb2/17af83ed545d1287df7575786c326009e459708b60a6821d2a4f5606ef8efb9e/vs_BuildTools.exe C:\TEMP\
 # ADD https://aka.ms/vs/16/release/vs_BuildTools.exe C:\TEMP\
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Any comment to why this is done would be helpful in this ticket.

brawner:
Added a comment to the dockerfiles. And pinned to 16.5.5 instead. CI build underway

[Build Status](https://ci.ros2.org/job/ci_windows/10900/)

I have to go walk my dog, so if this passes and people want to merge, or modify the PR and merge please feel free.

nuclearsandwich:
> Any comment to why this is done would be helpful in this ticket.

I re-wrote the PR body and title with more context.

brawner:
I don’t think I configured the CI job correctly. It installed 16.6.1 instead of 16.5.5. I accidentally set ci_branch_to_test instead of CI_Scripts_branch

cottsay:
> I accidentally set ci_branch_to_test instead of ci_scripts_branch

[Build Status](https://ci.ros2.org/job/ci_windows/10901/)

brawner:
Thanks Scott!

brawner:
Ok passing build. Merging

dirk-thomas:
@brawner Please create a follow up ticket to unpin.

dirk-thomas:
@brawner Did you run a full CI build with this change? Could these new warnings be related to this change: https://ci.ros2.org/job/ci_windows/10912/msbuild/new/

cottsay:
> Could these new warnings be related to this change: https://ci.ros2.org/job/ci_windows/10912/msbuild/new/

Maybe the updated version of VS has something to do with why they're new being reported on the buildfarm, but those warnings are coming from ros2/rclcpp#1123. Seems like they should have been present all along, so it seems to me that this is an improvement.

brawner:
> Did you run a full CI build with this change?

No, just up to rcpputils. It's odd that reverting to an older version of VS buildtools would have brought up more warnings. Hopefully there isn't a bloodbath tonight...

cottsay:
> Hopefully there isn't a bloodbath tonight...

I'd certainly rather we see some existing problems come to light than not get Windows builds at all :P

brawner:
Followup issue https://github.com/ros2/ci/issues/463

Issue #467, opened 2020-06-08
Error in ros2_batch_job due to pytest being imported before install.


nuclearsandwich:
Windows and macOS nightlies are failing after https://github.com/ros2/ci/pull/415/ because pytest is installed as a pip dependency during the batch job. 

The failure is not present on Linux because pytest is installed as an apt dependency of python3-pytest-mock: https://github.com/ros2/ci/blob/3f57d6d61488743d9cff27c53f89f4a2daffdf82/linux_docker_resources/Dockerfile#L62

j-rivero:
Does it make sense to move the import statement just before using it in the code? Should be a simple fix if it does not bring undesired side effects. I'll test that and send a new PR if it goes well.

nuclearsandwich:
Since we are reverting #469 this issue is open again. It is worth stressing that any solution which relies on the pytest version outside the venv matching the version inside is incomplete without enforcing that on all platforms and in all situations where a venv is used.

clalancette:
I think we've removed everything having to do with this problem, so I'll go ahead and close this.  Feel free to reopen if I'm mistaken.

PR #468, opened 2020-06-08
Revert "Force xunit2 for pytest 6.0 when legacy mode is enabled"

nuclearsandwich:
Reverts ros2/ci#415 in order to restore Windows and macOS CI which is broken by the absence of pytest in early stages of the build on those platforms. https://github.com/ros2/ci/issues/467 is open to track restoring #415 with corrected behavior after this revert is merged.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -14,11 +14,9 @@
 
 import argparse
 import configparser
-from distutils.version import StrictVersion
 import os
 from pathlib import Path
 import platform
-import pytest
 from shutil import which
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
CI up to rcutils only since we just want to know that builds can run.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11040/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6366/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8992/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10964/)

PR #471, opened 2020-06-09
Force xunit2 for pytest 6.0 when legacy mode is enabled (take III)

j-rivero:
Part III of my particular drama to implement correctly the check for pytest 6.0. The original implementation was in #415 and fixed after in #469. This last fix wrongly assumed that the pytest version check was happening inside the virtualenv. It was not detected by the CI jobs in the PR since the build node that run it happen to have pytest installed in the base system.

This PR tries to run the check inside the virtualenv on the testing job. First commit db9e19c back to enable the reverted code from yesterday, second commit are the real changes 8b6b63d I made to run on virtualenv.

Tested:
 * linux [Build Status](https://ci.ros2.org/job/ci_linux/11072/)
 * linux_aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6386/)
 * win [Build Status](https://ci.ros2.org/job/ci_windows/10988/)
 * osx [Build Status](https://ci.ros2.org/job/ci_osx/9028/)
 * osx in mini1 (doesn't have pytest in base system) [Build Status](https://ci.ros2.org/job/ci_osx/9034/)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -14,7 +14,6 @@
 
 import argparse
 import configparser
-from distutils.version import StrictVersion
 import os
 from pathlib import Path
 import platform
@@ -387,9 +386,13 @@ def build_and_test(args, job):
         ini_file.write('[pytest]\njunit_family=xunit2')
     # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
     # and patch configuration if needed to force the xunit2 value
-    # Import pytest here instead of the top of the file, see https://github.com/ros2/ci/issues/467
-    import pytest
-    xunit_6_or_greater = StrictVersion(pytest.__version__) >= StrictVersion('6.0.0')
+    xunit_6_or_greater = job.run([
+        '"%s"' % job.python, '-c', '\''
+        'from distutils.version import StrictVersion;'
+        'import pytest;'
+        'import sys;'
+        'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
+        '\''])
     for path in Path('.').rglob('pytest.ini'):
         config = configparser.ConfigParser()
         config.read(str(path))
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -362,12 +362,12 @@ def build_and_test(args, job):
     # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
     # and patch configuration if needed to force the xunit2 value
     pytest_6_or_greater = job.run([
-        '"%s"' % job.python, '-c', '\''
+        '"%s"' % job.python, '-c', '"'
         'from distutils.version import StrictVersion;'
         'import pytest;'
         'import sys;'
         'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-        '\''])
+        '"'])
     for path in Path('.').rglob('pytest.ini'):
         config = configparser.ConfigParser()
         config.read(str(path))
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -396,7 +396,7 @@ def build_and_test(args, job):
     for path in Path('.').rglob('pytest.ini'):
         config = configparser.ConfigParser()
         config.read(str(path))
-        if xunit_6_or_greater:
+        if pytest_6_or_greater:
             # only need to correct explicit legacy option if exists
             if not check_xunit2_junit_family_value(config, 'legacy'):
                 continue
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -356,37 +356,37 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: test')
 
-  # In Foxy and prior, xunit2 format is needed to make Jenkins xunit plugin 2.x happy
+    # In Foxy and prior, xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     # After Foxy, we introduced per-package changes to make local builds and CI
     # builds act the same.
     if args.ros_distro in ('dashing', 'eloquent', 'foxy'):
-	    # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
-	    with open('pytest.ini', 'w') as ini_file:
-		ini_file.write('[pytest]\njunit_family=xunit2')
-	    # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
-	    # and patch configuration if needed to force the xunit2 value
-	    pytest_6_or_greater = job.run([
-		'"%s"' % job.python, '-c', '\''
-		'from distutils.version import StrictVersion;'
-		'import pytest;'
-		'import sys;'
-		'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-		'\''])
-	    for path in Path('.').rglob('pytest.ini'):
-		config = configparser.ConfigParser()
-		config.read(str(path))
-		if pytest_6_or_greater:
-		    # only need to correct explicit legacy option if exists
-		    if not check_xunit2_junit_family_value(config, 'legacy'):
-		        continue
-		else:
-		    # in pytest < 6 need to enforce xunit2 if not set
-		    if check_xunit2_junit_family_value(config, 'xunit2'):
-		        continue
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -384,18 +384,27 @@ def build_and_test(args, job):
...
-            if config.get('pytest', 'junit_family') == 'xunit2':
+        if xunit_6_or_greater:
+            # only need to correct explicit legacy option if exists
+            if not check_xunit2_junit_family_value(config, 'legacy'):
                 continue
-        except configparser.NoOptionError:
-            pass
-        print('xunit2 patch applied to ' + str(path))
+        else:
+            # in xunit < 6 need to enforce xunit2 if not set
```
nuclearsandwich, line 404: I think this is a typo. Isn't it the pytest version we're checking here?
```suggestion
            # in pytest < 6 need to enforce xunit2 if not set
```

ros2_batch_job/__main__.py
```diff
@@ -384,18 +384,27 @@ def build_and_test(args, job):
     # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
-    # check if packages have a pytest.ini file and add the xunit2
-    # format if it is not present
+    # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
+    # and patch configuration if needed to force the xunit2 value
+    xunit_6_or_greater = job.run([
```
nuclearsandwich, line 389: I think this is a typo. Isn't it the pytest version we're checking here?
```suggestion
    pytest_6_or_greater = job.run([
```

ros2_batch_job/__main__.py
```diff
@@ -384,18 +384,27 @@ def build_and_test(args, job):
...
+        'import sys;'
+        'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
+        '\''])
     for path in Path('.').rglob('pytest.ini'):
         config = configparser.ConfigParser()
         config.read(str(path))
-        try:
-            # only if xunit2 is set continue the loop with the file unpatched
-            if config.get('pytest', 'junit_family') == 'xunit2':
+        if xunit_6_or_greater:
```
nuclearsandwich, line 399: Fix variable name to match earlier suggestion.

```suggestion
        if pytest_6_or_greater:
```

clalancette:
Can I make a suggestion here that might negate the need for this PR completely?

In order to have clean test runs, we need to install a pytest.ini file in all of the relevant repos.  Otherwise, pytest throws this xunit2 warning.  While this works around it, it still causes these warnings to show up when running locally.  I've been slowly adding pytest.ini to the relevant places as I have noticed them, but I might suggest that now would be a good opportunity to finish that work off.  Then we can remove this workaround from CI completely.

Thoughts?

nuclearsandwich:
> I've been slowly adding pytest.ini to the relevant places as I have noticed them, but I might suggest that now would be a good opportunity to finish that work off. Then we can remove this workaround from CI completely.

As long as pytest 3.3.2 in Bionic supports the configuration option I'm completely totally 100% in favor of that approach. I was reviewing under the assumption that adding the appropriate configuration directly to affected repositories was off the table, otherwise why work around it.

If we can't backport the config option to Bionic it will affect Dashing and Eloquent CI runs on ci.ros2.org.

clalancette:
Let me see what it will take to get rid of it for master.  Then I'll look at the options for backporting to bionic.

nuclearsandwich:
@j-rivero if there is some reason we don't want to or can't make these changes at the source level can you add them to the PR description and let us know.

clalancette:
All right, it looks doable.  It will take 19 PRs (that's just for master), plus probably an equivalent amount for Eloquent and Dashing, but I think that is a better way forward.  Once I get all of the PRs merged to master, I'll do a CI run with this workaround removed completely and see what it looks like.

clalancette:
(also, as far as I know this should work fine for Ubuntu 18.04, so eloquent and dashing should be fine)

j-rivero:
> @j-rivero if there is some reason we don't want to or can't make these changes at the source level can you add them to the PR description and let us know.

I [mentioned that briefly](https://github.com/ros2/ci/pull/406#issue-386759713) at the beginning of all the PRs serie. Nothing against it but that fix relies on no changes in current code with respect to how the format is configured in pytest.ini files and the same for new packages. If one of them is miss configured the whole build fails.

In order to be sure that the build is not going to fail in any case we can keep these changes in place together with fixes in source code level.

dirk-thomas:
> If one of them is miss configured the whole build fails.

Why would the build fail? It only results in a warning is a package doesn't set the option and the workaround in CI is removed.

dirk-thomas:
If all packages have the option defined we should revert the working in CI since there is a) no need for it and b) we want to be notified that any new package doesn't have option set (since a user building from source will experience the same).

j-rivero:
> Why would the build fail? It only results in a warning is a package doesn't set the option and the workaround in CI is removed.

Would not be the same situation that leaded to https://github.com/ros2/ci/pull/393? The `junit_family` value is not the one expected by the xunit 2.x Jenkins plugin so the check on the format fails and the Jenkins plugin kills the whole build.

dirk-thomas:
@j-rivero Can you please address the pending feedback so we can get this PR merged.

As a follow up we want to remove the workaround to generate custom `pytest.ini` files in CI for Rolling and forward.

j-rivero:
> @j-rivero Can you please address the pending feedback so we can get this PR merged.

ready. A latest run to check these latest changes: [Build Status](https://ci.ros2.org/job/ci_windows/11314/)

j-rivero:
>  A latest run to check these latest changes

probably would need a rebase or other alternatives to get changes from master in order to work.

dirk-thomas:
@j-rivero What is the status on this ticket?

j-rivero:
> @j-rivero What is the status on this ticket?

ready to merge as far as I can tell. I've merge master to get the final test build: [Build Status](https://ci.ros2.org/job/ci_windows/11669/)

ros2_batch_job/__main__.py
```diff
@@ -359,18 +359,27 @@ def build_and_test(args, job):
     # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     with open('pytest.ini', 'w') as ini_file:
         ini_file.write('[pytest]\njunit_family=xunit2')
-    # check if packages have a pytest.ini file and add the xunit2
-    # format if it is not present
+    # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
+    # and patch configuration if needed to force the xunit2 value
+    pytest_6_or_greater = job.run([
+        '"%s"' % job.python, '-c', '\''
```
dirk-thomas, line 365: Nitpick: assuming a double quote should work to use `'"'` instead of `'\''`.

Same for the closing quote.

j-rivero: this can not use double quotes since they are being using in 369.

dirk-thomas: Then avoid the escaping with `"'"`.

dirk-thomas: Or use single quotes for the version string in line 369.

j-rivero: thanks, I used `" ' "`

j-rivero:
After merge with default and solve the quoting problem, last test run [Build Status](https://ci.ros2.org/job/ci_windows/11713/). Merging.

ros2_batch_job/__main__.py
```diff
@@ -360,20 +360,30 @@ def build_and_test(args, job):
...
     # builds act the same.
     if args.ros_distro in ('dashing', 'eloquent', 'foxy'):
+        # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
         with open('pytest.ini', 'w') as ini_file:
             ini_file.write('[pytest]\njunit_family=xunit2')
-        # check if packages have a pytest.ini file and add the xunit2
-        # format if it is not present
+        # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
+        # and patch configuration if needed to force the xunit2 value
+        pytest_6_or_greater = job.run([
```
jacobperron, line 368: Looks like if this check fails, it fails the build ([for example](https://ci.ros2.org/job/ci_linux/11784/)).

jacobperron: See https://github.com/ros2/ci/pull/498 for a proposed fix.

PR #476, opened 2020-06-16
Add script to calculate coverage rates

j-rivero:
Following up discussion in https://github.com/ros2/ros2_documentation/pull/737#discussion_r436415662 , I'm importing the coverage calculator script to this repository. Not sure if the location proposed in the best, feel free to propose an alternative.

The script uses the new `--shallow` flag in vcstool version `0.2.10-1`.

/cc @chapulina

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,4 +1,4 @@
-# Coverage calculator for ROS 2 packages in ROS2 buildfarm
+# Coverage calculator for ROS 2 packages in ROS 2 buildfarm
 
 Script to get results from a `ci.ros2.org` coverage build through json and
 calculate the coverage rate from all the listed entries corresponding to
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,14 @@
+# Coverage calculator for ROS2 packages in ROS2 buildfarm
+
+Script to get results from a `ci.ros2.org` coverage build through json and
+calculate the coverage rate from all the listed entries corresponding to
+unit testing.
+
+## Usage
+
+```
+  $ ./get_coverage_ros2_pkg.py <jenkins_build_url> <ros2_package_name>
+
+Example:
+  $ ./get_coverage_ros2_pkg.py https://ci.ros2.org/job/ci_linux_coverage/182 rclcpp
+```
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,4 +1,4 @@
-# Coverage calculator for ROS2 packages in ROS2 buildfarm
+# Coverage calculator for ROS 2 packages in ROS2 buildfarm
 
 Script to get results from a `ci.ros2.org` coverage build through json and
 calculate the coverage rate from all the listed entries corresponding to
--- a/tools/get_coverage_ros2_pkg.py
+++ b/tools/get_coverage_ros2_pkg.py
@@ -104,11 +104,11 @@ if __name__ == '__main__':
                 total_lines_under_testing += lines_coverage['denominator']
                 total_lines_tested += lines_coverage['numerator']
                 print(f' * {entry_name}'
-                      f' [{lines_coverage['ratio']:.2f}] --'
-                      f' {lines_coverage['numerator']:.0f}/{lines_coverage['denominator']:.0f}')
+                      f" [{lines_coverage['ratio']:.2f}] --"
+                      f" {lines_coverage['numerator']:.0f}/{lines_coverage['denominator']:.0f}")
 
         if total_lines_under_testing == 0:
-            print('Package not found: ' + input_pkg, file=sys.stderr)
+            print(f'No coverage found for package {input_pkg}', file=sys.stderr)
             sys.exit(-1)
 
         print(f'\nCombined unit testing for {input_pkg}:'
--- a/tools/get_coverage_ros2_pkg.py
+++ b/tools/get_coverage_ros2_pkg.py
@@ -25,6 +25,7 @@ def get_src_path(package_name, colcon_ws):
     if not path:
         print('Package not found: ' + input_pkg, file=sys.stderr)
         sys.exit(-1)
+    assert path.count('\n') == 0, 'Found more than one line returned by colcon list'
     return path
 
 
--- a/tools/get_coverage_ros2_pkg.py
+++ b/tools/get_coverage_ros2_pkg.py
@@ -15,30 +15,6 @@ def is_slice_in_list(s, l):
     return any(s == l[i:len_s + i] for i in range(len(l) - len_s + 1))
 
 
-def create_colcon_workspace():
-    ros2_ws_path = os.path.join(tempfile.gettempdir(), 'ros2_coverage_tool')
-    ros2_repos_path = os.path.join(ros2_ws_path, 'ros2.repos')
-    # create the workspace if it does not exist
-    if not os.path.isdir(ros2_ws_path):
-        os.mkdir(ros2_ws_path)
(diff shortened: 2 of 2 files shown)


tools/README.md
```diff
@@ -0,0 +1,14 @@
+# Coverage calculator for ROS2 packages in ROS2 buildfarm
```
dirk-thomas, line 1: Nitpick: space between `ROS` and `2`.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import requests
+import subprocess
+import sys
+import tempfile
+
+parser = argparse.ArgumentParser()
```
dirk-thomas, line 10: Best practice recommends to move all lines which should be run when the file is invoked into a conditional block: `if __name__ == '__main__':`

That allows to import the file without executing the logic.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,121 @@
...
+parser.add_argument("jenkins_coverage_build", help="URL of a ci.ro2s.org build using coverage (i.e https://ci.ros2.org/job/ci_linux_coverage/182)")
```
dirk-thomas, line 11: Nitpick: following the ROS 2 developer guide use single quotes except if the string literal contains them.

Same below.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Compare if a list is inside other list in order
+# From here: https://stackoverflow.com/a/20789669
+def is_slice_in_list(s, l):
+    len_s = len(s)
+    return any(s == l[i:len_s + i] for i in range(len(l) - len_s + 1))
+
+
+def create_colcon_workspace():
+    ros2_ws_path = os.path.join(tempfile.gettempdir(), 'ros2_coverage_tool')
```
dirk-thomas, line 27: The temporary directory should be removed when the script doesn't need it anymore. I would suggest using `tempfile.TemporaryDirectory` with a context manager.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,121 @@
...
+    cmd = ['colcon', 'list', '--paths-only', '--base-paths', colcon_ws, '--packages-select', package_name]
+    try:
+        path = subprocess.check_output(cmd).decode('ascii').strip()
+    except subprocess.CalledProcessError as e:
+        print(e.output, file=sys.stderr)
+        sys.exit(-1)
+    if not path:
+        print("Package not found: " + input_pkg, file=sys.stderr)
+        sys.exit(-1)
+    return path
```
dirk-thomas, line 60: Maybe add a sanity check that `colcon list` only returned a single line (since technically `colcon` can handle packages with the same name).

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,121 @@
...
+    elif is_slice_in_list(input_pkg_rel_path, name_parts):
+        # source code: check if coverage entry contains exactly the source path
+        package_under_cov = input_pkg
+    else:
+        package_under_cov = "NOT-FOUND"
+
+    if package_under_cov == input_pkg:
+        total_lines_under_testing += lines_coverage['denominator']
+        total_lines_tested += lines_coverage['numerator']
+        print(" * %s [%04.2f] -- %i/%i" % (
```
dirk-thomas, line 106: Nitpick: `%i` is rarely used in Python and isn't support in format strings in Python 3. Please use `%f` instead.

In general using f-strings instead of `'...' % (...)` is preferred for newly written code.

Same below.

tools/README.md
j-rivero: c6602a6

tools/get_coverage_ros2_pkg.py
j-rivero: a507211 and c6cebed

tools/get_coverage_ros2_pkg.py
j-rivero: f2f652f

tools/get_coverage_ros2_pkg.py
j-rivero: d501288

tools/get_coverage_ros2_pkg.py
j-rivero: e64aca2

tools/get_coverage_ros2_pkg.py
j-rivero: 10d0915

tools/README.md
dirk-thomas: Same again later in the same line.

tools/get_coverage_ros2_pkg.py
dirk-thomas: Still some double quotes.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,116 @@
...
+
+            if package_under_cov == input_pkg:
+                total_lines_under_testing += lines_coverage['denominator']
+                total_lines_tested += lines_coverage['numerator']
+                print(f" * {entry_name}"
+                      f" [{lines_coverage['ratio']:.2f}] --"
+                      f" {lines_coverage['numerator']:.0f}/{lines_coverage['denominator']:.0f}")
+
+        if total_lines_under_testing == 0:
+            print('Package not found: ' + input_pkg, file=sys.stderr)
```
dirk-thomas, line 111: The message seems to not describe what was being tested.

tools/get_coverage_ros2_pkg.py
```diff
@@ -0,0 +1,116 @@
...
+    cmd = ['colcon', 'list', '--paths-only', '--base-paths', colcon_ws, '--packages-select', package_name]
+    try:
+        path = subprocess.check_output(cmd).decode('ascii').strip()
+    except subprocess.CalledProcessError as e:
+        print(e.output, file=sys.stderr)
+        sys.exit(-1)
+    if not path:
+        print('Package not found: ' + input_pkg, file=sys.stderr)
+        sys.exit(-1)
+    assert path.count('\n') == 0, 'Found more than one line returned by colcon list'
```
dirk-thomas, line 28: I would suggest to use `len(path.splitlines())` to be platform / newline agnostic and not care about trailing newlines.

tools/README.md
j-rivero: ouch sorry for that, 64d5a3f

tools/get_coverage_ros2_pkg.py
j-rivero: I changed the only place I found where I'm not using single quotes inside the string, e5bddaa . Am I missing other occurrences? 4d4d9fe

tools/get_coverage_ros2_pkg.py
j-rivero: I changed it to say it does not found coverage lines for package 4d4d9fe

tools/get_coverage_ros2_pkg.py
j-rivero: better, thanks 95415ba

tools/get_coverage_ros2_pkg.py
dirk-thomas: Only one more: https://github.com/ros2/ci/pull/476/files#diff-b160c856c85886178880e50860b3281fR76

PR #447, opened 2020-05-06
Parse pytest warning summaries from colcon logs.

nuclearsandwich:
The parser itself is actually part of the primary Jenkins configuration
and requires separate documentation and potentially being separated out as a warning
parser plugin.

Groovy-based parsers are not allowed to run on console output so this
scans the stderr.log files from colcon test output.

A test build with this parser enabled is here: https://ci.ros2.org/job/test_ci_linux/39/colcon-log-pytest-warnings

The regexp is
```
^\[[^]]+] .* \.\.\. (?<path>[^:]*):(?<lineno>\d+): (?<exname>[^:]*): (?<message>.*)$
```

and the Groovy script
```groovy
import edu.hm.hafner.analysis.Severity

builder.setFileName(matcher.group("path"))
        .setLineStart(Integer.parseInt(matcher.group("lineno")))
        .setSeverity(Severity.WARNING_NORMAL)
        .setType(matcher.group("exname"))
        .setMessage(matcher.group("message"))

return builder.buildOptional();
```

The above assumes the colcon log is printing elapsed time and the test which triggered the warning is being printed before the `...` by Pytest.

PR #477, opened 2020-06-16
read stdout_stderr.log from build steps to extract CMake / compiler warnings

### LINKED ISSUE #473: False positive Clang warnings on Linux from openssl output.
Lines such as

```
258: validate_local_permissions failed: (127) Failed to parse PKCS7 SMIME document: 281473198688096:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:../crypto/pkcs7/pk7_smime.c:284:Verify error:self signed certificate
```

from cyclonedds security tests are matching the [Jenkins analysis model regexp](https://github.com/jenkinsci/analysis-model/blob/ccc7d876160b5c610e62e103de08715e9609cada/src/main/java/edu/hm/hafner/analysis/parser/ClangParser.java#L20-L21) as a Clang warning.

[Linux aarch64 debug #1234](https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1234/clang/) is an example build exhibiting these warnings.

I've done some quick checking to see if there's a described format for clang messages in order to improve upstream's regexp and haven't found anything yet. Since our normal Linux builds use GCC we don't necessarily need to look at them for Clang errors so I'm going to open a PR that adds that parser only conditionally.

This is related to https://github.com/ros2/ci/issues/316

dirk-thomas:
Instead of using the whole console output which includes output from other steps of the process, e.g. testing as well as repeated `stderr` output from `colcon`.

Addresses #316 on ci.ros2.org. Fixes #473.

CI builds without this change (and a custom branch to produce a compiler warning):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11162/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6445/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9101/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11076/)

CI build with this change (the jobs will go away once this PR has been merged):
* Linux [Build Status](https://ci.ros2.org/job/ci_linux-dirk/1/) (still extracting the compiler warnings, but without the duplicate GCC warning)
* macOS: [Build Status](https://ci.ros2.org/job/ci_osx-dirk/3/) (still extracting the compiler warnings)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows-dirk/1/) (still extracting the compiler warnings) (no compiler warning anymore)

False positive resolved by this patch:
Before: [Build Status](https://ci.ros2.org/job/ci_linux/11168/) (test output was classified as a compiler warning)
After (the jobs will go away once this PR has been merged): [Build Status](https://ci.ros2.org/job/ci_linux-dirk/2/)

### DIFF
--- a/job_templates/snippet/publisher_warnings_ng.xml.em
+++ b/job_templates/snippet/publisher_warnings_ng.xml.em
@@ -3,7 +3,7 @@
     <io.jenkins.plugins.analysis.warnings.Cmake>
       <id></id>
       <name></name>
-      <pattern></pattern>
+      <pattern>ws/log/latest_build/*/stdout_stderr.log</pattern>
       <reportEncoding></reportEncoding>
       <skipSymbolicLinks>false</skipSymbolicLinks>
     </io.jenkins.plugins.analysis.warnings.Cmake>
@@ -11,7 +11,7 @@
    <io.jenkins.plugins.analysis.warnings.Gcc4>
       <id></id>
       <name></name>
-      <pattern></pattern>
+      <pattern>ws/log/latest_build/*/stdout_stderr.log</pattern>
       <reportEncoding></reportEncoding>
       <skipSymbolicLinks>false</skipSymbolicLinks>
     </io.jenkins.plugins.analysis.warnings.Gcc4>
@@ -20,14 +20,14 @@
     <io.jenkins.plugins.analysis.warnings.Clang>
       <id></id>
       <name></name>
-      <pattern></pattern>
+      <pattern>ws/log/latest_build/*/stdout_stderr.log</pattern>
       <reportEncoding></reportEncoding>
       <skipSymbolicLinks>false</skipSymbolicLinks>
     </io.jenkins.plugins.analysis.warnings.Clang>
     <io.jenkins.plugins.analysis.warnings.ClangTidy>
       <id></id>
       <name></name>
-      <pattern></pattern>
+      <pattern>ws/log/latest_build/*/stdout_stderr.log</pattern>
       <reportEncoding></reportEncoding>
       <skipSymbolicLinks>false</skipSymbolicLinks>
     </io.jenkins.plugins.analysis.warnings.ClangTidy>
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
These all hard-code the path `ws`. Do any builds randomly name the workspace path `workspace` or `work space` anymore?

When I was working on https://github.com/ros2/ci/pull/447 I had to handle the elapsed time output (eg `[0.554s] ...`). Do any of the other warning parsers in use have an anchor to the start of the line (`^`)?

dirk-thomas:
> These all hard-code the path `ws`. Do any builds randomly name the workspace path `workspace` or `work space` anymore?

Good point, I already forgot we have that parameter: d34af4808b1dd06f19ec2ceb9dcac7efac626a99.

> Do any of the other warning parsers in use have an anchor to the start of the line (`^`)?

I don't have an answer for this specific questions. But the above jobs show that compiler warnings are correctly being extracted - for GNU C, Clang, Clang-tidy, and MSBuild.

I have triggered another build to check if CMake warnings are captures too:
[Build Status](https://ci.ros2.org/job/ci_osx-dirk/5/) (one CMake warning as expected)

So I would conclude it is good as is.

azeey:
https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1905/ and https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1666/ failed I think because it was looking for log files in `work space/...`, but the files were in `workspace/...`. @dirk-thomas, can you verify?

dirk-thomas:
> When I was working on https://github.com/ros2/ci/pull/447 I had to handle the elapsed time output (eg `[0.554s] ...`). Do any of the other warning parsers in use have an anchor to the start of the line (`^`)?

While this patch still extracts the warnings correctly the grouping contains the timestamp prefix from the log and as a consequence doesn't deduplicate same warnings (see https://ci.ros2.org/view/packaging/job/packaging_linux-centos/446/clang/#folderContent).

I don't think it is a feasible option to customize the Jenkins warnings handling. So the only viable option I can think of is generating the log files we need to parse without timestamp...

dirk-thomas:
See colcon/colcon-output#33.

PR #478, opened 2020-06-16
set default ROS distro to rolling

dirk-thomas:
Currently there is no logic doing custom stuff for `rolling` (or `foxy`).

### DIFF
--- a/job_templates/snippet/property_parameter-definition_common.xml.em
+++ b/job_templates/snippet/property_parameter-definition_common.xml.em
@@ -60,7 +60,7 @@ choices.remove(ubuntu_distro)
             <a class="string-array">
               <string>@ros_distro</string>
 @{
-choices = ['dashing', 'eloquent', 'foxy']
+choices = ['dashing', 'eloquent', 'foxy', 'rolling]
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
> Currently there is no logic doing custom stuff for `rolling` (or `foxy`).

We do use this parameter on Windows to switch which Dockerfile / dependency set should be used (introduced in #451). I've pushed a commit that configures the rolling release to use the Foxy Dockerfile, which is what we should do until they need to diverge.

job_templates/snippet/property_parameter-definition_common.xml.em
```diff
@@ -60,7 +60,7 @@ choices.remove(ubuntu_distro)
             <a class="string-array">
               <string>@ros_distro</string>
 @{
-choices = ['dashing', 'eloquent', 'foxy']
+choices = ['dashing', 'eloquent', 'foxy', 'rolling]
```
dirk-thomas, line 63: Missing trailing quote fixed in: 25166affe12da3f505bf3564b94c5c7271ac14a8.

PR #479, opened 2020-06-17
combine 'ws' and 'work space' pattern since Jenkins requires them to be non-empty

dirk-thomas:
Fixes regression from #477 which fails build if there are no files matching the pattern.

Test builds:
* without this fix: [Build Status](https://ci.ros2.org/job/ci_linux-dirk/1/)
* with this fix: [Build Status](https://ci.ros2.org/job/ci_linux-dirk/2/)
* with this fix and a compiler warning: [Build Status](https://ci.ros2.org/job/ci_linux-dirk/3/)

### DIFF
--- a/job_templates/snippet/publisher_warnings_ng.xml.em
+++ b/job_templates/snippet/publisher_warnings_ng.xml.em
@@ -1,49 +1,47 @@
 <io.jenkins.plugins.analysis.core.steps.IssuesRecorder plugin="warnings-ng@@7.3.0">
   <analysisTools>
-@[for workspace in ['ws', 'work space']]@
     <io.jenkins.plugins.analysis.warnings.Cmake>
       <id></id>
       <name></name>
-      <pattern>@(workspace)/log/build_*/*/stdout_stderr.log</pattern>
+      <pattern>*/log/build_*/*/stdout_stderr.log</pattern>
       <reportEncoding></reportEncoding>
       <skipSymbolicLinks>false</skipSymbolicLinks>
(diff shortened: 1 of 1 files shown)


job_templates/snippet/publisher_warnings_ng.xml.em
```diff
@@ -1,49 +1,47 @@
 <io.jenkins.plugins.analysis.core.steps.IssuesRecorder plugin="warnings-ng@@7.3.0">
   <analysisTools>
-@[for workspace in ['ws', 'work space']]@
     <io.jenkins.plugins.analysis.warnings.Cmake>
       <id></id>
       <name></name>
-      <pattern>@(workspace)/log/build_*/*/stdout_stderr.log</pattern>
+      <pattern>*/log/build_*/*/stdout_stderr.log</pattern>
```
nuclearsandwich, line 6: I was thinking something more precise like
```suggestion
      <pattern>{ws,work space}/log/build_*/*/stdout_stderr.log</pattern>
```

I think it only matters if we notice these globs taking a long time to resolve.

dirk-thomas: I don't think that syntax is doing what you intend: [Build Status](https://ci.ros2.org/job/ci_linux-dirk/4/) which should have a compiler warning.

> [GNU C Compiler (gcc)] [-ERROR-] No files found for pattern '{ws,work space}/log/build_*/*/stdout_stderr.log'. Configuration error?

PR #481, opened 2020-06-19
Change coverage tool to use lcov

j-rivero:
This is the first PR of a series that will improve the coverage report generation and fix some of the current problems. In this PR, I use `lcov` in combination with `lcov_cobertura_fix` (to transform format to cobertura plugin) and remove the use of `gcovr`.

I found several important advantages using `lcov`:

 * Easy to aggregate tracefiles in a directory (the PR reduces the code in 30 lines)
 * Easier to understand and work with remove capabilities (regular expresions work on code paths)
 * The reports now displays `Method` metric together with the previous metris. @Blast545 found that useful.
 * Open the door for the next PR that will filter the report results based on input packages from user
 * Open the door to be able to merge unit and system coverage coming from different files

Testing:
 * This same branch tested with rclcpp [Build Status](https://ci.ros2.org/view/All/job/test_linux_coverage/30/)
 * Same code running all the default configuration (like a nightly altought excluding python packages not related to this code) [Build Status](https://ci.ros2.org/view/All/job/test_linux_coverage/29/). Using this build and the coverage script I'm not able to see significant differences in our set of packages in the coverage rate of our quality level list.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -274,62 +274,31 @@ def get_args(sysargv=None):
 
 def process_coverage(args, job):
     print('# BEGIN SUBSECTION: coverage analysis')
-    # Collect all .gcda files in args.workspace
-    output = subprocess.check_output(
-        [args.colcon_script, 'list', '--base-paths', args.sourcespace])
-    for line in output.decode().splitlines():
-        package_name, package_path, _ = line.split('\t', 2)
-        print(package_name)
-        package_build_path = os.path.join(args.buildspace, package_name)
-        gcda_files = []
-        for root, dirs, files in os.walk(package_build_path):
-            gcda_files.extend(
-                    [os.path.abspath(os.path.join(root, f))
-                        for f in files if f.endswith('.gcda')])
-        if len(gcda_files) == 0:
-            continue
-
-        # Run one gcov command for all gcda files for this package.
-        cmd = ['gcov', '--preserve-paths', '--relative-only', '--source-prefix', os.path.abspath('.')] + gcda_files
-        print(cmd)
-        subprocess.run(cmd, check=True, cwd=package_build_path)
-
-        # Write one report for the entire package.
-        # cobertura plugin looks for files of the regex *coverage.xml
-        outfile = os.path.join(package_build_path, package_name + '.coverage.xml')
-        print('Writing coverage.xml report at path {}'.format(outfile))
-        # --gcov-exclude remove generated .gcov files from previous gcov call. 
-        #                file names are in the form: #dir#sudir#file.*.gcov     
-        # -xml  Output cobertura xml
-        # -output=<outfile>  Pass name of output file
-        # -g  use existing .gcov files in the directory
-        cmd = [
-            'gcovr',
-            '--object-directory=' + package_build_path,
-            '-k',
-            '-r', os.path.abspath('.'),
-            '--xml', '--output=' + outfile,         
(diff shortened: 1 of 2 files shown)


dirk-thomas:
How does this related to [colcon-lcov-result](https://github.com/colcon/colcon-lcov-result)?

j-rivero:
> How does this related to [colcon-lcov-result](https://github.com/colcon/colcon-lcov-result)?

The approach in this PR is mainly the same of what `colcon-lcov-result` is doing but I think that there is at least one fundamental difference: `colcon-lcov-result` works with the `package-selection` logic of colcon which means that we would need to supply colcon arguments to make it work correctly and not to process all packages in the source tree. In our case we probably really want to mimic the package selection used in `CI_TEST_ARGS` (need to get .gcda files from all packages to be sure we are not losing integration coverage from other packages) and not add another Jenkins parameter for it. Not sure if the package selection could be extracted from `CI_TEST_ARGS` but assuming that is not possible I don't see a better option than using a new parameter.

I built a prototype using `colcon-lcov-result` to get a blackbox overview of reports:
 * rcutils with this branch https://ci.ros2.org/view/All/job/test_linux_coverage/70/cobertura/
 * rcutils with `filter_coverage_colcon` branch: https://ci.ros2.org/view/All/job/test_linux_coverage/69/cobertura/

Mostly the same but in the `colcon-lcov-result` branch I see no coverage for packages under `install/` and lack of Conditionals metric in rcutils. Both can be probably debugged with a bit of time.

dirk-thomas:
> Mostly the same but in the `colcon-lcov-result` branch I see no coverage for packages under `install/` and lack of Conditionals metric in rcutils.

The significant difference between both approaches I see is that reusing something like `colcon-lcov-result` here allows a user to invoke the same logic locally. If the CI system contains separate custom logic there is no way a user can replicate the result locally.

Therefore I think it would be better to contribute to `colcon-lcov-result` to make it e.g. consider `install/` and whatever else it is lacking for our use case.

ros2_batch_job/__main__.py
```diff
@@ -274,62 +274,31 @@ def get_args(sysargv=None):
...
-                with open(coverage_path, 'r') as h:
-                    content = h.read()
-                content = content.replace(
-                    '<source>%s/' % docker_base_path,
-                    '<source>%s/' % args.workspace_path)
-                with open(coverage_path, 'w') as h:
-                    h.write(content)
+    # Capture all gdca/gcno files (all them inside buildspace)
+    coverage_file = os.path.join(args.buildspace, 'coverage.info')
+    cmd = ['lcov',
```
dirk-thomas, line 279: Nitpick: please wrap after the open bracket and indent all following lines one level to match the style guide.

Same below.

j-rivero: ready 726409c

j-rivero:
> The significant difference between both approaches I see is that reusing something like `colcon-lcov-result` here allows a user to invoke the same logic locally. If the CI system contains separate custom logic there is no way a user can replicate the result locally.

I agree on the need of having a way to replicate the logic in local systems.
 
> Therefore I think it would be better to contribute to `colcon-lcov-result` to make it e.g. consider `install/` and whatever else it is lacking for our use case.

I've been looking into that, it requires non trivial changes on how the tool is working, I'm not sure that I would have the bandwidth necessary to port changes to the upstream tool but I'll try.

Final test run: [Build Status](https://ci.ros2.org/view/All/job/test_linux_coverage/71/)

PR #484, opened 2020-06-24
Add importlib-resources to CentOS dockerfile.

clalancette:
This is needed for the new code where we don't use pkg_resources
anymore.

Part of https://github.com/ros2/ros2/issues/919

CentOS packaging before this change: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-centos/22/)
CentOS packaging after this change: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-centos/23/)

### DIFF
--- a/linux_docker_resources/Dockerfile-CentOS
+++ b/linux_docker_resources/Dockerfile-CentOS
@@ -59,7 +59,7 @@ RUN sudo yum update \
 RUN dbus-uuidgen >/etc/machine-id
 
 # Install some stuff via Pip which we don't have available in EPEL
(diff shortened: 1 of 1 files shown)


clalancette:
> @clalancette, will this be a requirement in Foxy, or only moving forward in Galactic?

It somewhat depends.  It may be possible to backport these changes to Foxy, as they mostly change internal implementation details.  But I won't know for sure until I get the full set of fixes in place, i.e. the second big bullet list in https://github.com/ros2/ros2/issues/919 .

So for now, it will just be Galactic.

clalancette:
There's compiler warnings in the CentOS build, but I have to imagine that they were already there (and I find it unlikely that this change introduced them).  So I'm going to go ahead and merge, thanks for the reviews.

PR #485, opened 2020-06-24
disable nightly/packaging armhf jobs

dirk-thomas:
Since they are not passing anyway and there is no plan to make them pass again in the foreseeable future disabling the nightly / packaging jobs avoid the "noise" of these jobs being red.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -168,6 +168,8 @@ def main(argv=None):
 
     # configure os specific jobs
     for os_name in sorted(os_configs.keys()):
+        if os_name != 'linux-armhf':
+            continue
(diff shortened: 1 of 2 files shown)


create_jenkins_job.py
```diff
@@ -168,6 +168,8 @@ def create_job(os_name, job_name, template_file, additional_dict):
 
     # configure os specific jobs
     for os_name in sorted(os_configs.keys()):
+        if os_name != 'linux-armhf':
+            continue
```
dirk-thomas, line 172: Doh, this wasn't support to be part of the PR. Removed in a8f4cb5f3cba00d272082a910293c5c60a3cb069.

PR #486, opened 2020-06-24
add missing build/test args to nightly coverage job

dirk-thomas:
Fixes regression introduced in #466.


sloretz:
> Fixes regression introduced in #466.


Do I understand this correctly:  `data['build_args_default']` and `data['test_args_default']` weren't being explicitly passed in before, but they were implicitly there since `create_job()` starts by copying the `data` dict?

https://github.com/ros2/ci/blob/04a9386b96a791f3aa92698af0914fc7aa4e404e/create_jenkins_job.py#L161


https://github.com/ros2/ci/pull/466/files#diff-42a7cd652d15099c3f6b6c27e05dce31R444-R447

PR #489, opened 2020-06-26
Implement the per-package coverage reports

j-rivero:
The PR adds a new `CI_COVERAGE_FILTER_PKGS` Jenkins parameter to allow the user to specify a set of packages to limit the coverage report to those.

 * First commit 30da629 implements the Jenkins parameter.
 * Second commit cfb2ada adds logic to filter the coverage:
   * cpp/c coverage filtering is done on top of the current lcov file using the `--extract` capacity of lcov.
   * python coverage is generated by `python-cov` extension and that means that packages could have a coverage.xml already generated. To manage filtering, I've modify the path used by the cobertura plugin to only report coverage.xml files under `ws/build/*coverage.xml` and if the `CI_COVERAGE_FILTER_PKGS` is used the copy the coverage.xml of those packages to the plugin path.

**Testing:**
I've tried the parameter generation using #487 in `test_linux_coverage`. Seems to work fine.

* Use case of not using the filtering parameter, same behavior as now:
https://ci.ros2.org/view/All/job/test_linux_coverage/94/cobertura/

* Use case of selecting only a cpp package (rcuitls) 
https://ci.ros2.org/view/All/job/test_linux_coverage/76/cobertura/

* Use case of selection cpp package + python (rcutils + domain_coordinator)
https://ci.ros2.org/view/All/job/test_linux_coverage/93/cobertura/

* Use case of nightly build (default params)
https://ci.ros2.org/job/test_linux_coverage/101/cobertura

**Addedums:**
I've used these changes to be sure that we are reporting the effects of integration testing on the coverage rate.
rclcpp case:
 * rclcpp ([no system test](https://ci.ros2.org/view/All/job/test_linux_coverage/82/cobertura/))  69%
 * rclcpp ([+test_cli + test_communications](https://ci.ros2.org/view/All/job/test_linux_coverage/83/cobertura/)) 76%

With this PR we potentially can remove the coverage calculator script since the report should be doing the same calculus than the calculator, I run the following test: checkout rate from rcutils from a nightly build and check coverage rate of custom report:

```
./get_coverage_ros2_pkg.py https://ci.ros2.org/view/nightly/job/nightly_linux_coverage/901 rcutils
Getting ros2.repos sources to get packages source paths. Please wait
 * src.ros2.rcutils.include.rcutils [100.00] -- 8/8
 * src.ros2.rcutils.include.rcutils.types [100.00] -- 5/5
 * src.ros2.rcutils.src [94.47] -- 1709/1809

Combined unit testing for rcutils: 94.51% 1722/1822
```
https://ci.ros2.org/view/All/job/test_linux_coverage/79/cobertura/
Lines: 95% | 1724/1822


job_templates/snippet/publisher_cobertura.xml.em
```diff
@@ -1,5 +1,5 @@
     <hudson.plugins.cobertura.CoberturaPublisher plugin="cobertura@@1.14">
-      <coberturaReportFile>ws/build*/**/*coverage.xml</coberturaReportFile>
+      <coberturaReportFile>ws/build/*coverage.xml</coberturaReportFile>
```
dirk-thomas, line 2: These looks incorrect. Why is the `*` removed from the `build` directory name?

job_templates/ci_job.xml.em
```diff
@@ -337,6 +341,9 @@ if "!CI_COMPILE_WITH_CLANG!" == "true" (
 )
 if "!CI_ENABLE_COVERAGE!" == "true" (
   set "CI_ARGS=!CI_ARGS! --coverage"
+  if "!CI_COVERAGE_FILTER_PKGS!" NEQ "None" (
```
dirk-thomas, line 344: The declared default value of the parameter is an empty string, not `None`.

ros2_batch_job/__main__.py
```diff
@@ -14,10 +14,11 @@
 
 import argparse
 import configparser
+import glob
 import os
 from pathlib import Path
 import platform
-from shutil import which
+from shutil import which, copyfile
```
dirk-thomas, line 21: Nitpick: one line per import (as per style guide).

ros2_batch_job/__main__.py
```diff
@@ -257,9 +261,15 @@ def get_args(sysargv=None):
         argv, test_args = extract_argument_group(argv, '--test-args')
     else:
         build_args, test_args = extract_argument_group(build_args, '--test-args')
+    if '--coverage-filter-packages' in argv:
+        argv, coverage_filter_packages = extract_argument_group(argv, '--coverage-filter-packages')
+    else:
+        coverage_filter_packages = None
```
dirk-thomas, line 267: I would expect this to require some special handling like line 263 depending on the order of the various greedy arguments `--build-args`, `--test-args` and now `--coverage-filter-packages`.

ros2_batch_job/__main__.py
```diff
@@ -408,7 +473,7 @@ def build_and_test(args, job):
     info("colcon test-result returned: '{0}'".format(ret_test_results))
     print('# END SUBSECTION')
     if args.coverage and args.os == 'linux':
-        process_coverage(args, job)
+        process_coverage(args, job, args.coverage_filter_packages)
```
dirk-thomas, line 476: Since `args` are already passed why separately pass `args.coverage_filter_packages`?

ros2_batch_job/__main__.py
```diff
@@ -273,7 +283,56 @@ def get_args(sysargv=None):
...
+    # no arguments means report all packages in buildspace
+    root_directory_to_report = args.buildspace
+    if package:
+        root_directory_to_report = os.path.join(root_directory_to_report, package)
+
+    coverage_xml_files = glob.glob(
+        os.path.join(root_directory_to_report, '**', 'coverage.xml'),
+        recursive=True)
+    for coverage_xml_path in coverage_xml_files:
+        # python coverage detected: move the coverage.xml file to buildspace to be reported
```
dirk-thomas, line 296: Nitpick: while the comment says "move" the code actually copies.

ros2_batch_job/__main__.py
```diff
@@ -273,7 +283,56 @@ def get_args(sysargv=None):
...
+        dest_file_to_report = os.path.join(args.buildspace, coverage_xml_path.replace(os.path.sep, '.'))
```
dirk-thomas, line 297: Can you share an example name for `dest_file_to_report`? Or point to a test build with these change to look at the resulting workspace.

ros2_batch_job/__main__.py
```diff
@@ -273,7 +283,56 @@ def get_args(sysargv=None):
...
+        include_python_coverage_in_report(args, package_name)
+        # collect paths to run lcov in order to process C/C++ coverage information
+        cmd = [
+            args.colcon_script,
+            'list',
+            '--paths-only',
+            '--base-paths', args.sourcespace,
+            '--packages-select', package_name]
+        print(cmd)
+        src_path = subprocess.check_output(cmd).decode('ascii').strip()
```
dirk-thomas, line 315: Depending on the number of packages it looks inefficient to invoke a subprocess for each. Why not invoke one command with all desired package names?

job_templates/snippet/publisher_cobertura.xml.em
j-rivero: > These looks incorrect. Why is the `*` removed from the `build` directory name?

is there any option in CI which modified the [ws/build/ in the workspace](https://ci.ros2.org/view/All/job/test_linux_coverage/ws/ws/) or other reason to keep the wildcard on the build directory name? Not sure why was there before.

dirk-thomas: Yes, there is a reason to keep the wildcard: https://github.com/ros2/ci/blob/dea332718366fe745f4eabab735bb0db9d65a450/ros2_batch_job/__main__.py#L187-L190

job_templates/ci_job.xml.em
j-rivero: d165461e071611edad33e77dcca824f14bbcae6c

job_templates/snippet/publisher_cobertura.xml.em
j-rivero: Thanks for the context: reverted aafe41e53164f49cd7b06abec318bbeb723ceb6f

ros2_batch_job/__main__.py
j-rivero: 92723966a77d6293be53ef60718ed7246dea947e

ros2_batch_job/__main__.py
j-rivero: ec4acd89415d644be33c34457bf09ea1f019d754

ros2_batch_job/__main__.py
j-rivero: You are right. Removed in 158f17449c98a746191768783c737c2635940b68

ros2_batch_job/__main__.py
j-rivero: My expectation is to have file paths like: `build/my_python_foo_pkg.subdir1.coverage.xml` . The goal was to make them unique and store all them in a single dir (build/) so the cobertura plugin does not need to look into the whole filesystem to find them. Let me launch a testing run to demonstrate that this is actually true,

ros2_batch_job/__main__.py
j-rivero: Excuse me if the answer is simple but, why if '--coverage-filter-packages' is not in argv do we need to extract any argument group? BTW I have no clue why is that done in 263 for test/build which probably explains why I can not answer your question.

dirk-thomas:
> found problems to run testing builds using this PR. It probably needs to be synced with master. @dirk-thomas do you prefer a rebase or a merge commit?

Either is fine with me.

ros2_batch_job/__main__.py
dirk-thomas: > why if '--coverage-filter-packages' is not in argv do we need to extract any argument group?

I wasn't referring to this specific line rather than to the logic to extract `--coverage-filter-packages`.

> I have no clue why is that done in 263 for test/build which probably explains why I can not answer your question.

Both existing options `--build-args` and `--test-args` are greedy and are extracted using `extract_argument_group`. Since their order is unknown the above logic makes sure that it works independent of which of the two are specified first.

With a third option the combinations in which order they appear is getting more complex (2 different orders for 2 options -> 6 different orders for 3 options).

j-rivero: I added a bit more of logic in c1c28a2 although I'm not so sure that we cover all scenarios.

ros2_batch_job/__main__.py
j-rivero: Run `colcon list` just once ccb0ee8

ros2_batch_job/__main__.py
j-rivero: This is one the files generated https://ci.ros2.org/job/test_linux_coverage/ws/ws/build/build.domain_coordinator.coverage.xml. The patch is visible in the URL.

j-rivero:
Test new changes in: [Build Status](https://ci.ros2.org/job/test_linux_coverage/137/)

ros2_batch_job/__main__.py
dirk-thomas: I don't see how the referenced commit addressed the problem?

ros2_batch_job/__main__.py
dirk-thomas: The referenced resource isn't available anymore.

ros2_batch_job/__main__.py
```diff
@@ -276,6 +291,51 @@ def get_args(sysargv=None):
...
+    cmd = [
+        args.colcon_script,
+        'list',
+        '--paths-only',
+        '--base-paths', args.sourcespace,
+        '--packages-select'] + args.coverage_filter_packages
+    print(cmd)
+    src_paths_collection = subprocess.check_output(cmd).decode('ascii').strip().splitlines()
+    if len(src_paths_collection) != len(args.coverage_filter_packages):
+            print('One of the packages to report coverage rate for was not found')
```
dirk-thomas, line 326: Error messages should go to `stderr`. It might also be helpful to include which package names were missing in the message.

clalancette:
@j-rivero Friendly ping on this PR; what's the current status?

j-rivero:
> @j-rivero Friendly ping on this PR; what's the current status?

Sorry Chris for the delay. The status is that the last comment of Dirk needs to be addressed and keep going with the review. I don't have time allocated to finish this anymore and other priorities appeared so either I have time back to work on this or other person needs to continue the work. If the later, I available to help understanding the changes.

mabelzhang:
Backlogging per @chapulina

j-rivero:
Not active anymore.

PR #425, opened 2020-04-09
Use chef configuration in windows dockerfile setup

brawner:
This PR enables the windows dockerfile to install the ROS 2 dependencies from a chef cookbook. This will allow regular testing of the cookbook so that everyone can make use of it.

Currently this targets my personal repo, but I'll switch it over once the ros2 cookbook has a permanent home.

Depends on #424
New commit: https://github.com/ros2/ci/commit/ca34ec30a9b1c75f4bab4b4beaa19843823f5ed6

PR #490, opened 2020-06-30
Install openssl 1.0.2 on windows to use with connext

mikaelarguedas:
Redo of https://github.com/ros2/ci/pull/421 as requested in https://github.com/ros2/ci/pull/421#issuecomment-649861033

This installs openssl 1.0.2 alongside openssl 1.1.1.

Relies on https://github.com/ros2/system_tests/pull/439
First commit is unrelated to this change but was useful to reduce CI time, it can be removed from this PR if not sensible 

Debug
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11314/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6554/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9279/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11213/)

Release:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6577/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9303/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11237/)

### DIFF
--- a/windows_docker_resources/Dockerfile.foxy
+++ b/windows_docker_resources/Dockerfile.foxy
@@ -20,7 +20,8 @@ FROM mcr.microsoft.com/windows:$WINDOWS_RELEASE_VERSION
 
 # These are versioned files, so they shouldn't invalidate caches. Renaming to fixed names.
 # OpenSSL
-ADD https://slproweb.com/download/Win64OpenSSL-1_1_1g.exe C:\TEMP\Win64OpenSSL.exe
+ADD https://slproweb.com/download/Win64OpenSSL-1_0_2u.exe C:\TEMP\Win64OpenSSL1_0_2.exe
+ADD https://slproweb.com/download/Win64OpenSSL-1_1_1g.exe C:\TEMP\Win64OpenSSL1_1_1.exe
 
 # OpenCV
 ADD https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip C:\TEMP\opencv.zip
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
This PR introduces further conflicts with #425 which I'm hesitant to merge.

Regarding the removal of opensplice as a dependency I would rather see that change made across all platforms rather than "snuck in" to an unrelated PR. I don't mind if it stays in the meantime while you iterate.

How confident are you that both OpenSSL installations are non-conflicting and being used? The 1.1.1g binaries default to installing in a separate path from the 1.0.2 binaries *unless* there was an existing installation of 1.0.2 previously in which case I've seen them use the same install paths.

If there are log excerpts from the test builds which show the correct OpenSSL version being used in both scenarios can you please add them to the PR body.

mikaelarguedas:
> Regarding the removal of opensplice as a dependency I would rather see that change made across all platforms rather than "snuck in" to an unrelated PR. I don't mind if it stays in the meantime while you iterate.

:+1:

> How confident are you that both OpenSSL installations are non-conflicting and being used? The 1.1.1g binaries default to installing in a separate path from the 1.0.2 binaries unless there was an existing installation of 1.0.2 previously in which case I've seen them use the same install paths.

Not very, this is why I was originally attempting to use the ones provided by RTI where we had a good handle on the installation location. Unfortunately those never seem to have worked on the windows job configuration.
TBH the only things I'm confident with are the fact that when adding new install location to the paths the connext jobs start passing and that https://github.com/ros2/system_tests/pull/439 fixes cases of mixing openssl versions on master (e.g. Fast-RTPS tests on MacOS currently use openssl 1.0.2 instead of openssl 1.1.1g and would be fixed with that PR).
There is a chance that these installs are conflicting somehow. And currently we don't know from the tests which one is being used.

mikaelarguedas:
I'm not sure how to go about this, I'll go ahead and close this PR.
I'm not familiar with chef, how would I go about adding that to the chef cookbook ? Just add another "openssl_version" entry in ci/install_ros2_foxy.json ?
The chef cookbook seems to reference openssl 1.0.2u but the PR you linked is using 1.0.2n, which one should I use?

PR #493, opened 2020-07-09
Use older debug wheels when building Dashing or Eloquent.

nuclearsandwich:
The Python Debug interpreter on Windows is not ABI compatible with the standard interpreter so wheels using python C extensions need to be compiled with the Debug interpreter.

When we updated the Python version for Foxy we also built new wheels, but we need to continue using the old wheels for Dashing and Eloquent.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -522,14 +522,24 @@ def run(args, build_function, blacklisted_package_names=None):
         pip_packages = list(pip_dependencies)
         if sys.platform == 'win32':
             if args.cmake_build_type == 'Debug':
-                pip_packages += [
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Windows None Rolling [Build Status](https://ci.ros2.org/job/ci_windows/11303/)
Windows Debug Rolling [Build Status](https://ci.ros2.org/job/ci_windows/11304/)
Windows None Dashing  [Build Status](https://ci.ros2.org/job/ci_windows/11305/)
Windows Debug Dashing [Build Status](https://ci.ros2.org/job/ci_windows/11306/)
Windows Debug Eloquent [Build Status](https://ci.ros2.org/job/ci_windows/11307/)

PR #496, opened 2020-07-20
Update cookbook usage to handle multiple rosdistros.

nuclearsandwich:
This update should be applied to #425 before merging it.

With this change the differences in configuration and installed software are delegated to the chef cookbook which means there's no longer a reason for rosdistro-specific dockerfiles.

The different json files set the ros_distro attribute appropriately and change the run-list to install additional rmw providers if they are supported by the distro.


nuclearsandwich:
> Has this been tested or should I run it once it's merged into the target branch?

I've run a handful of test builds to make sure I got this right during development. These just build up to rcutils and so aren't comprehensive but they test that chef runs to completion. We'll want to do some full CI builds of the final RC PR.

Foxy [Build Status](https://ci.ros2.org/job/test_ci_windows/171/)
Dashing [Build Status](https://ci.ros2.org/job/test_ci_windows/173/)

PR #497, opened 2020-07-23
Remove the pytest.ini workaround for distros after Foxy.

clalancette:
After Foxy was released, we added pytest.ini to all of the
packages that required it.  This was so that local tests behaved
the same as CI.  In Foxy and prior, we still need the workaround
so CI will succeed.

Still in draft since I need to run tests on it.  @j-rivero FYI, this relates to (but does not completely remove the need for) #471.  This also relates to https://github.com/ros2/ros2/issues/951

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -359,7 +359,7 @@ def build_and_test(args, job):
     # In Foxy and prior, xunit2 format is needed to make Jenkins xunit plugin 2.x happy
     # After Foxy, we introduced per-package changes to make local builds and CI
     # builds act the same.
-    if args.ros_distro in ['dashing', 'eloquent', 'foxy']:
+    if args.ros_distro in ('dashing', 'eloquent', 'foxy'):
         with open('pytest.ini', 'w') as ini_file:
             ini_file.write('[pytest]\njunit_family=xunit2')
         # check if packages have a pytest.ini file and add the xunit2
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -356,24 +356,27 @@ def build_and_test(args, job):
 
     print('# BEGIN SUBSECTION: test')
 
-    # xunit2 format is needed to make Jenkins xunit plugin 2.x happy
-    with open('pytest.ini', 'w') as ini_file:
-        ini_file.write('[pytest]\njunit_family=xunit2')
-    # check if packages have a pytest.ini file and add the xunit2
-    # format if it is not present
-    for path in Path('.').rglob('pytest.ini'):
-        config = configparser.ConfigParser()
-        config.read(str(path))
-        try:
-            # only if xunit2 is set continue the loop with the file unpatched
-            if config.get('pytest', 'junit_family') == 'xunit2':
-                continue
-        except configparser.NoOptionError:
-            pass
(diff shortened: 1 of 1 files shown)


clalancette:
CI with distro set to Rolling: https://ci.ros2.org/job/ci_linux/11601/
CI with distro set to Foxy: https://ci.ros2.org/job/ci_linux/11602/

Both are expected to succeed, but the Foxy one should have some extra output at the beginning of the test section that says:

```
xunit2 patch applied to ...
```

While the rolling one should not.

ros2_batch_job/__main__.py
```diff
@@ -356,24 +356,27 @@ def build_and_test(args, job):
...
-        except configparser.NoOptionError:
-            pass
-        print('xunit2 patch applied to ' + str(path))
-        config.set('pytest', 'junit_family', 'xunit2')
-        with open(path, 'w+') as configfile:
-            config.write(configfile)
+    # In Foxy and prior, xunit2 format is needed to make Jenkins xunit plugin 2.x happy
+    # After Foxy, we introduced per-package changes to make local builds and CI
+    # builds act the same.
+    if args.ros_distro in ['dashing', 'eloquent', 'foxy']:
```
dirk-thomas, line 362: Nitpick: using `(...)` instead of `[...]` when not needing a mutable list.

clalancette: Good call, 0f6b507.

j-rivero:
> Still in draft since I need to run tests on it. @j-rivero FYI, this relates to (but does not completely remove the need for) #471. This also relates to [ros2/ros2#951](https://github.com/ros2/ros2/issues/951)

+1 although I keep the warning I wrote here https://github.com/ros2/ci/pull/471#issuecomment-646679805

clalancette:
Rolling CI failed because we are still on a version before my fix: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/185 .  I'll talk to the ros2_tracing people and see if we can get an update.

clalancette:
ros2_tracing is now on master, so we can try again.  Here's another CI run:

CI with distro set to Rolling: [Build Status](http://ci.ros2.org/job/ci_linux/11744/)
CI with distro set to Foxy: [Build Status](http://ci.ros2.org/job/ci_linux/11605/)

clalancette:
> +1 although I keep the warning I wrote here [#471 (comment)](https://github.com/ros2/ci/pull/471#issuecomment-646679805)

Right, it is definitely still the case that if *any* package is added that doesn't have the proper pytest.ini, it will cause the build to fail.  While that isn't exactly nice, the good news is that it is very noisy and it should be obvious why it failed.  So I still think we should go forward with this.

I'm going to go ahead and do a full CI run on all platforms just to make sure there aren't any surprises with Rolling.  After that, I'll probably go ahead and merge unless @j-rivero has any other concerns with this PR.

clalancette:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11756/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6833/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9591/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11699/)

PR #498, opened 2020-08-06
Do not exit if pytest version check fails

jacobperron:
Fixes failing CI builds for dashing, eloquent, and foxy.

Addresses https://github.com/ros2/ci/pull/471#discussion_r466079652

Before: [Build Status](https://ci.ros2.org/job/ci_linux/11784/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/11786/)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -371,7 +371,8 @@ def build_and_test(args, job):
             'import pytest;'
             'import sys;'
             'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-            "'"])
+            "'"],
+            exit_on_error=False)
         for path in Path('.').rglob('pytest.ini'):
             config = configparser.ConfigParser()
             config.read(str(path))


nuclearsandwich:
@j-rivero can you comment on your intention here? If I had to guess I'd say that exit_on_error was added without realizing it would halt the build since what you're trying to capture is whether pytest >=6.0.0 is present not assert that it is. If that's the case then this PR looks like it exhibits the originally intended behavior.

j-rivero:
> @j-rivero can you comment on your intention here? If I had to guess I'd say that exit_on_error was added without realizing it would halt the build since what you're trying to capture is whether pytest >=6.0.0 is present not assert that it is. If that's the case then this PR looks like it exhibits the originally intended behavior.

That was the intention originally, yes. Not sure why it was not catch during my initial tests in pull request #471. Thanks @jacobperron, @nuclearsandwich  for the fix.

ros2_batch_job/__main__.py
```diff
@@ -371,7 +371,8 @@ def build_and_test(args, job):
             'import pytest;'
             'import sys;'
             'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-            "'"])
+            "'"],
+            exit_on_error=False)
```
dirk-thomas, line 375: Using the return value for both the version comparison result as well as potential errors looks fragile to me. I would suggest to refactor this logic to:
* either print the version comparison result (and interpret the output) and ensure that the command has a return code of zero
* or just print `pytest.__version__` and perform the comparison in the calling Python code rather than within the invoked command.

jacobperron:
This patch does not work as it appears the result `job.run` is always `True`. I should have tested with a package that involves a `pytest.ini` file, my mistake.

j-rivero:
> This patch does not work as it appears the result `job.run` is always `True`. I should have tested with a package that involves a `pytest.ini` file, my mistake.

I'm testing a fix for the problem, based on Dirk's suggestion: https://github.com/ros2/ci/compare/master...shell_pytest6

PR #499, opened 2020-08-06
Remove dead jobs from ci.ros2.org

clalancette:
Neither the armhf jobs nor the turtlebot jobs are currently viable.  Remove both of them in this pull request.

I've verified that this doesn't actually change any job configurations on https://ci.ros2.org by running `create_jenkins_jobs.py` in dry-run mode.  Once this is approved and merged, I'll manually delete the armhf and turtlebot jobs from the farm, and then they will not be created again.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -107,7 +107,6 @@ def main(argv=None):
         'compile_with_clang_default': 'false',
         'enable_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
-        'turtlebot_demo': False,
         'build_timeout_mins': 0,
         'ubuntu_distro': 'focal',
         'ros_distro': 'rolling',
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
I'm still using CI to produce armhf archives for Dashing ([patch release 7](https://github.com/ros2/ros2/releases/tag/release-dashing-20200722)). I'd prefer not to remove armhf-support from the internals and the manual jobs. I recall discussing this change offline but I must have misinterpreted the full extend of the proposed change.

clalancette:
> I'm still using CI to produce armhf archives for Dashing ([patch release 7](https://github.com/ros2/ros2/releases/tag/release-dashing-20200722)). I'd prefer not to remove armhf support from the internals and the manual jobs. I recall discussing this change offline but I must have misinterpreted the full extend of the proposed change.

Oh,I forgot we were still supporting it for Dashing.  Good call.

In that case, I'll revert most of that and just remove the nightly armhf jobs.  That would leave all other armhf jobs available.  Does that satisfy your needs?

nuclearsandwich:
> In that case, I'll revert most of that and just remove the nightly armhf jobs. That would leave all other armhf jobs available. Does that satisfy your needs?

Yes I just need the manual  ci_ (which will preserve the test_ci_ job as well) and the manual packaging job but not the nightlies or nightly packaging job.

clalancette:
All right, I've rewritten it to just get rid of the nightly jobs.  The following ones will still be available:

```
Skipped 'ci_linux-armhf' because the config is the same (dry run)
Skipped 'test_ci_linux-armhf' because the config is the same (dry run)
Skipped 'ci_packaging_linux-armhf' because the config is the same (dry run)
Skipped 'packaging_linux-armhf' because the config is the same (dry run)
```

create_jenkins_job.py
```diff
@@ -264,15 +263,16 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
-            'time_trigger_spec': PERIODIC_JOB_SPEC,
-            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
-        })
+        if os_name != 'linux-armhf':
+            job_name = 'nightly_' + job_os_name + '_debug'
+            if os_name == 'windows':
+                job_name = job_name[:15]
+            create_job(os_name, job_name, 'ci_job.xml.em', {
+                'cmake_build_type': 'Debug',
+                'disabled': False,
```
dirk-thomas, line 272: I would guess this line isn't necessary anymore.

Same below.

clalancette: Good point.  Removed in ea4fd7c.

clalancette:
Thanks, merging.  I'll deploy next (though that is mostly a no-op).  Then I'll manually delete the dead jobs.

PR #500, opened 2020-08-07
Add a command-line option to show more context lines for diff.

clalancette:
Sometimes the default diff doesn't give enough context on what
changed in a job.  This adds a command-line paramter to allow
the user to increase the number of lines of context.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -46,6 +46,14 @@ template_prefix_path[:] = \
     [os.path.join(os.path.abspath(os.path.dirname(__file__)), 'job_templates')]
 
 
+def nonnegative_int(inval):
+    ret = int(inval)
+    if ret < 0:
+        # The error message here gets completely swallowed by argparse
+        raise ValueError('Value must be positive or zero')
+    return ret
+
+
 def main(argv=None):
     if argv is None:
         argv = sys.argv[1:]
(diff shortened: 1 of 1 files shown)


create_jenkins_job.py
```diff
@@ -46,6 +46,14 @@
     [os.path.join(os.path.abspath(os.path.dirname(__file__)), 'job_templates')]
 
 
+def nonnegative_int(inval):
+    ret = int(inval)
+    if ret < 0:
+        # The error message here gets completely swallowed by argparse
```
dirk-thomas, line 52: This comment should make you suspicious that something is wrong. Please check the `argparse` [docs](https://docs.python.org/3.8/library/argparse.html#type) about `type` and `ArgumentTypeError`.

create_jenkins_job.py
```diff
@@ -46,6 +46,14 @@
     [os.path.join(os.path.abspath(os.path.dirname(__file__)), 'job_templates')]
 
 
+def nonnegative_int(inval):
+    ret = int(inval)
```
dirk-thomas, line 50: You probably want to catch the possible exception here too.

create_jenkins_job.py
clalancette: Fixed in f13827d

create_jenkins_job.py
clalancette: Fixed in f13827d

clalancette:
Thanks, merging.

PR #501, opened 2020-08-07
Simplify the logic run for pytest version inside the venv

j-rivero:
Mostly inspired by this comment of @dirk-thomas https://github.com/ros2/ci/pull/498#discussion_r466516070. Should help to ix the regressions found in #498.

Implementation: print the pytest version inside the venv using job.run and move the version check logic outside of the venv.

Tested with `ament_package` on Foxy in:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11817/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6888/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9644/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11751/)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -372,7 +372,7 @@ def build_and_test(args, job):
         job.run(["%s -c 'import pytest; print(pytest.__version__)' > %s" % (job.python, pytest_6_version_file)])
         pytest_6_or_greater = False
         with open(pytest_6_version_file, 'r') as current_version_file:
-            pytest_6_or_greater = StrictVersion(str(current_version_file.read()) >= StrictVersion("6.0.0"))
+            pytest_6_or_greater = StrictVersion(current_version_file.read()) >= StrictVersion("6.0.0")
 
         for path in Path('.').rglob('pytest.ini'):
             config = configparser.ConfigParser()
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -372,7 +372,7 @@ def build_and_test(args, job):
         job.run(["%s -c 'import pytest; print(pytest.__version__)' > %s" % (job.python, pytest_6_version_file)])
         pytest_6_or_greater = False
         with open(pytest_6_version_file, 'r') as current_version_file:
-            pytest_6_or_greater = StrictVersion(current_version_file.read() >= StrictVersion("6.0.0"))
+            pytest_6_or_greater = StrictVersion(str(current_version_file.read()) >= StrictVersion("6.0.0"))
 
         for path in Path('.').rglob('pytest.ini'):
             config = configparser.ConfigParser()
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -14,6 +14,7 @@
 
 import argparse
 import configparser
+from distutils.version import StrictVersion
 import os
 from pathlib import Path
 import platform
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -365,14 +366,14 @@ def build_and_test(args, job):
...
-            'import pytest;'
-            'import sys;'
-            'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-            "'"],
-            exit_on_error=False)
+        # Note: workaround using a file as output since job BatchLinux class does not have
+        # check_output capabilities
+        pytest_6_version_file = os.path.join(args.buildspace, 'pytest6_version_result.txt')
+        job.run(["%s -c 'import pytest; print(pytest.__version__)' > %s" % (job.python, pytest_6_version_file)])
+        pytest_6_or_greater = False
```
dirk-thomas, line 373: This line is unnecessary.

j-rivero: if we define the variable only inside the `with` clause, it could be used outside of that scope?

dirk-thomas: Yes.

j-rivero:
Please do not merge, using foxy ros2 repos file is failing:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11820/)

Looking into this now.

j-rivero:
This is confusing me: the test display to be using pytest-6.0.1, I've verified that this information is being generated fine in this PR and goes to the temporary file. Some lines after this message appear `PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0. See:` like if it was not using pytest6. Or do I am missing something obvious?

```
[1m============================= test session starts ==============================[0m
platform linux -- Python 3.8.2, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
cachedir: /home/jenkins-agent/workspace/ci_linux/ws/build/ament_package/.pytest_cache
rootdir: /home/jenkins-agent/workspace/ci_linux/ws/src/ament/ament_package, configfile: pytest.ini
plugins: colcon-core-0.6.0, mock-3.2.0, repeat-0.8.0, rerunfailures-9.0, cov-2.10.0
[1mcollecting ... [0m[1m
collected 1 item                                                               [0m

test/test_flake8.py [32m.[0m[33m                                                    [100%][0m

[33m=============================== warnings summary ===============================[0m
/home/jenkins-agent/workspace/ci_linux/venv/lib/python3.8/site-packages/_pytest/junitxml.py:446
  /home/jenkins-agent/workspace/ci_linux/venv/lib/python3.8/site-packages/_pytest/junitxml.py:446: PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0. See:
    https://docs.pytest.org/en/stable/deprecations.html#junit-family-default-value-change-to-xunit2
  for more information.
    _issue_warning_captured(deprecated.JUNIT_XML_DEFAULT_FAMILY, config.hook, 2)
```

j-rivero:
> Or do I am missing something obvious?

Don't have more time to investigate what's going on with the pytest enviroment, an alternative to these changes that seems not to fix the problem is pr #502

jacobperron:
It does seem odd. Locally, I can reproduce the test warning with pytest 6.0.1 building ament_package `foxy` branch.
Maybe we need to make the patch regardless?

jacobperron:
After some reading and experimentation, I think we should expect the deprecation warnings to appear if `junit_family` is not explicitly set (in the pytest.ini or otherwise). 

The problem here is that we're assuming if `junit_family` is set to "legacy", then we don't have to patch the pytest.ini file; however, we'll still get a deprecation warning if it is not set at all.

Unless there's some reason we shouldn't always set `junit_unit=xunit2` for Dashing, Eloquent, and Foxy, I think something like #502 is a cleaner solution. Furthermore, I don't think we need to patch the pytest.ini file, and can instead set junit_family from the command-line (see https://github.com/ros2/ci/commit/c6b5dc77d5f0dfdfd00ee37f68491e68c7f42145)

PR #502, opened 2020-08-07
Do not check for pytest6 and patch pytest.ini unconditionally

j-rivero:
An alternative to #501

Given the problems with pytest version detailed in https://github.com/ros2/ci/pull/501#issuecomment-670668797, this PR removes all the logic to calculate the pytest version and just patch unconditionally. Note that the scope of this patching is limited to Dashing, Eloquent and Foxy after #497.

Branch test for the `ament_package`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11828/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6895/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9651/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11758/)

Some test done in previous releases:
 * Dashing [Build Status](https://ci.ros2.org/job/ci_linux/11823/)
 * Dashing without this changes, same UNSTABLE [Build Status](https://ci.ros2.org/job/ci_linux/11824/) 
 * Eloquent [Build Status](https://ci.ros2.org/job/ci_linux/11825/)

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -365,25 +365,11 @@ def build_and_test(args, job):
             ini_file.write('[pytest]\njunit_family=xunit2')
         # check if packages have a pytest.ini file that doesn't choose junit_family=xunit2
         # and patch configuration if needed to force the xunit2 value
-        pytest_6_or_greater = job.run([
-            '"%s"' % job.python, '-c', "'"
-            'from distutils.version import StrictVersion;'
-            'import pytest;'
-            'import sys;'
-            'sys.exit(StrictVersion(pytest.__version__) >= StrictVersion("6.0.0"))'
-            "'"],
-            exit_on_error=False)
         for path in Path('.').rglob('pytest.ini'):
             config = configparser.ConfigParser()
             config.read(str(path))
-            if pytest_6_or_greater:
-                # only need to correct explicit legacy option if exists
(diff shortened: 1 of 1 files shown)


j-rivero:
> I'm okay with this patch. If we want to go forward with my patch in #503, then it looks like we'll need to make changes to the CMake packages adding pytests for Foxy (see [#503 (comment)](https://github.com/ros2/ci/pull/503#issuecomment-671471092)). @j-rivero Let me know if you want to go forward with #503 instead and I can help make changes to the affected packages.

I really like the approach in #503  so my impression is that is going to be the final fix. Not sure if you need this one meantime all the pieces are in place, if that is the case feel free to merge. if that is not the case, please close this one. Thanks @jacobperron

jacobperron:
> I really like the approach in #503 so my impression is that is going to be the final fix

Unfortunately, I don't think #503 will work unless we backport the pytest.ini changes for CMake packages to Foxy, Eloquent, and Dashing. To avoid a bunch of backport PRs, I think we should just merge this one.

jacobperron:
Let me try backporting changes to Foxy and see if that's enough for #503

jacobperron:
I've merged https://github.com/ros2/ci/pull/503.

PR #504, opened 2020-08-14
Update qtaccount.ini used for Windows builds.

nuclearsandwich:
A second commit will be added here updating the ros2-cookbooks repo when https://github.com/ros-infrastructure/ros2-cookbooks/pull/11 lands to avoid showing this token in CI builds.


nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_windows/11903/)

PR #508, opened 2020-08-27
Fast-forward gazebo_ros_pkgs branch

jacobperron:
This PR reapplies commits for `gazebo_ros_pkgs` on recent changes to the `master` branch.
I noticed this was required to fix Linux CI.

Before: [Build Status](https://ci.ros2.org/job/ci_linux/12063/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/12064/)

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -63,11 +63,13 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-yaml \
   uncrustify \
   libgazebo9-dev \
-  gazebo9
+  gazebo9 \
+  libboost-all-dev
 
 # Install and self update pip/setuptools to the latest version.
 RUN apt-get update && apt-get install --no-install-recommends -y python3-pip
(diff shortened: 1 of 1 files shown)


linux_docker_resources/Dockerfile
```diff
@@ -61,23 +61,31 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-pyparsing \
   python3-pytest-mock \
   python3-yaml \
-  uncrustify
+  uncrustify \
+  libgazebo9-dev \
```
chapulina, line 65: Note that Noetic and Foxy should use `gazebo11`

jacobperron: Do you think it's worth maintaining a ros2/ci branch for Gazebo 9? I guess we can rely on the build.ros2.org CI jobs for released distros (like Dashing and Eloquent).

chapulina: Good point, we only need 11 at this point :+1:

PR #522, opened 2020-10-26
embed the Jenkins BUILD_URL into the archive

dirk-thomas:
To be able to determine which build number is being used in an unpacked archive.

Linux: https://ci.ros2.org/job/ci_packaging_linux/424/ which has a file `BUILD_URL` in the root of the archive pointing to the Jenkins build which produced the artifact.


jacobperron (approved):
LGTM

PR #528, opened 2020-11-13
Explain where to find the markdown snippet from ci_launcher jobs.

nuclearsandwich:
A small addition to the CI builder docs since I realized we never explain what to actually do once CI is triggered. More than this might be desired but I figured I'd start with this.


clalancette (approved):
Oh, good call.  That is useful information that people need.

PR #531, opened 2021-01-08
Add Python CFFI as dependency

sloretz:
Use packages for Linux
Use pip for OSX and Windows

For ros2/rclpy#656

@cottsay @iantheengineer @azeey FYI


nuclearsandwich:
We would need a debug wheel for cffi on Windows but we are already installing it as a dependency for `cryptography`.

sloretz:
Closing since rclpy went the `pybind11` route instead of the CFFI route

PR #534, opened 2021-01-13
Don't COLCON_IGNORE fastcdr, we want to use it for rosbag2 type conversion

emersonknapp:
@nuclearsandwich let me know what you think about this, we currently have a dependency from rosbag2 converters on rmw_fastrtps_cpp, but only need the CDR implementation and are looking to pull a dependency on it directly. Noticed that in the packaging builds it is ignored if "USE_FASTRTPS_*" is not checked. We're thinking this library should be usable as a dependency even if the middleware itself is not being used.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -715,7 +715,6 @@ def run(args, build_function, blacklisted_package_names=None):
             ]
         if 'rmw_fastrtps_cpp' in args.ignore_rmw and 'rmw_fastrtps_dynamic_cpp' in args.ignore_rmw:
             blacklisted_package_names += [
-                'fastcdr',
                 'fastrtps',
                 'fastrtps_cmake_module',
                 'rmw_fastrtps_shared_cpp',


clalancette (approved):
That makes sense to me, but I'll let @nuclearsandwich review before merging.

Karsten1987 (approved):
nice

nuclearsandwich (approved):

clalancette:
@emersonknapp Can you please run a CI with this change in place, just to ensure that nominal things still work?

emersonknapp:
> @emersonknapp Can you please run a CI with this change in place, just to ensure that nominal things still work?

Given that I do not have write access to this repo and can't push my feature branch to it, I'm not sure that I am able to do so. My understanding is that  `CI_SCRIPTS_BRANCH` parameter is used on builds to achieve this, but that the branch has to be present on this repository, not a development fork.

nuclearsandwich:
@emersonknapp I pushed your branch to `ros2/ci` directly so it can be used with CI_SCRIPTS_BRANCH. The branch name is unchanged (`emersonknapp/no-ignore-cdr`)

emersonknapp:
Thanks! Anything I should run beyond:

* ci_packaging_linux: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/447/)

?

clalancette:
I guess the other thing test to run is when both Fast-RTPS static and Fast-RTPS dynamic are unchecked.  I don't expect problems, really (we're just adding a new package in that case), but we may as well check it.

emersonknapp:
Of course, makes sense
* [Build Status](https://ci.ros2.org/job/ci_packaging_linux/448/)

PR #535, opened 2021-01-19
Use offline installer and skip qt-maintenance

nuclearsandwich:
In its current form this PR is just a test to see if not running the maintenance tool is sufficient to get a successful build.


nuclearsandwich:
It took a day of iteration but we're now working around the downtime of download.qt.io with a mirrored offline installer of qt 5.12.10 up to qt_cpp_gui was built [Build Status](https://ci.ros2.org/job/ci_windows/13511/) and succeeded.

Once download.qt.io is back we can figure out where to go from here.

nuclearsandwich:
I'm going to yolomerge this in order to give us a chance at nightlies. Worst case scenario something else is broken.

PR #538, opened 2021-01-26
Return to using qt online installer by default.

nuclearsandwich:
This change in tandem with refactored qt5 recipes will allow us to use
either the qt5 online installer with the latest version of qt5 on
Windows or the 5.12 offline installer. This is designed to be used
whenever the Qt5 online installer is unavailable. Since the offline
installer stopped being made available for open source builds after
5.12 it is not recommended unless the online installer is unusable.

### DIFF
--- a/windows_docker_resources/install_ros2_rolling.json
+++ b/windows_docker_resources/install_ros2_rolling.json
@@ -2,6 +2,9 @@
   "ros2_windows": {
     "download_sources": "false",
     "vs_version": "buildtools",
+    "qt5": {
+      "installation_method": "offline"
+    },
     "ros2_ws": "C:/ci",
     "ros_distro": "rolling",
     "rti_connext": {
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Checking the online installer: [Build Status](https://ci.ros2.org/job/ci_windows/13564/)

nuclearsandwich:
Having tested the offline installer logic I reverted the commit which switched to it. Before merging here we should get reviews on the cookbook PR so that the last commit on this PR will return us to the `latest` branch there: https://github.com/ros-infrastructure/ros2-cookbooks/pull/24

brawner (approved):

nuclearsandwich:
One last build to make sure that there are no regressions after the merge onto latest [Build Status](https://ci.ros2.org/job/ci_windows/13577/)

emersonknapp:
I'm not positive but I think that since this PR I have not been able to have any successful windows builds e.g. https://ci.ros2.org/job/ci_windows/13592/console

nuclearsandwich:
Investigating now. Due to the nature of docker cache-busting I'm not sure whether a revert will immediately bring CI back online but I'll try that first.

nuclearsandwich:
@emersonknapp the revert has been merged so CI should be unblocked. Thanks for reporting

PR #540, opened 2021-01-28
Revert "Return to using qt online installer by default."

nuclearsandwich:
Reverts ros2/ci#538 due to issues reported https://github.com/ros2/ci/pull/538#issuecomment-768664034

PR #541, opened 2021-01-28
Use latest ros2-cookbooks repository with Qt online installer.

nuclearsandwich:
This is a re-application of #538 which was reverted in #540 due to regressions during the Qt maintenance app tool run.

As of writing I haven't made any changes to the cookbook I'm just opening a fresh pull request to test and report results on. Although #538 updated the submodule it did not update the branch name in .gitmodules which would have meant that the incorrect branch would be fetched when used with the `git submodule update --remote`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,4 +19,4 @@
 [submodule "windows_docker_resources/ros2-cookbooks"]
 	path = windows_docker_resources/ros2-cookbooks
 	url = git@github.com:ros-infrastructure/ros2-cookbooks
-	branch = no-qt-maintenance-app
+	branch = latest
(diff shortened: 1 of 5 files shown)


nuclearsandwich:
There are renewed reports of issues with the online Qt installer so in bd5ba75 I've added attributes to use the offline installer from a known good mirror (currently the file is available from download.qt.io but if they're having availability issues I'd rather park somewhere else while they are ironed out).

nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_windows/13598/)

nuclearsandwich:
Closing in favor of #545

PR #542, opened 2021-02-03
Remove the extra_rmw_release jobs.

clalancette:

As we discussed in the ROS 2 meeting the other day, we probably don't need to keep around the extra_rmw_release jobs anymore.  The only thing they are testing beyond the standard set is rmw_fastrtps_dynamic, which isn't fully supported anyway.  This PR removes the jobs from the configuration.  If it is approved and merged, I will then go in and manually delete the nightly jobs from ci.ros2.org.

@cottsay I know you mentioned that the `extra_rmw_release` jobs are being used as the basis for other things.  But I'm pretty sure that you were talking about on https://build.ros2.org, so these should be safe to delete.  Please let me know if that isn't the case.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -491,21 +491,6 @@ def main(argv=None):
                                      ' --packages-up-to ' + ' '.join(quality_level_pkgs + testing_pkgs_for_quality_level),
             })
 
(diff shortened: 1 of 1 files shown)


cottsay (approved):
> pretty sure that you were talking about on https://build.ros2.org

Correct.

PR #543, opened 2021-02-16
Add shiboken to the Linux apt packages.

clalancette:
This is so that the CI jobs are consistent with what is on
build.ros2.org.


clalancette:
CI (Linux only, since this only affects Linux right now):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13676/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8555/)

linux_docker_resources/Dockerfile-RHEL
```diff
@@ -38,6 +38,8 @@ RUN yum install \
   python36-vcstool \
   python36-virtualenv \
   python-rosdep \
+  shiboken \
+  shiboken-python36-devel \
```
cottsay, line 42: It should probably be noted that these packages are for the Qt4 version of Shiboken (v1), while the packages you're adding to the Ubuntu Dockerfile support Qt5 (Shiboken v2).

At least that's the way I understand it.

clalancette: Yep, you are totally right; RHEL-7 did not use the shiboken generator.

I'm going to change this PR to only do the fix on Ubuntu for now.  We should probably figure out the situation with Shiboken2 on RHEL, because otherwise we'll have the GPL problem there that we discussed earlier.

clalancette:
Closing this one since we decided to stick with SIP.

PR #545, opened 2021-03-04
Update ros2-cookbooks to account for Visual Studio change.

nuclearsandwich:
Update the ros2-windows cookbook to work around the removal of a parameter in the visual studio installer. https://github.com/ros-infrastructure/ros2-cookbooks/pull/28

This PR absorbs #541 since the vs-changes branch is based on latest rather than no-qt-maintenance-app. The offline installer is still used.


clalancette (commented):
Looks good to me!

PR #548, opened 2021-03-12
This is no longer a package that rosbag2 provides.

clalancette:

This should fix the build currently failing at https://ci.ros2.org/view/nightly/job/nightly_linux_coverage/

@emersonknapp FYI, this is a result of ros2/rosbag2#670

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -394,7 +394,6 @@ def main(argv=None):
             'ros2param',
             'ros2topic',
             'rosbag2_compression',
-            'rosbag2_converter_default_plugins',
             'rosbag2_cpp',
             'rosbag2_storage',
             'rosbag2_storage_default_plugins',


clalancette:
I went in and modified a CI job "by hand" to show what would happen after we remove this line: https://ci.ros2.org/job/ci_linux_coverage/213/console

I also did a "dry-run" deployment.  This will remove this package both from `nightly_linux_coverage`, `test_linux_coverage`, `ci_linux_coverage`, and `nightly_linux_foxy_coverage`.  That all seems OK to me, but opinions welcome (particularly about the change to Foxy).

emersonknapp (approved):

clalancette:
I'm going to go ahead and merge this and deploy the changes so we get builds over the weekend.  If anyone has comments or objections, please leave them here and I can address them next week.

emersonknapp:
I was confused to learn that this had hardcoded lists of packages to check, at all.

I am starting to think that something like package.xml needs a field for quality level declaration, so that it is 1) easily machine-parseable (right now it's very difficult to reliably get quality levels in an automated way) and 2) can inform the buildfarm instead of this list

tfoote:
+1 you got to this before me Chris. Related to removing the hardcoded list here. There's definitely value in capturing the quality levels in a machine readable format. However, I think that something like a metapackage/variant definition is the best future evolution for this list. We do still want it to be manually curated. (ala we don't want anyone setting QL1 on their package to automatically included it in our more rigerous batch testing regimes necessarily.)  On the flip side, that sort of declaration would be  a good for enabling more extensive testing in the individual devel jobs etc.

clalancette:
> I am starting to think that something like package.xml needs a field for quality level declaration, so that it is 1) easily machine-parseable (right now it's very difficult to reliably get quality levels in an automated way) and 2) can inform the buildfarm instead of this list

I will say that I wrote a tool that can walk a checked out ROS 2 installation and tell you the quality levels: https://github.com/clalancette/walk-qds .  It essentially expects that all packages have a QUALITY_DECLARATION.md that contains the string `claims to be in the **Quality Level <num>**`.  All of the core package do that, so it works fairly well in the core.

That being said, it would be nice if this were a bit more machine-parseable.  Instead of putting something in the package.xml, I'd really more like to extend REP-2004 to add some required machine-readable data.  That way you don't spread the quality stuff over multiple files, where it is all but guaranteed to get out-of-sync.

PR #547, opened 2021-03-11
Add support for `rmw_connextdds`, and disable `rmw_connext_cpp`

asorbini:
This PR adds support for the new RMW implementation for RTI Connext DDS, `rmw_connextdds`, while also disabling the existing one, `rmw_connext_cpp`.

The changes are being proposed to gradually phase out `rmw_connext_cpp` in favor of `rmw_connextdds` (see [rticommunity/rmw_connextdds#9](https://github.com/rticommunity/rmw_connextdds/issues/9#issuecomment-793002295)).

`rmw_connextdds` will become available in the default `ros2.repos` once [ros2/ros2#1099](https://github.com/ros2/ros2/pull/1099) is merged.

PR #549, opened 2021-03-16
Add support for `rmw_connextdds`, and disable `rmw_connext_cpp` (#547)

asorbini:
This PR includes changes from branch `asorbini/rmw_connextdds` which switch the default RMW implementation for RTI Connext DDS from `rmw_connext_cpp` to `rmw_connextdds`.

The changes include:
   - Add a new variable to `ci_launcher` to enable `rmw_connextdds` in a plan (`CI_USE_CONNEXTDDS`, disabled by default).
   - Add `rmw_connextdds` to the "ignored RMWs" if `CI_USE_CONNEXTDDS` is false.
   - Automatically disable `rmw_connextdds` for `armhf`, and `aarch64` builds.
   - Remove variables `CI_USE_CONNEXT_STATIC`, and `CI_USE_CONNEXT_DYNAMIC`.
   - Always add `rmw_connext_cpp`, and `rmw_connext_dynamic_cpp` to the list of "ignored RMWs".
   - Remove some references to (deprecated) `rmw_opensplice_cpp`.

Merging of this PR should be coordinated with two other PRs which will enable `rmw_connextdds` in package `rmw_implementation`, and in the default `ros2.repos`:

- [ros2/rmw_implementation#182](https://github.com/ros2/rmw_implementation/pull/182)
- [ros2/ros2#1099](https://github.com/ros2/ros2/pull/1099)

See [rticommunity/rmw_connextdds#9](https://github.com/rticommunity/rmw_connextdds/issues/9) for a general discussion about the inclusion of `rmw_connextdds` in ROS2 Rolling (and phasing out of `rmw_connext_cpp`).

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -139,12 +139,12 @@ def main(argv=None):
         'linux-aarch64': {
             'label_expression': 'linux_aarch64',
             'shell_type': 'Shell',
-            'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'},
+            'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_connextdds'},
         },
         'linux-armhf': {
             'label_expression': 'linux_armhf',
             'shell_type': 'Shell',
-            'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'},
+            'ignore_rmw_default': data['ignore_rmw_default'] | {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_connextdds'},
             'build_args_default': data['build_args_default'].replace(
                 '--cmake-args', '--cmake-args -DCMAKE_CXX_FLAGS=-Wno-psabi -DCMAKE_C_FLAGS=-Wno-psabi -DDISABLE_SANITIZERS=ON'),
         },
@@ -160,7 +160,7 @@ def main(argv=None):
     os_config_overrides = {
         'linux-rhel': {
             'mixed_overlay_pkgs': '',
-            'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp'},
+            'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp', 'rmw_connextdds'},
             'use_connext_debs_default': 'false',
         },
     }
@@ -218,7 +218,7 @@ def main(argv=None):
         # configure a manual version of the packaging job
         ignore_rmw_default_packaging = {'rmw_opensplice_cpp'}
         if os_name in ['linux-aarch64', 'linux-armhf']:
-            ignore_rmw_default_packaging |= {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp'}
+            ignore_rmw_default_packaging |= {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_connextdds'}
         create_job(os_name, 'ci_packaging_' + os_name, 'packaging_job.xml.em', {
             'build_discard': {
                 'days_to_keep': 180,
@@ -327,7 +327,8 @@ def main(argv=None):
         # packages of ros2.repos that are used by the qualitly level packages during
         # tests.
 
-        # out of the list since ignored by colcon: shape_msgs, stereo_msgs, rmw_connext, rmw_cyclonedds
+        # out of the list since ignored by colcon: shape_msgs, stereo_msgs,
+        # rmw_connext, rmw_connextdds, rmw_cyclonedds.
         quality_level_pkgs = [
             'action_msgs',
             'ament_index_cpp',
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -89,8 +89,7 @@
         <script plugin="script-security@@1.70">
           <script><![CDATA[build.setDescription("""\
 branch: ${build.buildVariableResolver.resolve('CI_BRANCH_TO_TEST')}, <br/>
-use_connext_static: ${build.buildVariableResolver.resolve('CI_USE_CONNEXT_STATIC')}, <br/>
-@# use_connext_dynamic: ${build.buildVariableResolver.resolve('CI_USE_CONNEXT_DYNAMIC')}, <br/>
+use_connextdds: ${build.buildVariableResolver.resolve('CI_USE_CONNEXTDDS')}, <br/>
 use_connext_debs: ${build.buildVariableResolver.resolve('CI_USE_CONNEXT_DEBS')}, <br/>
 use_cyclonedds: ${build.buildVariableResolver.resolve('CI_USE_CYCLONEDDS')}, <br/>
 use_fastrtps_static: ${build.buildVariableResolver.resolve('CI_USE_FASTRTPS_STATIC')}, <br/>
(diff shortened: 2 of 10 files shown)


clalancette (changes requested):
So the major problem with this as-is is that the same jobs are used for Rolling, Dashing, and Foxy.  In the case of Dashing and Foxy, we need to keep the default Connext rmw as `rmw_connext_cpp`, while for Rolling we want to change it to `rmw_connextdds`.

The additional complication is that we use the same job for all of them, so there can be only one default (we distinguish them at runtime by a drop-down which has the name of the release in it).  That piece of information leads to the probable solution: we have only a single checkbox for "Use Connext", and at runtime we select the correct RMW based on the name of the release (`rmw_connext_cpp` for 'dashing' or 'foxy', `rmw_connextdds` for anything else).

@asorbini Let me know if this makes sense, or if you need more information to change this PR to meet that.

ivanpauno:
Another detail: we sometimes run a job only selecting one rmw implementation, but the problem is that `rosidl_typesupport_fastrtps` is ignored if `CI_USE_FASTRTPS_STATIC` box isn't checked (see [here](https://github.com/ros2/ci/blob/28988874df9cb55d8aa65a28406bf77534450f11/ros2_batch_job/__main__.py#L709-L710)), and that will cause the job to fail if only the rmw connextdds box was checked.
The `rosidl_typesupport_fastrtps_*` packages should only be ignored when neither rmw fastrtps or connext boxes were selected.

`rosidl_typesupport_fastrtps_*` also requires `fastrtps`, though it could actually depend only on `fastcdr` (there seems to be an unneded find_package making it required, but fastrtps is never used in code).
So until that is fixed, these [two ignores](https://github.com/ros2/ci/blob/28988874df9cb55d8aa65a28406bf77534450f11/ros2_batch_job/__main__.py#L718-L719) will also be a problem.

PS: This only matters if we want the jobs to continue passing with only the connext box checked, if we're fine with remembering that also the fastrtps box has to be checked it doesn't matter.

asorbini:
Started some CI jobs to validate the latest changes:

- [ci_linux#13955](https://ci.ros2.org/job/ci_linux/13955/) (rolling, all)
- [ci_linux#13956](https://ci.ros2.org/job/ci_linux/13955/) (foxy, up to `rmw_connext_cpp`)
- [ci_osx#11658](https://ci.ros2.org/job/ci_osx/11658) (rolling, all)
- [ci_windows#14040](https://ci.ros2.org/job/ci_windows/14040/) (rolling)
- [ci_windows#14041](https://ci.ros2.org/job/ci_windows/14041/) (dashing, up to `rmw_connext_cpp`)

clalancette:
@asorbini Where are we with this?  We're starting to get awfully close to the Galactic freeze deadline.

asorbini:
Hi @clalancette, as I said in my last comment, I don't have a way to validate this changes on ci.ros2.org and I was asking for your input. Do you have any suggestions on how I should proceed?

I can only think of trying to set up a local Jenkins instance, but I'm not sure that's going to be sufficient testing, since I'd only be able to test on Linux.

asorbini:
I'm trying to set up a local Jenkins instance on my machine, but it's a whole new process and I'm having some issues, so it might take me some time. I also tried to determine which version of Jenkins you are running on ci.ros2.org, and which plugins you have installed, but my user doesn't seem to have sufficient credentials for that (or maybe I just couldn't find the info).

I did find a problem in my changes to `create_jenkins_job.py` which I [committed to `asorbini:asorbini/rmw_connextdds`](https://github.com/asorbini/ci/commit/14954f986a155456faf58e2b9893087d3dc37b5d). Before opening another PR to merge that change here, I'll see if I can validate things further and squash any remaining issues.

Looking forward to any input you may have on how to speed this whole process up and move forward with integration of `rmw_connextdds`.

asorbini:
I've got as far as having the jenkins plans created on my local instance (this allowed to fix a couple more errors in my changes), and I'm almost able to also run them.

Unfortunately my plans fail to execute because they fail to clone repository `osrf/qtaccount` as a submodule, to which my GitHub user doesn't have access.

If you can grant me access to this repository, I might be able to continue and actually complete the work to reproduce the CI infrastructure locally.

In the mean time I will create another PR to merge the most recent changes from my branch `asorbini/rmw_connextdds`.

job_templates/ci_job.xml.em
```diff
@@ -132,12 +131,18 @@ fi
...
-export CI_ARGS="$CI_ARGS --ignore-rmw"
-if [ "$CI_USE_CONNEXT_STATIC" = "false" ]; then
+export CI_ARGS="$CI_ARGS --ignore-rmw rmw_connext_dynamic_cpp"
+# TODO(asorbini) `rmw_connext_cpp` is still the default for dashing" and foxy.
+if [ -n "$CI_ROS_DISTRO" -a \( "$CI_ROS_DISTRO" = dashing -o "$CI_ROS_DISTRO" = foxy \) ]; then
+  if [ "$CI_USE_CONNEXTDDS" = "false" ]; then
+    export CI_ARGS="$CI_ARGS rmw_connext_cpp"
+  fi
+else
+  # Always ignore `rmw_connext_cpp` in favor of `rmw_connextdds` for older releases.
```
clalancette, line 141: Shouldn't this comment say:

```
# Always ignore `rmw_connext_cpp` in favor of `rmw_connextdds` for newer releases.
```

?  I'm just making sure I understand what is going on here.

(there are several other instances of this comment in the code below)

clalancette (changes requested):
I've got one clarification for now.  I'll do another review pass over this later, but I want to start some CI jobs first.

clalancette:
I'm going to run several different CI setups (besides the usual) to see what happens.

Linux CI using this branch, but *not* adding rmw_connextdds to the ros2.repos file.  This may be an unexpected situation for this PR, so it is OK if it fails:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13984/)

create_jenkins_job.py
```diff
@@ -216,9 +216,9 @@ def create_job(os_name, job_name, template_file, additional_dict):
             packaging_label_expression = 'macos &amp;&amp; mojave'
 
         # configure a manual version of the packaging job
-        ignore_rmw_default_packaging = {'rmw_opensplice_cpp'}
```
clalancette, line 219: As we discussed privately, we shouldn't remove this yet.  Once dashing goes out of support, we can remove all of the Opensplice support together.

clalancette (changes requested):

create_jenkins_job.py
asorbini: Fixed in #552

job_templates/ci_job.xml.em
asorbini: Fixed in #552

asorbini: (Your analysis was correct, my bad for that typo)

create_jenkins_job.py
```diff
@@ -216,9 +216,9 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
+        ignore_rmw_default_packaging = { 'rmw_opensplice_cpp' }
```
asorbini, line 219: I accidentally introduced spaces between the braces. If needed I can fix it.

clalancette: I just went ahead and fixed it, no worries.

create_jenkins_job.py
```diff
@@ -160,7 +160,7 @@ def main(argv=None):
     os_config_overrides = {
         'linux-rhel': {
             'mixed_overlay_pkgs': '',
-            'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp'},
+            'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_connextdds'},
```
clalancette, line 163: This is technically removing opensplice, but since we are never going to target Dashing with RHEL, this is fine.

clalancette (approved):
This looks OK to me.  I'd appreciate a review from either @cottsay or @nuclearsandwich just to double-check.

create_jenkins_job.py
asorbini: Oops, missed this one, sorry.

nuclearsandwich (commented):
I have one comment on an outstanding conversation but otherwise I think this looks as reasonable as something written in templated shell and batch can look. @asorbini I really appreciate you digging in and iterating on this one.

I'm going to do some pre-deployment checks and one of the suggestions for RHEL should be selected before the big :heavy_check_mark: but we should be able to wrap this up today.

create_jenkins_job.py
asorbini: I don't have a strong preference and I'm not even sure it would matter (since I'm not a maintainer :) ), but your second suggestion looks like a good way to be consistent with other parts of the script, and to be more flexible going forward by automatically detecting `ignore_rmw_default` from the command line.

I can apply whatever fix you select, although it might be easier to do a "maintainer change" (is that the term?) like @clalancette did earlier since you have push access to the branch on this repo, than me opening another PR to incorporate changes.

clalancette: I'll go make the change to change this back to the old code; that is maximally compatible (if not optimal).

clalancette, line 163:
```suggestion
            'ignore_rmw_default': {'rmw_connext_cpp', 'rmw_connext_dynamic_cpp', 'rmw_opensplice_cpp', 'rmw_connextdds'},
```

clalancette:
All right, one final set of CI before I merge and deploy this:

Rolling/master:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14029/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8850/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14121/)

Foxy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14030/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8851/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14122/)

clalancette:
Rolling aarch64 had one warning, but I'm pretty sure that will go away once I deploy this change.

Foxy Linux (amd64 and aarch64) had some test failures in sros2, but I doubt it is because of this PR.

Foxy macOS had a few test failures, but again, I doubt it is from this PR.

Foxy Windows seems stuck (at least, it has been in the same place for the last hour).  I'm going to go ahead without that and merge this anyway.

PR #554, opened 2021-03-24
Remove connext packages from quality level checking.

clalancette:
They don't have a quality level assigned anyway.

This should fix the overnight job failure of https://ci.ros2.org/view/nightly/job/nightly_linux_coverage/1184/

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -325,7 +325,7 @@ def main(argv=None):
 
         # Proposed list of packages to maximize coverage while testing quality level
         # packages. The list is composed by the list of qualitly level packages plus
-        # packages of ros2.repos that are used by the qualitly level packages during
+        # packages of ros2.repos that are used by the quality level packages during
         # tests.
 
         # out of the list since ignored by colcon: shape_msgs, stereo_msgs,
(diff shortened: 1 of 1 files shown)


clalancette:
ci_linux_coverage job: * Linux [Build Status](http://ci.ros2.org/job/ci_linux_coverage/214/)

create_jenkins_job.py
```diff
@@ -456,8 +454,6 @@ def create_job(os_name, job_name, template_file, additional_dict):
                                      '--packages-up-to ' + ' '.join(quality_level_pkgs + testing_pkgs_for_quality_level),
             })
 
-        # configure nightly coverage job on x86 Linux only
-        if os_name == 'linux':
```
cottsay, line 460: The description doesn't really give any indication of why this part if the change is made. Could you elaborate a little?

clalancette: That's true, good point.  This is a cleanup; while I was in here, I noticed that this condition is repeated twice with no difference between them (that is, this condition is already above).  Because of that, we can just nest this change inside the previous if statement.

It's just a cleanup, so if you think it is a problem I can pull it out of here.

nuclearsandwich: I'd be fine with keeping the change in the PR if it is split into a separate commit and rebase-merged rather than squash merged. Although at that point making two separate PRs isn't that much more effort.

clalancette: I'm just going to drop the change for now, it is not needed for the fix.

clalancette: All right, removed now, ready for another review.

cottsay (approved):

PR #561, opened 2021-04-05
Create a version of the clang+libcxx job intended for manual run

emersonknapp:
Currently there is only `nightly_linux_clang_libcxx`, which doesn't really feel appropriate to use when trying to validate a PR. Add an additionall `ci_linux_clang_libcxx` for this purpose.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -287,15 +287,13 @@ def main(argv=None):
                 'test_args_default': data['test_args_default'] + ' --packages-up-to rcpputils',
             })
 
-        # configure job for compiling with clang+libcxx on linux
+        # configure jobs for compiling with clang+libcxx on linux
         if os_name == 'linux':
             # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
-            clang_libcxx_build_args = data['build_args_default'].replace('--cmake-args',
-                '--cmake-args -DRCL_LOGGING_IMPLEMENTATION=rcl_logging_noop') + \
-                ' --mixin clang-libcxx --packages-skip intra_process_demo'
-            # Only running test from the lowest-level C package to ensure "working" binaries are generated.
+            clang_libcxx_build_args = data['build_args_default'] + ' --mixin clang-libcxx --packages-skip intra_process_demo'
+            # Only running tests from the lowest-level C package to ensure "working" binaries are generated.
             # We do not want to test more than this as we observe issues with the clang libcxx standard library
-            # we don't plan to tackle for now. The important part of this nightly is to make sure the code compiles
+            # we don't plan to tackle for now. The important part of the jobs is to make sure the code compiles
             # without emitting thread-safety related warnings.
             clang_libcxx_test_args = data['test_args_default'].replace(' --retest-until-pass 2', '') + ' --packages-select rcutils'
             # nightly job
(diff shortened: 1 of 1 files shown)


create_jenkins_job.py
```diff
@@ -287,23 +287,32 @@ def create_job(os_name, job_name, template_file, additional_dict):
                 'test_args_default': data['test_args_default'] + ' --packages-up-to rcpputils',
             })
 
-        # configure nightly job for compiling with clang+libcxx on linux
+        # configure job for compiling with clang+libcxx on linux
```
clalancette, line 290:
```suggestion
        # configure jobs for compiling with clang+libcxx on linux
```

create_jenkins_job.py
```diff
@@ -287,23 +287,32 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
         if os_name == 'linux':
             # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
             clang_libcxx_build_args = data['build_args_default'].replace('--cmake-args',
                 '--cmake-args -DRCL_LOGGING_IMPLEMENTATION=rcl_logging_noop') + \
                 ' --mixin clang-libcxx --packages-skip intra_process_demo'
+            # Only running test from the lowest-level C package to ensure "working" binaries are generated.
```
clalancette, line 296:
```suggestion
            # Only running tests from the lowest-level C package to ensure "working" binaries are generated.
```

create_jenkins_job.py
```diff
@@ -287,23 +287,32 @@ def create_job(os_name, job_name, template_file, additional_dict):
...
+            # We do not want to test more than this as we observe issues with the clang libcxx standard library
+            # we don't plan to tackle for now. The important part of this nightly is to make sure the code compiles
```
clalancette, line 298:
```suggestion
            # we don't plan to tackle for now. The important part of the jobs is to make sure the code compiles
```

create_jenkins_job.py
```diff
@@ -287,23 +287,32 @@ def create_job(os_name, job_name, template_file, additional_dict):
                 'test_args_default': data['test_args_default'] + ' --packages-up-to rcpputils',
             })
 
-        # configure nightly job for compiling with clang+libcxx on linux
+        # configure job for compiling with clang+libcxx on linux
         if os_name == 'linux':
             # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
             clang_libcxx_build_args = data['build_args_default'].replace('--cmake-args',
                 '--cmake-args -DRCL_LOGGING_IMPLEMENTATION=rcl_logging_noop') + \
```
clalancette, line 294: While we are in here, we might consider removing forcing the logging implementation to 'noop'.  I know for sure that it works with spdlog.  I know it isn't strictly related to this PR, but I still think it is worth doing.

clalancette (changes requested):
Overall, the idea looks fine to me.  I've got some comments inline.

clalancette (approved):
Looks good to me.  To test this out, I'm going to run a "dry" deploy of this change to CI.  Assuming that looks like I expect, I'm going to merge and deploy this for real.  Since there isn't really a good way to test the rest of the changes before all of that (without running the nightly job, which I'm reluctant to do), I'll then immediately run a ci_linux_clang_libcxx job.

clalancette:
Dry deployment looked good to me, merging.  Thanks for the change.

PR #562, opened 2021-04-07
Iceoryx ci

clalancette:


clalancette:
This is needed to build iceoryx, which I accidentally merged before this in https://github.com/ros2/ros2/pull/1119 .  Since I ran CI with this in https://github.com/ros2/ros2/pull/1119#issuecomment-814374170 , and since this code was actually written by @Karsten1987 , I'm going to go ahead and self-approve and merge this.  If there is any follow-up, please leave comments here and I'll take care of it.

PR #563, opened 2021-04-07
Suppress warnings on RHEL-8 from Eigen.

clalancette:
There is a bug in the version of Eigen in RHEL-8 that makes
it emit a lot of warnings anytime the header file is included.
Just suppress the warnings for tf2_eigen in CI, since they
aren't in our code.

This should suppress all of the warnings in https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_release/739/gcc/


clalancette:
RHEL: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/118/)

cottsay (approved):

clalancette:
The single CMake warning in the test build is there in the nightlies, so is not because of this PR.  Going ahead and merging this one.

PR #565, opened 2021-04-08
Disable more warnings for RHEL-only.

clalancette:
It turns out that rviz_rendering also includes Eigen, so it
also needs the warning suppression flag in RHEL.


clalancette:
RHEL CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/120/)

clalancette:
There's still a CMake warning in `ament_pytest`, but that is unrelated to this PR.

cottsay (approved):

PR #568, opened 2021-04-14
Use offline Qt5 installer.

nuclearsandwich:
Somewhere along the way this attribute got unset or was never properly set after the refactoring of the offline installer support. Because the installer is stuck at 5.12.10 we should not have any further issues updating it but it also means we're stuck building against 5.12.10. @cottsay that may be something that we have to document if we don't return to the online installer for Galactic.

### DIFF
--- a/windows_docker_resources/install_ros2_dashing.json
+++ b/windows_docker_resources/install_ros2_dashing.json
@@ -2,6 +2,9 @@
   "ros2_windows": {
     "download_sources": "false",
     "vs_version": "buildtools",
+    "qt5": {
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_windows/14471/)

nuclearsandwich:
FYI @jacobperron I'm updating the Foxy jobs since I know there's work ongoing for Foxy as well. Likewise for Dashing even though I hvaen't started up the next patch release.

PR #573, opened 2021-05-04
Pin cookbook dependency versions on Windows.

nuclearsandwich:
We're using Chef/Cinc 15 on Windows until some underlying issues can be resolved. One of our cookbook dependencies has had a major version bump which drops Chef 15 support.

Most of the discussion will go in https://github.com/ros-infrastructure/ros2-cookbooks/pull/32


nuclearsandwich:
This issue is currently affecting Foxy builds (likely due to cache misses since the previous foxy builds) so here's a foxy build to test on [Build Status](https://ci.ros2.org/job/ci_windows/14664/)

But we also need to make sure it doesn't regress the rolling / nightly builds so here is a default build [Build Status](https://ci.ros2.org/job/ci_windows/14665/)

azeey (approved):

PR #578, opened 2021-06-01
Update fetched repository key location.

nuclearsandwich:
1. Use GitHub https url to fetch the repository key.
2. Update the date in the cache bust line for the new key.


cottsay (approved):
Any reason not to use the Dockerfile `ADD` command directly with the URL? It would eliminate the need to bust the cache manually.

This is fine as-is, though.

nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/456/)

jacobperron (approved):

PR #579, opened 2021-06-03
Fix reference before assignment error

jacobperron:
If we run without `--do-venv` then we get the following runtime error:

    UnboundLocalError: local variable 'venv_path' referenced before assignment

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -604,7 +604,7 @@ def run(args, build_function, blacklisted_package_names=None):
                 ['colcon/%s' % name for name in colcon_packages],
                 shell=True)
 
-        if sys.platform != 'win32':
+        if args.do_venv and sys.platform != 'win32':
             colcon_script = os.path.join(venv_path, 'bin', 'colcon')
         else:
             colcon_script = which('colcon')


clalancette (approved):
Seems reasonable to me.  Can you run a short job with and without `--do-venv` just to confirm that this works in both cases?

jacobperron:
I'm not sure how to trigger CI without `--do-venv`, it's set in the job template here:
https://github.com/ros2/ci/blob/41c09163181696a12c9545eb3404e67b6d074f30/job_templates/packaging_job.xml.em#L141

Maybe we don't need to trigger job for this case since there's no way to configure it on our infrastructure. I'll trigger CI for the nominal use-case once we're unblocked by https://github.com/ros2/ci/pull/578

nuclearsandwich (approved):
I'm good merging without a job assuming this was verified locally (since from what I know about this PR's origins it was used to try and reproduce CI behavior locally).

jacobperron:
Packaging CI: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/460/)

clalancette:
> I'm good merging without a job assuming this was verified locally (since from what I know about this PR's origins it was used to try and reproduce CI behavior locally).

Yeah, same.  It is clearly a bug, I was just trying to get verification.  I'd say go ahead and merge this one with a successful packaging job and two approvals.

PR #580, opened 2021-06-18
Remove Dashing references from CI scripts.

nuclearsandwich:
Now that Dashing is no longer supported we don't need to maintain CI
script support for it.


clalancette (commented):
Should we also remove all references to Opensplice, since Dashing is the last distro that supported it?  Or do you want to do that separately?

nuclearsandwich:
> Should we also remove all references to Opensplice, since Dashing is the last distro that supported it? Or do you want to do that separately?

We should and I'll do it in a separate PR in case for some wild reason we need to resurrect opensplice not Dashing.

clalancette (approved):

PR #582, opened 2021-06-21
Remove Opensplice configuration.

nuclearsandwich:
Dashing was the last ROS 2 distribution that supported Opensplice.

Now that it has been retired we no longer need Opensplice configuration.

Ideally this will rebase onto the Dashing removal in #580 to make sure that cross-references between Dashing and Opensplice are all caught correctly.


clalancette (approved):
Looks good.  Should we also remove opensplice.rb from https://github.com/ros-infrastructure/ros2-cookbooks/tree/latest/cookbooks/ros2_windows/recipes ?

PR #583, opened 2021-06-21
Use 'system' installation of mypy on Linux.

nuclearsandwich:
This draft PR will need a rebase once #581 merges.

When discussing the mypy stubs issue offline, there was a base assumption that we were already using the version of mypy packaged in Ubuntu 20.04 and that using that across all platforms (as opposed to for example the last version of mypy with bundled stubs) was the most reasonable place to pin the dependency. However we're not currently using the system mypy in CI. It's only use is in our ROS 2 apt packages and individuals following our development environment setup instructions will shadow that package with the pip package for their own development.

As demonstrated by https://github.com/ros2/ci/pull/581#issuecomment-864314822 the system version of mypy on Ubuntu works at least as well as the pip-installed version. So the question is:

Should we use the system version where it is available. This PR "assumes" it is available on Linux platforms since our container images for [RHEL](https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile-RHEL) and [Ubuntu](https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile#L57) install mypy system-wide.
When we update our CI to support preliminary Ubuntu 22.04 images this would allow us to immediately start testing with the packaged version of mypy in 22.04 as opposed to the current setup which would require us to remember to switch it. We'd still have to manually update the pin on other platforms.


nuclearsandwich:
>  consistency between our various build/install methods is more important than being on the latest. Thus, I would suggest that we use system versions as much as possible, and use pinning for pip-installed dependencies as needed.

This is definitely where I'm at as well. It will also make it easier for us to eventually run a consistent set of linter versions for each stable ROS distribution rather than needing to ignore or skip linter tests when running CI for stable releases which have drifted from the latest linter results.



> What do other people think?

@ros2/team any other input?

clalancette:
We actually ended up doing this elsewhere, so I'm going to close this out.

PR #596, opened 2021-10-01
Update ros2-cookbooks to address xmllint download failure.

nuclearsandwich:
Incorporating https://github.com/ros-infrastructure/ros2-cookbooks/pull/37

Before merging the cookbook PR should be merged and this PR updated to restore the latest branch of it.


nuclearsandwich:
This change does not get us back to operational as invalidating the docker cache triggers additional downloads which we cannot complete for the same reason. The solution will have to involve using an OpenSSL version which is compatible with the current LetsEncrypt root cert.

PR #597, opened 2021-10-01
Upgrade to chef 16 and work around expired root certificate issue.

nuclearsandwich:
There are three changes here:
1. Use cinc 16 instead of cinc 15 for the main configuration run.
2. Apply a vendored `cacert.pem` file to work around the [expiration of a Let's Encrypt root CA](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/).
3. Switching to chef-workstation chocolatey package for Berkshelf installation

These changes could be de-coupled but they were co-developed and once Cinc 16.16.11 or later is available on Windows we can remove our own workaround so if we get to 16 a little early and without incident I'm not bothered.

Since this PR makes changes to early Dockerfile layers we performed some operations that have been cached for quite some time. The chefdk chocolatey package has been deprecated like chefdk itself and the package is now a proxy package for chef-workstation. As a result of installing chef-workstation rather than chefdk the location of the berks binary has moved. (It appears that there are "unstable" builds of cinc workstation for windows as of 0.17.


nuclearsandwich:
[Build Status](https://ci.ros2.org/job/test_ci_windows/223/)

cottsay:
Rebuild after fixing the test arguments: [Build Status](https://ci.ros2.org/job/test_ci_windows/224/)

cottsay (approved):

PR #598, opened 2021-10-21
Add jammy as an Ubuntu distro choice.

nuclearsandwich:
Adds Jammy as an Ubuntu distro option.

There's no corresponding ROS 1 distro for Jammy so right now I've set that to empty. Running ros1_bridge builds on Jammy may require the installation of ROS packages from Ubuntu.

One thing to keep an eye on is that Jammy forward will use the ROS 2 repository for bootstrap packages such as vcstool, which would make it easier for a transition from fixed dependencies to rosdep to unintentionally pull in ROS 2 binaries.
Since there is currently no rosdep support for ci.ros2.org and adding it isn't planned. I'm not too concerned by this but it is a challenge I've faced bringing Space ROS CI up on build.ros2.org.

This is currently deployed to test_ci_linux and https://ci.ros2.org/job/test_ci_linux/55/ is running with `image_tools` and `intra_process_demo` ignored due to https://bugs.launchpad.net/ubuntu/+source/gdal/+bug/1949362

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -35,7 +35,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
-RUN apt-get update && apt-get install --no-install-recommends -y python3-lark-parser python3-opencv
+RUN if test ${UBUNTU_DISTRO} = bionic; then LARK_PKG=python3-lark-parser; else LARK_PKG=python3-lark; fi && apt-get update && apt-get install --no-install-recommends -y $LARK_PKG python3-opencv
 
 # Install build and test dependencies of ROS 2 packages.
 RUN apt-get update && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
> Does this maintain the current default, or is the "top" value automatically the default?

The default is still `focal` and is configured in the job creation data. The default ubuntu_distro for our jobs is set here: https://github.com/ros2/ci/blob/f85ac5581d177854f9b97c486b04c667e7adcf67/create_jenkins_job.py#L111 with a few overrides for specific jobs elsewhere in the file.

linux_docker_resources/Dockerfile
```diff
@@ -34,7 +35,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
-RUN apt-get update && apt-get install --no-install-recommends -y python3-lark-parser python3-opencv
+RUN if test ${UBUNTU_DISTRO} = bionic; then LARK_PKG=python3-lark-parser; else LARK_PKG=python3-lark; fi && apt-get update && apt-get install --no-install-recommends -y $LARK_PKG python3-opencv
```
clalancette, line 38: Just as a note for later, I think we can remove `bionic` support at this point.  We have no releases using it.

clalancette (approved):

PR #605, opened 2021-12-22
Remove armhf support.

clalancette:
We have no platforms that currently use it, and the possibility
of adding it back anytime soon seems remote.

Assuming people are OK with this in theory, I'll run some test jobs to see what happens.


nuclearsandwich (approved):
Good night sweet 32-bit prince

clalancette:
I did a test deployment with this, and there are no changes to the existing jobs (other than the RHEL jobs, which are out-of-date for reasons unrelated to this change).  So I'm going to go ahead and merge this and then deploy.

PR #611, opened 2022-01-24
Update launcher label expression for built-in node.

nuclearsandwich:
Jenkins is updating the terminology for the node built into the Jenkins
server [1]. This label expression should make the launcher job, which is
the only job in ros2/ci which runs on the built-in node,
forward-compatible with the updated name once we update Jenkins to a
version which uses the updated naming and continue working in the
meantime. Once the name-change is sufficiently far back we can simplify
the expression again by removing the old agent name.

[1]: https://www.jenkins.io/doc/book/managing/built-in-node-migration/


clalancette (approved):

nuclearsandwich:
Deployed last night when merged and running smoothly on the new `built-in` node.

PR #614, opened 2022-01-28
Use RTI Connext 6.0.1 on Jammy and when building Rolling on Windows

nuclearsandwich:
Installs RTI Connext 6.0.1 via debs on Linux.
* I updated the entry_point.sh script to source the rtisetenv script since there's "guessing" logic in rti_connext_dds_cmake_module which is looking for 5.3.1 and does not find 6.0.1.
* I'm not yet sure whether the .debs are able to be used for DDS security tests. [Build Status](https://ci.ros2.org/job/ci_linux/16028/) should give us a hint. If they can, I'd propose that we make that the default and only installation method for 6.0.1 on Linux in order to remove some CI complexity.

For Windows I updated the private submodule to ship both 5.3.1 and 6.0.1 and updated the cookbook parameters to use 6.0.1 in Rolling. A build for that is being tested [Build Status](https://ci.ros2.org/job/ci_windows/16354/)

Once we're through these initial builds I'll start full CI runs to verify that this doesn't break anything above it.

### DIFF
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -33,6 +33,12 @@ if [ "${ARCH}" != "aarch64" ]; then
         case "${CI_ARGS}" in
           *--connext-debs*)
             echo "Using Debian package of Connext"
+            if test -r /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh; then
+                echo "Sourcing RTI setenv script /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh"
+                . /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh
+            elif test -r /opt/rti.com/rti_connext_dds-5.3.1/resource/scripts/rtisetenv_x64Linux3gcc5.4.0.bash
+                echo "rti_connextdds_cmake_module will guess the location of Connext 5.3.1 so don't source anything."
+            fi
             ;;
           *)
             echo "Installing Connext binaries off RTI website..."
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -36,7 +36,7 @@ if [ "${ARCH}" != "aarch64" ]; then
             if test -r /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh; then
                 echo "Sourcing RTI setenv script /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh"
                 . /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh
-            elif test -r /opt/rti.com/rti_connext_dds-5.3.1/resource/scripts/rtisetenv_x64Linux3gcc5.4.0.bash
+            elif test -r /opt/rti.com/rti_connext_dds-5.3.1/resource/scripts/rtisetenv_x64Linux3gcc5.4.0.bash; then
                 echo "rti_connextdds_cmake_module will guess the location of Connext 5.3.1 so don't source anything."
             fi
             ;;
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,7 +8,7 @@
 [submodule "windows_docker_resources/rticonnextdds-src"]
 	path = windows_docker_resources/rticonnextdds-src
 	url = git@github.com:osrf/rticonnextdds-src.git
-	branch = binaries/windows/amd64/5.3.1
+	branch = binaries/windows/amd64/omnibus
 [submodule "windows_docker_resources/rticonnextdds-license"]
 	path = windows_docker_resources/rticonnextdds-license
 	url = git@github.com:osrf/rticonnextdds-src.git
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -78,7 +78,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y lcov
 RUN pip3 install -U lcov_cobertura_fix
 
 # Install the Connext binary from the OSRF repositories.
-RUN if test \( ${PLATFORM} = x86 -a ${INSTALL_CONNEXT_DEBS} = true \); then apt-get update && RTI_NC_LICENSE_ACCEPTED=yes apt-get install -y rti-connext-dds-5.3.1; fi
+RUN if test \( ${PLATFORM} = x86 -a ${INSTALL_CONNEXT_DEBS} = true -a ${UBUNTU_DISTRO} != jammy \); then apt-get update && RTI_NC_LICENSE_ACCEPTED=yes apt-get install -y rti-connext-dds-5.3.1; fi
+RUN if test \( ${PLATFORM} = x86 -a ${INSTALL_CONNEXT_DEBS} = true -a ${UBUNTU_DISTRO} = jammy \); then apt-get update && RTI_NC_LICENSE_ACCEPTED=yes apt-get install -y rti-connext-dds-6.0.1; fi
 
 # Install the RTI dependencies.
 RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-install-recommends -y default-jre-headless; fi
--- a/windows_docker_resources/install_ros2_rolling.json
+++ b/windows_docker_resources/install_ros2_rolling.json
@@ -12,9 +12,9 @@
       "min_vs_version": "2017",
       "license_file": "C:/TEMP/rticonnextdds-license/rti_license.dat",
       "installer_dir": "C:/TEMP/rticonnextdds-src",
-      "version": "5.3.1",
+      "version": "6.0.1",
       "edition": "pro",
-      "openssl_version": "1.0.2n"
+      "openssl_version": "1.1.1k"
     }
   },
   "run_list": [


linux_docker_resources/entry_point.sh
```diff
@@ -33,6 +33,12 @@ if [ "${ARCH}" != "aarch64" ]; then
         case "${CI_ARGS}" in
           *--connext-debs*)
             echo "Using Debian package of Connext"
+            if test -r /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh; then
+                echo "Sourcing RTI setenv script /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh"
+                . /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh
+            elif test -r /opt/rti.com/rti_connext_dds-5.3.1/resource/scripts/rtisetenv_x64Linux3gcc5.4.0.bash; then
+                echo "rti_connextdds_cmake_module will guess the location of Connext 5.3.1 so don't source anything."
```
clalancette, line 40: Out of curiousity, is it actively harmful to source the 5.3.1 setup script?  If not, I think it would be more consistent to just source it here (even if `rti_connextdds_cmake_module` may redo some work).

clalancette (commented):
One minor comment inline, but the code otherwise looks decent to me.

CI seems to be unhappy, though :(.

linux_docker_resources/entry_point.sh
nuclearsandwich: Probably not. I was aiming for a minimal change. I think there's a fair bit of refactoring we could do to simplify the plumbing here if we take a step back and look.

nuclearsandwich: Oh! one minor speed bump is that we can't source the 5.3.1 script from a POSIX sh script without `sed`ing it first as we do in the release repository for connext_cmake_module (and which hasn't been ported to rti_connext_cmake_module due to the guessing behavior)

nuclearsandwich:
> * I'm not yet sure whether the .debs are able to be used for DDS security tests. [Build Status](https://ci.ros2.org/job/ci_linux/16028/) should give us a hint.

Based on the test output it looks like the answer is no. That's a bummer. I'll have to look at getting the evaluation binaries installed on Linux as well.

nuclearsandwich:
> For Windows I updated the private submodule to ship both 5.3.1 and 6.0.1 and updated the cookbook parameters to use 6.0.1 in Rolling. A build for that is being tested [Build Status](https://ci.ros2.org/job/ci_windows/16354/)



I was missing the step of re-combining the split files. [Build Status](https://ci.ros2.org/job/ci_windows/16358/)

nuclearsandwich:
> > 
> 
> I was missing the step of re-combining the split files. [Build Status](https://ci.ros2.org/job/ci_windows/16358/)

This finished but wasn't able to find Connext. @asorbini @clalancette what are your thoughts on setting NDDSHOME or CONNEXTDDS_DIR via the Chef scripts or as part of the ROS 2 batch job? Also @asorbini is there a preferred/recommended variable between the two?

nuclearsandwich:
Attempt the first to install Connext evaluation binaries. These are no longer tar'd with openssl but it appears to bundle it directly in the .run file. We'll see if there's any additional steps required to get security working.

[Build Status](https://ci.ros2.org/job/ci_linux/16034/)

asorbini:
Thank you for these updates @nuclearsandwich, a few comments on some of the issues:

> I updated the entry_point.sh script to source the rtisetenv script since there's "guessing" logic in rti_connext_dds_cmake_module which is looking for 5.3.1 and does not find 6.0.1.

The "guessing logic" in rti_connext_dds_cmake_module should only focus on coming up with a value for CONNEXTDDS_ARCH, if unspecified. The module calls `find_package(RTIConnextDDS  ...)` with version "5.3.1" but that should only be the "minimum required version". In my experience, I haven't had any problems with finding 6.0.1, since that version number is greater or equal to the specified minimum.

Can you provide some more info on what problems you experienced when trying to load 6.0.1?

rti_connext_dds_cmake_module doesn't really need the `rtisetenv` script to be sourced, all that is needed is setting `CONNEXTDDS_DIR` and `CONNEXTDDS_ARCH`. Since those variables are basically encoded in the `rtisetenv` script's path, the call to source the script could be replaced with an `export`, e.g.:

```sh
# instead of:
 if test -r /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh; then
                echo "Sourcing RTI setenv script /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh"
                . /opt/rti.com/rti_connext_dds-6.0.1/resource/scripts/rtisetenv_x64Linux4gcc7.3.0.sh
elif test -r /opt/rti.com/rti_connext_dds-5.3.1/resource/scripts/rtisetenv_x64Linux3gcc5.4.0.bash; then
                echo "rti_connextdds_cmake_module will guess the location of Connext 5.3.1 so don't source anything."

# we could do:
if test -d /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0; then
    export CONNEXTDDS_DIR=/opt/rti.com/rti_connext_dds-6.0.1 \
           CONNEXTDDS_ARCH=x64Linux4gcc7.3.0
if test -d /opt/rti.com/rti_connext_dds-5.3.1/lib/x64Linux3gcc5.4.0; then
    export CONNEXTDDS_DIR=/opt/rti.com/rti_connext_dds-5.3.1 \
           CONNEXTDDS_ARCH=x64Linux3gcc5.4.0

```

The `rtisetenv` also sets LD_LIBRARY_PATH, but that's not needed to load the libraries and build `rmw_connextdds`. At runtime, `rti_connext_dds_cmake_module` installs a shell hook which will update that variable on its own.

> I'm not yet sure whether the .debs are able to be used for DDS security tests. Build Status should give us a hint. If they can, I'd propose that we make that the default and only installation method for 6.0.1 on Linux in order to remove some CI complexity.

As you found out, the .debs do not contain the Security extensions, only the basic libraries.

> Based on the test output it looks like the answer is no. That's a bummer. I'll have to look at getting the evaluation binaries installed on Linux as well.

Keep in mind that you have also access to the non-eval bundles (for both 5.3.1 and 6.0.1) which can be installed in "unattended" mode, without requiring a dedicated `pexpect` python script. I guess they could be provisioned similarly to the approach used for the Windows ones.

> This finished but wasn't able to find Connext. @asorbini @clalancette what are your thoughts on setting NDDSHOME or CONNEXTDDS_DIR via the Chef scripts or as part of the ROS 2 batch job?

That would be the preferred way for me. The `rti_connext_dds_cmake_module` expects either one of those variables to be set by the calling environment and it will not try to guess them, as you found out.

> Also @asorbini is there a preferred/recommended variable between the two?

NDDSHOME is the variable that's been used "historically" by Connext ("ndds" in an earlier life), to identify the installation directory. We have gradually introduced CONNEXTDDS_DIR in recent years as an alternative, and I chose to support both in `rmw_connextdds`/`rti_connext_dds_cmake_module`, with CONNEXTDDS_DIR taking priority over NDDSHOME, in order to more easily support compilation and usage of `rmw_connextdds` + Connext 6 alongside `rmw_connext_cpp` + Connext 5 in the early days of development.

My preference would go to CONNEXTDDS_DIR, since I'm looking forward to a future where NDDSHOME is eventually phased out :)

nuclearsandwich:
> Can you provide some more info on what problems you experienced when trying to load 6.0.1?

I think all my specific issues are discussed upthread. I've been documenting some of the whack-a-mole approach to getting this working through the various places this hooks into our infrastructure but I think with your latest comments I've got paths forward everywhere. I'll keep documenting the whack-a-mole primarily to aid my future self in adding support for additional connext versions going forward and ping you directly if I run into any roadblocks or would like fresh advice. Thanks!



> As you found out, the .debs do not contain the Security extensions, only the basic libraries.

I'll send you an email about this as it complicates some future infrastructure changes that we have in the pipeline. Thanks for confirming it.

> rti_connext_dds_cmake_module doesn't really need the `rtisetenv` script to be sourced, all that is needed is setting `CONNEXTDDS_DIR` and `CONNEXTDDS_ARCH`. Since those variables are basically encoded in the `rtisetenv` script's path, the call to source the script could be replaced with an `export`, e.g.:

> The `rtisetenv` also sets LD_LIBRARY_PATH, but that's not needed to load the libraries and build `rmw_connextdds`. At runtime, `rti_connext_dds_cmake_module` installs a shell hook which will update that variable on its own.

This is good info. We have a couple of different build scenarios we need to handle across our infrastructure and documentation:
* User installation from source on Linux
* User installation from source on Windows
* ci.ros2.org on Windows
* ci.ros2.org on Linux (debs or eval bundles)
* CI jobs on build.ros2.org
* Binary packaging jobs on build.ros2.org

Ideally each of these would work the exact same way so that the procedures we document for end-users match what we verify each night in our infrastructure nightlies. The binary packaging jobs for the both the old [connext_dds_module](https://github.com/ros2/rosidl_typesupport_connext/tree/master/connext_cmake_module) and current rti_connext_dds_cmake_module packages include [patches to support sourcing](https://github.com/ros2-gbp/rosidl_typesupport_connext-release/blob/383defc6ce774f811846ac5a860bb91f90d17eda/debian/rules.em#L29) a [slightly modified for sh compatibility](https://github.com/ros2-gbp/rosidl_typesupport_connext-release/blob/383defc6ce774f811846ac5a860bb91f90d17eda/debian/rules.em#L68-L73) copy of the setenv script in that package.
From their the _cmake_module can include environment hooks to make sure all downstream builds have the needed environment variables set.

Because that's what we're doing for our binary packages I'd like to use the same procedure (sourcing the rti setenv scripts) across all our infra and projects. If there's a preference to change those patches to _just_ set `CONNEXTDDS_DIR` then I would want to propagate that change throughout our infrastructure and installation instructions.

In a pros/cons discussion. I like the transparency of "just setting an environment variable" versus running an executable script that can perform any number of operations but I like the future-proof nature of running a script provided by RTI to make any adjustments to the environs required so that there's a consistent interface that's somewhat immune to platform-specific changes. When adding the inertia of continuing to do what we're already doing I'd probably lean toward continuing to source the setenv script but I could very easily be swayed by additional arguments to make the switch to the environment variables.

nuclearsandwich:
> My preference would go to CONNEXTDDS_DIR, since I'm looking forward to a future where NDDSHOME is eventually phased out :)

This is exactly what I wanted to know. Thanks!

nuclearsandwich:
Attempt to use the unattended installer on Linux [Build Status](https://ci.ros2.org/job/ci_linux/16078/)

nuclearsandwich:
Windows build setting CONNEXTDDS_DIR [Build Status](https://ci.ros2.org/job/ci_windows/16417/)
Linux build using pexpect installer [Build Status](https://ci.ros2.org/job/ci_linux/16081/)

nuclearsandwich:
> Linux build using pexpect installer

I was not necessarily expecting these security tests to pass....

nuclearsandwich:
Windows full rebuild using 6.0.1 [Build Status](https://ci.ros2.org/job/ci_windows/16419/)
Linux full rebuild using 6.0.1 eval bundle [Build Status](https://ci.ros2.org/job/ci_linux/16083/)
Linux full rebuild using 6.0.1 debs [Build Status](https://ci.ros2.org/job/ci_linux/16084/)

linux_docker_resources/entry_point.sh
```diff
@@ -33,19 +33,30 @@ if [ "${ARCH}" != "aarch64" ]; then
...
+	    if test -r /tmp/rti_connext_dds-5.3.1-eval-x64Linux3gcc5.4.0.run -a -r /tmp/rti_security_plugins-5.3.1-eval-x64Linux3gcc5.4.0.rtipkg -a -r /tmp/openssl-1.0.2n-5.3.1-host-x64Linux.rtipkg; then
+		    python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.1-eval-x64Linux3gcc5.4.0.run /home/rosbuild/rti_connext_dds-5.3.1 --rtipkg_paths /tmp/rti_security_plugins-5.3.1-eval-x64Linux3gcc5.4.0.rtipkg /tmp/openssl-1.0.2n-5.3.1-host-x64Linux.rtipkg
+		    mv /tmp/openssl-1.0.2n /home/rosbuild/openssl-1.0.2n
+		    export RTI_OPENSSL_BIN=/home/rosbuild/openssl-1.0.2n/x64Linux3gcc5.4.0/release/bin
+		    export RTI_OPENSSL_LIBS=/home/rosbuild/openssl-1.0.2n/x64Linux3gcc5.4.0/release/lib
+	    elif test -r /tmp/rti_connext_dds-6.0.1-eval-x64Linux4gcc7.3.0.run; then
+		    python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-6.0.1-eval-x64Linux4gcc7.3.0.run /home/rosbuild/rti_connext_dds-6.0.1
+		    export CONNEXTDDS_DIR=/home/rosbuild/rti_connext_dds-6.0.1
+	    fi
```
clalancette, line 53: The indentation in this section seems off compares to the surrounding code.

clalancette (commented):
I've got one style thing inline.  Otherwise this generally looks fine.

Looking at the CI runs, I think most of the failures are expected.  Something weird is going on in the Windows builds, but that also happens in the nightlies and needs to be looked at.  The Linux debian build has all of the same warnings as before, so that's good.  The Linux eval bundle build has one failure that is not elsewhere (https://ci.ros2.org/job/ci_linux/16083/testReport/junit/rclcpp/TestWaitForAllAckedWithParm_TestPublisherWaitForAllAcked/check_wait_for_all_acked_with_QosPolicy_0___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00________________________/).

Other than that, I'd like to see a Foxy and Galactic build on Linux just to ensure that the logic for installing Connext 5.3.1 works properly.

nuclearsandwich:
I've updated the Linux builds to use Connext 6.0.1 pro with security extensions distributed via private submodule.

We'll see how many shell typos I made along the way [Build Status](https://ci.ros2.org/job/ci_linux/16104/)

nuclearsandwich:
Well it took a few tries but here we are with Security extensions on Linux [Build Status](https://ci.ros2.org/job/ci_linux/16111/)

nuclearsandwich:
Ubuntu Jammy AMD64 Connext Debs full CI [Build Status](https://ci.ros2.org/job/ci_linux/16113/)
Ubuntu Jammy AMD64 Connext pro bundle full CI [Build Status](https://ci.ros2.org/job/ci_linux/16112/)
Windows AMD64 Connext pro bundle full CI [Build Status](https://ci.ros2.org/job/ci_windows/16438/)
Ubuntu Focal AMD64 Connext Debs (5.3.1) up to test_security (installation regression check) [Build Status](https://ci.ros2.org/job/ci_linux/16115/)
Ubuntu Focal AMD64 Connext eval bundle (5.3.1) up to test_security (installation regression check) [Build Status](https://ci.ros2.org/job/ci_linux/16114/)

nuclearsandwich:
Once https://github.com/ros-infrastructure/ros2-cookbooks/pull/40 is reviewed for Windows we can update this PR to point back to `latest` in the Windows chef cookbook.

linux_docker_resources/entry_point.sh
clalancette: That's fine, this isn't a blocking problem anyway.  Just curious about it.

.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.sh]
```
clalancette, line 1: Did you mean to check this file in?

linux_docker_resources/entry_point.sh
```diff
@@ -33,19 +33,43 @@ if [ "${ARCH}" != "aarch64" ]; then
...
-                exit 1
+            if test -x /tmp/rti_connext_dds-5.3.1-eval-x64Linux3gcc5.4.0.run -a -r /tmp/rti_security_plugins-5.3.1-eval-x64Linux3gcc5.4.0.rtipkg -a -r /tmp/openssl-1.0.2n-5.3.1-host-x64Linux.rtipkg; then
+                    python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti_connext_dds-5.3.1-eval-x64Linux3gcc5.4.0.run /home/rosbuild/rti_connext_dds-5.3.1 --rtipkg_paths /tmp/rti_security_plugins-5.3.1-eval-x64Linux3gcc5.4.0.rtipkg /tmp/openssl-1.0.2n-5.3.1-host-x64Linux.rtipkg
+                    if [ $? -ne 0 ]; then
+                        echo "Connext not installed correctly (maybe you're on an ARM machine?)." >&2
+                        exit 1
+                    fi
+                    mv /tmp/openssl-1.0.2n /home/rosbuild/openssl-1.0.2n
+                    export RTI_OPENSSL_BIN=/home/rosbuild/openssl-1.0.2n/x64Linux3gcc5.4.0/release/bin
+                    export RTI_OPENSSL_LIBS=/home/rosbuild/openssl-1.0.2n/x64Linux3gcc5.4.0/release/lib
```
clalancette, line 53: I'm really picking at nits here, but the indentation seems to be 8 spaces here whereas it is 4 spaces in the rest of the shell scripting.  Same below.

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -64,7 +65,7 @@ def install_plugin(pkg_path, install_directory):
     try:
         child.expect_exact('Do you want to continue? [Y/n]:')
         child.sendline('y')
-        child.expect(pexpect.EOF)
+        child.expect(pexpect.EOF, timeout=120)
```
clalancette, line 68: This seems to increase the timeout from the default of 30 up to 120.  Do we need that?

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -19,6 +19,7 @@
 
 
 def install_connext(installer_path, install_directory):
+
```
clalancette, line 22: Nit: unnecessary whitespace change.

windows_docker_resources/Dockerfile
```diff
@@ -41,6 +41,8 @@ COPY rticonnextdds-license\ C:\TEMP\rticonnextdds-license
 COPY rticonnextdds-src\ C:\TEMP\rticonnextdds-src
 RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe
 RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg
+RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-host-x64Win64.exe.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-host-x64Win64.exe
+RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64VS2017.rtipkg.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64VS2017.rtipkg
```
clalancette, line 45: Would it make sense here to conditionally copy the correct version depending on the ROS 2 version we are targeting (do we have that information here)?

clalancette (approved):
Everything I left here is a nit.  I'll leave it up to you whether it is worth addressing all, none, or some of the things I pointed out.

.editorconfig
nuclearsandwich: Yeah, if you're gonna nitpick me on hard/soft tabs then I'd like the project to enforce the correct tabbing.

linux_docker_resources/rti_web_binaries_install_script.py
nuclearsandwich: Yeah we definitely do. Installing several of the rtipkgs timed out during test runs.

```
00:00:07.300 Installing Connext...
00:00:07.300 Installing Connext binaries off RTI website...
00:01:02.550 Traceback (most recent call last):
00:01:02.550   File "/tmp/rti_web_binaries_install_script.py", line 67, in install_plugin
00:01:02.551     child.expect(pexpect.EOF)
00:01:02.551   File "/usr/lib/python3/dist-packages/pexpect/spawnbase.py", line 343, in expect
00:01:02.557     return self.expect_list(compiled_pattern_list,
00:01:02.557   File "/usr/lib/python3/dist-packages/pexpect/spawnbase.py", line 372, in expect_list
00:01:02.557     return exp.expect_loop(timeout)
00:01:02.557   File "/usr/lib/python3/dist-packages/pexpect/expect.py", line 181, in expect_loop
00:01:02.565     return self.timeout(e)
00:01:02.565   File "/usr/lib/python3/dist-packages/pexpect/expect.py", line 144, in timeout
00:01:02.565     raise exc
00:01:02.565 pexpect.exceptions.TIMEOUT: Timeout exceeded.
00:01:02.565 <pexpect.pty_spawn.spawn object at 0x7fba90548a90>
00:01:02.565 command: /home/rosbuild/rti_connext_dds-6.0.1/bin/rtipkginstall
00:01:02.565 args: [b'/home/rosbuild/rti_connext_dds-6.0.1/bin/rtipkginstall', b'/tmp/rticonnextdds-src/rti_connext_dds-6.0.1.25-pro-host-x64Linux.rtipkg']
00:01:02.565 buffer (last 100 chars): 'g rti_connext_dds-6.0.1.25-pro-host-x64Linux.rtipkg. Please wait, this may take several minutes...\r\n'
```

linux_docker_resources/rti_web_binaries_install_script.py
nuclearsandwich: Whoops. Nice catch. That was left from a pdb session.

windows_docker_resources/Dockerfile
nuclearsandwich: I don't think that it does. Conditionalizing would result in smaller docker images at the expense of being able to share cache layers between different combinations of CI options and since we don't transport images but instead build them on-host it makes sense to prefer cache-friendliness to image size. Parameters in the Chef configuration will configure which installer is actually run.

linux_docker_resources/entry_point.sh
nuclearsandwich: Appreciate the close read.

nuclearsandwich:
> Ubuntu Focal AMD64 Connext Debs (5.3.1) up to test_security (installation regression check) [Build Status](https://ci.ros2.org/job/ci_linux/16117/)

:white_check_mark: Connext 5.3.1 is installed via debs and rti_connext_dds_cmake_module finds it.
:white_check_mark: Security tests do not pass due to lack of security extensions.

> Ubuntu Focal AMD64 Connext eval bundle (5.3.1) up to test_security (installation regression check) [Build Status](https://ci.ros2.org/job/ci_linux/16116/)

:white_check_mark: Connext 5.3.1 is installed via pexpect script and rti_connext_dds_module finds it.
:white_check_mark: test_security tests pass
:white_check_mark: [this error](https://ci.ros2.org/job/ci_linux/16116/testReport/rclcpp/TestWaitForAllAckedWithParm_TestPublisherWaitForAllAcked/check_wait_for_all_acked_with_QosPolicy_0___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00________________________/) seems present in both 5.3.1 and 6.0.1 when Connext is the only/default rmw_implementation

nuclearsandwich:
> Ubuntu Jammy AMD64 Connext pro bundle full CI [Build Status](https://ci.ros2.org/job/ci_linux/16112/)

:white_check_mark: Connext 6.0.1 is installed via pro bundle with pexpect script (we could use the unattended mode but the pexpect script is working and I'd rather wait to refactor until we don't need it for 5.3.1)
:white_check_mark: test_security tests pass
:white_check_mark: [this error](https://ci.ros2.org/job/ci_linux/16112/testReport/junit/rclcpp/TestWaitForAllAckedWithParm_TestPublisherWaitForAllAcked/check_wait_for_all_acked_with_QosPolicy_0___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00________________________/) seems presen in both 5.3.1 and 6.0.1 when Connext is the only/default rmw implementation.

nuclearsandwich:
> Ubuntu Jammy AMD64 Connext Debs full CI [Build Status](https://ci.ros2.org/job/ci_linux/16113/)

:white_check_mark: mypy tests are expected on Jammy right now (see #617)
:white_check_mark: [this error](https://ci.ros2.org/job/ci_linux/16113/testReport/rclcpp/TestWaitForAllAckedWithParm_TestPublisherWaitForAllAcked/check_wait_for_all_acked_with_QosPolicy_0___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00___88_byte_object__01_00_00_00_00_00_00_00_01_00_00_00_00_00_00_00_01_00_00_00_02_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00_00________________________/) seems present in both 5.3.1 and 6.0.1 when Connext is the only/default rmw implementation.
:white_check_mark: security tests fail because the deb lacks support for RTI security extensions.

nuclearsandwich:
Based on the [test results](https://ci.ros2.org/job/ci_windows/16445/testReport/test_security/) for the last Windows job it does not appear that rmw_connextdds was used to run security tests.

nuclearsandwich:
> Windows AMD64 Connext pro bundle full CI [Build Status](https://ci.ros2.org/job/ci_windows/16445/)


:white_check_mark: ConnextDDS 6.0.1 is properly installed and detected.
:white_check_mark: rclpy and other test failures are present in nightlies on Windows
:white_check_mark: security tests [are not run](https://github.com/ros2/ci/pull/614#issuecomment-1033849213) in a full CI build with rmw_connextdds

nuclearsandwich:
One last test to verify that 5.3.1 still installs successfully for a non-Rolling configuration.

Windows AMD64 Galactic pro bundle [Build Status](https://ci.ros2.org/job/ci_windows/16451/)

PR #615, opened 2022-02-03
Include typesupport_introspection_* packages in quality level testing

j-rivero:
After the work done to rise typesupport_introspection_c/cpp packages to Q1 (https://github.com/ros2/rosidl/pull/656 and https://github.com/ros2/rosidl/pull/657), it would be great to add them to nightly build for coverage. 

The PR moves the two packages to `quality_level_pkgs` and include the test package that exercise both in `testing_pkgs_for_quality_level`.


gbiggs (approved):

clalancette:
So this broke two jobs last night:

* https://ci.ros2.org/view/nightly/job/nightly_linux_foxy_coverage/
* https://ci.ros2.org/view/nightly/job/nightly_linux_galactic_coverage/

The issue is that Foxy and Galactic don't have the new `rosidl_typesupport_introspection_tests` packages, so fail to build now.  I think we need to make this conditional on `!foxy !galactic`.  @j-rivero can you take a look?

PR #616, opened 2022-02-05
Set Jammy as the default Ubuntu distribution.

nuclearsandwich:
Make Jammy the default Ubuntu distribution for Linux CI.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -108,7 +108,7 @@ def main(argv=None):
         'enable_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Note that this won't affect the galactic or foxy coverage jobs because they override the default distribution:

https://github.com/ros2/ci/blob/34b4484a359268c1b2f819058eb8e100fed93c9f/create_jenkins_job.py#L508-L509
https://github.com/ros2/ci/blob/34b4484a359268c1b2f819058eb8e100fed93c9f/create_jenkins_job.py#L490-L491

clalancette (approved):

nuclearsandwich:
This was merged but not yet deployed. I am deploying it now

PR #617, opened 2022-02-07
Unpin mypy.

clalancette:

This should fix the warnings we are seeing on Jammy, like https://ci.ros2.org/job/ci_linux/16057/ .  On the other hand, I don't know what the situation is going to be like on Foxy and Galactic on Focal, so testing still needs to be done here.

### DIFF
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -42,7 +42,7 @@ RUN dnf update \
 RUN rosdep init && rosdep update
 
 # Install some stuff via Pip which we don't have available in EPEL
-RUN python3 -m pip install mypy==0.761 pydocstyle
+RUN python3 -m pip install pydocstyle
 
 # Install a newer version of pytest
 RUN python3 -m pip install -U pytest pytest-rerunfailures
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -518,13 +518,13 @@ def run(args, build_function, blacklisted_package_names=None):
 
         # Mypy is a special case.  We prefer to get it from the distribution, if it exists there.
         # If not, we want to install it via pip.  So we try to import it here to see if it exists.
-        need_mypy_from_pip = False
+        need_mypy_from_pipy = False
         try:
             import mypy
         except ModuleNotFoundError:
-            need_mypy_from_pip = True
+            need_mypy_from_pipy = True
 
-        if need_mypy_from_pip:
+        if need_mypy_from_pipy:
             if args.ros_distro in ["foxy", "galactic"]:
                 pip_packages += ["mypy==0.761"]
             else:
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -65,7 +65,6 @@ pip_dependencies = [
     'importlib-metadata',
     'lark-parser',
     'mock',
-    'mypy==0.761',
     'nose',
     'pep8',
     'pydocstyle',
@@ -516,6 +515,21 @@ def run(args, build_function, blacklisted_package_names=None):
         job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
         # Install pip dependencies
         pip_packages = list(pip_dependencies)
+
+        # Mypy is a special case.  We prefer to get it from the distribution, if it exists there.
+        # If not, we want to install it via pip.  So we try to import it here to see if it exists.
+        need_mypy_from_pip = False
+        try:
+            import mypy
+        except ModuleNotFoundError:
+            need_mypy_from_pip = True
+
+        if need_mypy_from_pip:
+            if args.ros_distro in ["foxy", "galactic"]:
+                pip_packages += ["mypy==0.761"]
+            else:
+                pip_packages += ["mypy==0.931"]
+
         if sys.platform == 'win32':
             # Install fork of pyreadline containing fix for deprecation warnings
             # TODO(jacobperron): Until upstream issue is resolved https://github.com/pyreadline/pyreadline/issues/65


clalancette:
CI using this branch.

Rolling:
* Focal Linux [Build Status](http://ci.ros2.org/job/ci_linux/16095/)
* Focal Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10740/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16429/)
* Jammy Linux [Build Status](http://ci.ros2.org/job/ci_linux/16097/)
* Jammy Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10742/)

Foxy:
* Focal Linux [Build Status](http://ci.ros2.org/job/ci_linux/16098/)
* Focal Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10743/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16432/)

(if everything above is happy, I'll run Galactic CI tomorrow)

nuclearsandwich:
> Install mypy from distribution packages on Focal and Jammy, and only install from pip for Windows.

If I recall correctly pinning to the version in Focal was the agreed upon solution in the 2021-06-15 ROS 2 meeting but there's no mention there of doing so via system package or pip. I'm in favor of using the debs and rpms in CI in order to more closely replicate the buildfarm environment.

clalancette:
> If I recall correctly pinning to the version in Focal was the agreed upon solution in the 2021-06-15 ROS 2 meeting but there's no mention there of doing so via system package or pip. I'm in favor of using the debs and rpms in CI in order to more closely replicate the buildfarm environment.

Yep, agreed.  That is my preferred solution as well.  I'll try this out and see how it goes.

clalancette:
All right, the new code I pushed is more complicated, but also should get us closer to what we want.  Let's see how this goes:

Rolling:
* Focal Linux [Build Status](http://ci.ros2.org/job/ci_linux/16118/)
* Focal Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10749/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16446/)
* Jammy Linux [Build Status](http://ci.ros2.org/job/ci_linux/16119/)

Foxy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16120/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16447/)

clalancette:
I've redone this completely, and what I've done should make it much more robust.  Let's try another CI to see what happens:

Rolling:
* Ubuntu Focal [Build Status](http://ci.ros2.org/job/ci_linux/16160/)
* Ubuntu Focal-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10789/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16496/)
* Ubuntu Jammy [Build Status](http://ci.ros2.org/job/ci_linux/16161/)

Foxy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16162/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10790/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16497/)

linux_docker_resources/Dockerfile-RHEL
```diff
@@ -42,7 +42,7 @@ RUN dnf update \
 RUN rosdep init && rosdep update
 
 # Install some stuff via Pip which we don't have available in EPEL
-RUN python3 -m pip install mypy==0.761 pydocstyle
+RUN python3 -m pip install pydocstyle
```
nuclearsandwich, line 45: I don't see mypy in the system dependencies installed later.
But because the other changes in this branch will install rosdistro-dependent versions of mypy dropping it here seems reasonable.

ros2_batch_job/__main__.py
```diff
@@ -516,6 +515,21 @@ def run(args, build_function, blacklisted_package_names=None):
...
+
+        # Mypy is a special case.  We prefer to get it from the distribution, if it exists there.
+        # If not, we want to install it via pip.  So we try to import it here to see if it exists.
+        need_mypy_from_pip = False
+        try:
+            import mypy
+        except ModuleNotFoundError:
+            need_mypy_from_pip = True
+
+        if need_mypy_from_pip:
```
nuclearsandwich, line 527: Just for the rhythm of it :grin:

```suggestion
        need_mypy_from_pipy = False
        try:
            import mypy
        except ModuleNotFoundError:
            need_mypy_from_pipy = True

        if need_mypy_from_pipy:
```

nuclearsandwich (commented):
Looks good so far. Hanging on to the :heavy_check_mark: until CI comes back happy.

linux_docker_resources/Dockerfile-RHEL
clalancette: Ah yes, RHEL is an interesting case.  RHEL does not package mypy, either in the main repositories or in EPEL.  So we have to install it from pip (which this patch nicely accomplishes without special cases).

As a side note, we get away with releasing `ament_mypy` into RHEL on https://build.ros2.org because we specifically patch out mypy as a dependency there.  It means that an `dnf install ros-rolling-ament-mypy` on RHEL doesn't result in a fully-functional system, but that was deemed an acceptable tradeoff until the day that mypy gets packaged into RHEL.

clalancette:
CI was really happy with this, so I converted it to a real PR.  Please take another look when you get a chance.

nuclearsandwich (approved):

clalancette:
Here's CI just up to ament_cmake, to test that the latest variable names don't have a typo.  If this is successful, I'm going to go ahead and merge.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16176/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16517/)

PR #622, opened 2022-02-21
Only install ros1 dependencies when there is a valid ROS 1 distro.

nuclearsandwich:
When on Ubuntu Jammy there is no official ROS 1 distribution available.

Before merging this I want to run it and see how the ros1_bridge builds (hopefully it does just trivially) but it may not in which case more changes will need to be added to better support disabling the bridge.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -135,7 +135,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y python3-dev
 RUN echo "@today_str"
 
 # Install build and test dependencies of ros1_bridge if there is a valid ROS 1 distro.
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Ubuntu [Build Status](https://ci.ros2.org/job/ci_packaging_linux/478/)
Ubuntu ARM64 [Build Status](https://ci.ros2.org/job/ci_packaging_linux-aarch64/86/)

nuclearsandwich:
Merging pre-review in order to try and get CI in an improved state.

clalancette:
Looks like an improvement to me, so :+1: .

Blast545 (commented):
LGTM

PR #624, opened 2022-02-25
Check if test_branch exists in at least one repo among the ones in repos file

adityapande-1995:
Currently, if the supplied ``test_branch`` is not found in any of the repos in the .repos file, the workflow will just skip it and run CI on the default branches. There should be an error, or at least a warning saying that none of the repos contain the branch, as discussed offline with @clalancette 

Signed-off-by: Aditya <aditya050995@gmail.com>

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -654,6 +654,10 @@ def run(args, build_function, blacklisted_package_names=None):
                 info('Attempting to create a well known branch name for all the default branches')
                 job.run(vcs_cmd + ['custom', '.', '--git', '--args', 'checkout', '-b', '__ci_default'])
 
+                # Check if the branch exists in at least one repo from the given repos file
+                for filename in repos_filenames:
+                    branch_found = check_branch_exists(filename, args.test_branch)
+                    if not branch_found: return False
                 # Attempt to switch all the repositories to a given branch
                 info("Attempting to switch all repositories to the '{0}' branch"
                      .format(args.test_branch))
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -654,6 +654,11 @@ def run(args, build_function, blacklisted_package_names=None):
                 info('Attempting to create a well known branch name for all the default branches')
                 job.run(vcs_cmd + ['custom', '.', '--git', '--args', 'checkout', '-b', '__ci_default'])
 
+                # Check if the branch exists in at least one repo from the given repos file
+                for filename in repos_filenames:
+                    branch_found = check_branch_exists(filename, args.test_branch)
+                    if not branch_found:
+                        print('Warning: None of the repos in ', filename, ' contain the branch ', args.test_branch)
```
jacobperron, line 661: How visible is this warning? Does it make the job yellow or do we have to notice it in the log?

IMO, it might be better to just abort the job early since it seems most likely not what the user intended and continuing with the build is probably not worth the time.

ros2_batch_job/__main__.py
```diff
@@ -777,5 +782,24 @@ def _fetch_repos_file(url, filename, job):
...
+# Checks if the given branch exists in at least one repo
+def check_branch_exists(filename, branch):
+    # Populate all .git repos
+    lines = []
+    with open(filename) as f: lines = f.readlines()
+
+    urls = []
+    for line in lines:
+        text = line.strip()
+        if text[-4:] == ".git":
```
jacobperron, line 794: It'd be nice if there is a vcs command we could use instead. I think checking for a ".git" suffix might be too brittle since I don't think it's a requirement that it's there. Also, since the repositories are already cloned locally, it seems a bit redundant to re-parse the repos files.

ros2_batch_job/__main__.py
clalancette: Agreed; if the user specified a branch, but we couldn't find any of them, let's just abort the build early.  It's probably not what was intended.  We also have to be careful to succeed if the branch name is empty; that is the "normal" state of affairs, in which case we just want to continue on.

ros2_batch_job/__main__.py
clalancette: That's right, this is too brittle.  The `.git` isn't required (though it is recommended).

Actually, I wonder if it would be better to just parse these `repos` files as the YAML files they are.  That will be way less brittle, and doesn't require us to use `os.system` (or this brittle parsing) at all.

PR #627, opened 2022-03-04
Update job retention policy and refactor retention configuration.

nuclearsandwich:
Due to the way the job creation script is laid out updating the job retention configuration was proving to be a pain so this PR creates a helper function for setting job retention data based on the job name prefix and automatically merges that with the job configuration inside the configure_job helper function.

This will also hopefully make it easier to adjust and maintain job retention policies in the future. There are two separate things to review here. The initial commit is a refactor of the retention configuration which maintains the current retention values and the follow-up commit (potentially commits) update the actual policy values based on current capacity and usage patterns.


clalancette (approved):
Looks good to me!

PR #628, opened 2022-03-06
Increase default job timeout

jlblancoc:
We saw build timeouts in one (big) package ([Jenkins log](https://build.ros2.org/job/Gdev__mrpt2__ubuntu_focal_amd64/lastBuild/)), so it would be great to have more time before timing out. 
Alternatively we could disable part of the features / cli tools of this particular package, but this patch would be less painful if you agree with the change.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -107,7 +107,7 @@ def main(argv=None):
         'compile_with_clang_default': 'false',
         'enable_coverage_default': 'false',
         'dont_notify_every_unstable_build': 'false',
-        'build_timeout_mins': 0,
+        'build_timeout_mins': 200,
         'ubuntu_distro': 'jammy',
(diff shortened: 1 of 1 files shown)


clalancette:
I know this is a bit confusing, but actually the code in here has nothing to do with what happens on https://build.ros2.org.  This code is what is used to run nightly CI on https://ci.ros2.org, so this won't likely help your use case.

If you want to propose increasing the timeouts on https://build.ros2.org, the correct place is https://github.com/ros2/ros_buildfarm_config .  That said, those timeouts are there to prevent runaway jobs from holding onto our limited workers for long periods of time, so you'll have to provide a good justification for it.

With all of that, I'm going to close this.  Feel free to reopen if I've misunderstood something.

jlblancoc:
Thanks, very clarifying, indeed here's not the right place for my request!

PR #632, opened 2022-03-17
Add pytest-timeout as pip dependency

audrow:
This is necessary for https://github.com/ros2/ros2cli/pull/701.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -538,7 +538,7 @@ def run(args, build_function, blacklisted_package_names=None):
                 pip_packages += ["mypy==0.931"]
 
         # We prefer to get pytest-timeout from the distribution if it exists.  If not we install it via pip.
-        if need_package_from_pipy("pytest-timeout"):
+        if need_package_from_pipy("pytest_timeout"):
             pip_packages += ["pytest-timeout==2.1.0"]
 
         # We prefer to get lark from the distribution if it exists.  If not we install it via pip.
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -76,7 +76,6 @@ pip_dependencies = [
     'pytest-repeat',
     'pytest-rerunfailures',
     'pytest-runner',
-    'pytest-timeout',
     'pyyaml',
     'vcstool',
     'yamllint',
@@ -537,6 +536,10 @@ def run(args, build_function, blacklisted_package_names=None):
                 pip_packages += ["mypy==0.761"]
             else:
                 pip_packages += ["mypy==0.931"]
+
+        # We prefer to get pytest-timeout from the distribution if it exists.  If not we install it via pip.
+        if need_package_from_pipy("pytest-timeout"):
+            pip_packages += ["pytest-timeout==2.1.0"]
 
         # We prefer to get lark from the distribution if it exists.  If not we install it via pip.
         if need_package_from_pipy("lark"):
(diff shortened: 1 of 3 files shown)


ros2_batch_job/__main__.py
```diff
@@ -76,6 +76,7 @@
     'pytest-repeat',
     'pytest-rerunfailures',
     'pytest-runner',
+    'pytest-timeout',
```
clalancette, line 79: We're generally preferring to use the system versions of dependencies on Linux now.  What this means is that you should add `python3-pytest-timeout` as an apt installed dependency to https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile , and as a dnf installed dependency to https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile-RHEL .  Then in this file you should follow the pattern in https://github.com/ros2/ci/blob/master/ros2_batch_job/__main__.py#L535-L540 and try to import `pytest_timeout`, and only install it via pip if that fails.

clalancette (changes requested):

ros2_batch_job/__main__.py
audrow: Should be fixed in https://github.com/ros2/ci/pull/632/commits/54c6aafcfa866a79d0dc46d672205581fc169ca3.

ros2_batch_job/__main__.py
```diff
@@ -532,6 +532,10 @@ def need_package_from_pipy(pkg_name):
             else:
                 pip_packages += ["mypy==0.931"]
 
+        # We prefer to get pytest-timeout from the distribution if it exists.  If not we install it via pip.
+        if need_package_from_pipy("pytest-timeout"):
```
clalancette, line 536: This is annoying, but I'm pretty sure that to work, this needs to be:

```suggestion
        if need_package_from_pipy("pytest_timeout"):
```

audrow: Is it different than the pip install instructions? `pip install pytest-timeout`

https://pypi.org/project/pytest-timeout/

clalancette: Yeah, the problem is that you can't do things like:

```
import pytest-timeout
```

as the Python interpreter things that is a minus sign.  In this particular case, `need_package_from_pipy` is using importlib with a string, but in my local testing this still doesn't work:

```
Python 3.10.2+ (main, Mar 14 2022, 05:18:40) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib
>>> importlib.import_module('pytest-timeout')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'pytest-timeout'
>>> importlib.import_module('pytest_timeout')
<module 'pytest_timeout' from '/usr/lib/python3/dist-packages/pytest_timeout.py'>
>>> 
```

audrow: Ah, I see. Thanks for the catch.

audrow: Fixed in https://github.com/ros2/ci/pull/632/commits/472c21265116108b02d4f8930ee233bbf8117424.

clalancette (approved):
This looks good to me now, but let's hold off merging it until we get CI in a bit happier shape.

audrow:
CI was run here: https://github.com/ros2/ros2cli/pull/701#issuecomment-1082382430.

PR #633, opened 2022-03-17
Use the system version of virtualenv on Linux.

clalancette:

In theory this should help us deal with issues with the pinned version of virtualenv and Python 3.10.  Because of https://github.com/ros2/ci/commit/bd47ae1fbe20133a95f2ae527e91007e25cd2218, I'm not sure what's going to happen, but we'll see.


clalancette:
All right, that mostly worked.  There are some new flaky tests, but that as much may be 3.10 changes as this change.  So I'm going to transition this to a real PR and run full CI on it.

clalancette:
Superseded by #638

PR #634, opened 2022-03-17
Ignore packages that require pyside2 when using python3.10.

nuclearsandwich:
While using python 3.10 ignore these packages.

https://bugs.launchpad.net/ubuntu/+source/pyside2/+bug/1963583


clalancette (approved):
This looks fine to me for now; note that this is temporary until we get pyside2/shiboken rebuilt for Python 3.10 in Jammy.

clalancette:
We shouldn't need this anymore, so closing this out.

PR #636, opened 2022-03-17
Automatically pull the base image when running docker builds on Linux.

nuclearsandwich:
I would like to make this an optional thing but in the meantime I just want to try it.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -218,11 +218,11 @@ sed -i "s|@@workdir|`pwd`|" linux_docker_resources/entry_point.sh
 echo "# END SECTION"
 echo "# BEGIN SECTION: Build Dockerfile"
 @[    if os_name == 'linux-aarch64']@
-docker build ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_batch_ci_aarch64 linux_docker_resources
+docker build --pull ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_batch_ci_aarch64 linux_docker_resources
 @[    elif os_name == 'linux-rhel']@
-docker build ${DOCKER_BUILD_ARGS} -t ros2_batch_ci_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
 @[    elif os_name == 'linux']@
-docker build ${DOCKER_BUILD_ARGS} -t ros2_batch_ci linux_docker_resources
+docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci linux_docker_resources
 @[    else]@
 @{ assert False, 'Unknown os_name: ' + os_name }@
 @[    end if]@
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
> I would like to make this an optional thing but in the meantime I just want to try it.

Coming back to this after the weekend, I'm no longer sure that it makes sense make this optional. I was trying to decide whether it should be on or off by default. Off by default is consistent with the current behavior but the current behavior gets us stuck on outdated base images for the lifetime of a host's docker image cache and _never_ updates the Docker base image.

So part of the reason for adding this change would be changing the behavior and defaulting it to on. This matches `ros_buildfarm` which independently runs `docker pull ...` before running `docker build` meaning that in addition to the daily docker cache invalidation we'll also fully invalidate whenever the base image is updated on Docker. This takes control of *when* we update away from us but matches better with our current behavior elsewhere. With that in mind I was then struggling to come up with a reason that we would want to turn the pull _off_ at all. If the fresh image is busted in some way, we'll likely not find that out until after it starts getting used which means that some hosts will already have it so just not pulling isn't a complete solution and we'll have to use something like #640 to fully restore service.

So now I'm leaning towards just plonking this in as-is. Alternatives would be some kind of "last updated" check which only does the pull if the docker image wasn't pulled within the current day which I would match to the daily invalidation so that they're done at the same time or putting a cron job on the agent hosts which runs `docker pull` in the background but I'm not a big fan of that option even as I put it on the table.

nuclearsandwich:
I'm intentionally leaving the conflicts in here as they'll go away once #642 is merged.

clalancette:
My feeling on this is complicated, and clearly influenced by recent events.

On the one hand, I agree with you that always doing a pull makes us more consistent with the buildfarm.  This also ensures that we are always building against the latest up-to-date docker images, which is a good idea for security and other issues.

However, it is exactly the discrepancy between systems which let us continue to run CI while the buildfarm was broken waiting on changes from upstream Ubuntu.  That is, we could continue to make progress on ROS 2 while waiting for upstream to settle down.

Now, clearly this change was a fairly remarkable event; it really only has the potential to happen when we change Linux platforms.  But since we do that every 2 years, it is not *that* unremarkable.  You can also argue that we should have done something more to deal with this problem earlier on.

At the end of the day, I think I'm OK with this going in, as long as we have some strategy for dealing with "bad" Docker images in the future.  I don't know what form that takes, but that is at least my feeling on the subject.

nuclearsandwich:
> However, it is exactly the discrepancy between systems which let us continue to run CI while the buildfarm was broken waiting on changes from upstream Ubuntu. That is, we could continue to make progress on ROS 2 while waiting for upstream to settle down.
> ...
> At the end of the day, I think I'm OK with this going in, as long as we have some strategy for dealing with "bad" Docker images in the future. I don't know what form that takes, but that is at least my feeling on the subject.

I think you're totes right but the problem with the status quo is that CI doesn't even tell you things are broken so as a team there was time spent [disbelieving Cassandra](https://en.wikipedia.org/wiki/Cassandra) which caused the problems we were having to be dismissed as "not mine" for a while. I think that in general, when there are platform-level issues, our CI should tell us. In preparing #640 we've got a template for how to roll out "pinning" a docker image. I'm not ready to making doing so a configurable CI option, I'd rather see us run into problems like this a few more times and solve them with patches to demonstrate the need.

cottsay (approved):

audrow (approved):
I think this is a good idea.

PR #639, opened 2022-03-18
Perform all apt operations except those targeting the ROS repositories before daily invalidation.

nuclearsandwich:
There is no reason that these installations should be performed after the daily cache bust. As far as I can tell their presence here is entirely accidental.

The ROS 1 packages being installed after invalidation makes some sense since ROS distributions change more frequently than the base Ubuntu or Debian release.


clalancette (approved):
This looks reasonable to me; I agree that this was probably an oversight.  I'll just leave a gentle reminder that any changes to the Dockerfile are going to interfere with our current "use a cached Jammy docker image with Python 3.9" strategy.  But I'm sure you'll merge this when appropriate.

PR #643, opened 2022-03-21
Use system versions of setuptools/pip unless there is none available.

nuclearsandwich:
Installing these packages from pip results in CI volatility. Specifically, setuptools 60.10.0 appears to interfere with the install paths, unconditionally inserting `local` in the path so we're left with `$PREFIX/local/bin` rather than `$PREFIX/bin` as expected.

pip is somewhat of a coattail rider on this commit but is being sourced from apt for similar reasons.

I've omitted the installation of virtualenv entirely because of parallel patches currently in flight. That may be revisited before this is ready for merge.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -505,18 +505,6 @@ def run(args, build_function, blacklisted_package_names=None):
 
     # Now inside of the workspace...
     with change_directory(args.workspace):
-        print('# BEGIN SUBSECTION: install Python packages')
-        # Update setuptools
-        job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U', 'pip', 'setuptools'],
-                shell=True)
-        # Print setuptools version
-        job.run(['"%s"' % job.python, '-c', '"import setuptools; print(setuptools.__version__)"'],
-                shell=True)
-        # Print the pip version
-        job.run(['"%s"' % job.python, '-m', 'pip', '--version'], shell=True)
-        # Install pip dependencies
-        pip_packages = list(pip_dependencies)
-
         def need_package_from_pipy(pkg_name):
             try:
                 importlib.import_module(pkg_name)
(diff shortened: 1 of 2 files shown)


ros2_batch_job/__main__.py
```diff
@@ -517,14 +527,6 @@ def run(args, build_function, blacklisted_package_names=None):
...
         pip_packages = list(pip_dependencies)
 
-        def need_package_from_pipy(pkg_name):
-            try:
-                importlib.import_module(pkg_name)
-            except ModuleNotFoundError:
-                return True
-
-            return False
-
```
nuclearsandwich, line 527: This function isn't removed it's just relocated so it can be used earlier.

https://github.com/ros2/ci/pull/643/files#diff-54f7a5a1191410c54331d5e337e5760c7de5e219f01dfc732b9380dbf71b46b7R508-R515

ros2_batch_job/__main__.py
```diff
@@ -505,10 +505,20 @@ def run(args, build_function, blacklisted_package_names=None):
...
+
         print('# BEGIN SUBSECTION: install Python packages')
-        # Update setuptools
-        job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U', 'pip', 'setuptools'],
-                shell=True)
+
+        # Update setuptools if it is not already available.
+        if need_package_from_pipy('setuptools'):
+            job.run(['"%s"' % job.python, '-m', 'pip', 'install', '-U', 'pip', 'setuptools'],
+                    shell=True)
```
nuclearsandwich, line 521: I think that this is actually dead code without MacOS and we should consider removing it entirely.

* Ubuntu is installing setuptools from apt during image build.
* RHEL in installing setuptools from dnf during image build.
* Windows is installing setuptools from pip (via chef) during image build.

So only macOS lacks any kind of per-job container (aside from a virtualenv) where setuptools would need to be updated each build in order to stay current.

clalancette: It's totally fine by me to remove it.

nuclearsandwich: Done in 82c018c

audrow (approved):
Looks good!

PR #650, opened 2022-03-30
Pin setuptools to a version we know works on Windows Debug.

clalancette:

This should theoretically fix our Windows Debug issues.  Still need to run CI to verify this.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -553,6 +553,10 @@ def run(args, build_function, blacklisted_package_names=None):
             # TODO(jacobperron): Until upstream issue is resolved https://github.com/pyreadline/pyreadline/issues/65
             pip_packages += ['git+https://github.com/osrf/pyreadline']
 
+            # Setuptools > 61 somehow have broken Windows Debug.  Pin it to 59.6.0 here which
+            # matches Ubuntu Jammy, and wait until upstream setuptools settles down.
+            pip_packages += ["setuptools==59.6.0"]
+
             if args.cmake_build_type == 'Debug':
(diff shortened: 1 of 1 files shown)


clalancette:
* Windows Debug before [Build Status](http://ci.ros2.org/job/ci_windows/16829/)
* Windows Debug after [Build Status](http://ci.ros2.org/job/ci_windows/16828/)

clalancette:
OK, yeah, that does it.  I don't love this solution, but I think it will get Windows Debug back to !red for today.  Marking this as ready for review.

Blast545 (approved):

ros2_batch_job/__main__.py
```diff
@@ -553,6 +553,10 @@ def need_package_from_pipy(pkg_name):
             # TODO(jacobperron): Until upstream issue is resolved https://github.com/pyreadline/pyreadline/issues/65
             pip_packages += ['git+https://github.com/osrf/pyreadline']
 
+            # Setuptools > 61 somehow have broken Windows Debug.  Pin it to 59.6.0 here which
+            # matches Ubuntu Jammy, and wait until upstream setuptools settles down.
+            pip_packages += ["setuptools==59.6.0"]
```
nuclearsandwich, line 558: I think this should supplant the cookbook configuration but it may make sense to update the cookbook to account for this as well. Especially if the problems this resolves occur during image build https://github.com/ros-infrastructure/ros2-cookbooks/blob/latest/cookbooks/ros2_windows/recipes/pip_installs.rb#L48-L52

nuclearsandwich (approved):

PR #651, opened 2022-03-30
Assume pre-configured environment variables instead of setup script for Connext

nuclearsandwich:
The ROS 2 cookbook is setting environment variables used by rti_connext_dds_cmake_module and test_security. The Windows setup should check that those are set rather than overriding them with files found on the host.


nuclearsandwich:
Rolling [Build Status](https://ci.ros2.org/job/ci_windows/16838/)

clalancette (approved):
Looks reasonable to me assuming that CI comes back green on Rolling, Galactic, and Foxy.

PR #652, opened 2022-04-08
Use Connext 6 for Rolling and beyond on RHEL.

nuclearsandwich:
This is a continuation of #614 which adds support for RHEL while retaining the ability to build using 5.3.1 for Foxy and Galactic.


linux_docker_resources/Dockerfile
```diff
@@ -3,6 +3,7 @@ ARG BRIDGE=false
 ARG INSTALL_CONNEXT_DEBS=false
 ARG PLATFORM=x86
 ARG ROS1_DISTRO=noetic
+ARG ROS_DISTRO=rolling
```
nuclearsandwich, line 6: Although this build argument is not yet used, and is not strictly necessary for this patch, I felt it worthwhile to apply the argument to both Ubuntu and RHEL docker builds. If preferred I could walk this back though.

linux_docker_resources/Dockerfile-RHEL
```diff
@@ -119,8 +120,21 @@ RUN dnf install \
     --refresh -y
 
 # Get and install the RTI web binaries.
-RUN cd /tmp && curl --silent https://s3.amazonaws.com/RTI/Bundles/5.3.1/Evaluation/rti_connext_dds_secure-5.3.1-eval-x64Linux3gcc5.4.0.tar.gz | tar -xz
-RUN cd /tmp && tar -xvf /tmp/openssl-1.0.2n-target-x64Linux3gcc5.4.0.tar.gz
+# Connext 5.3.1 for galactic and earlier.
+RUN if [ "$ROS_DISTRO" = "foxy" -o "$ROS_DISTRO" = "galactic" ]; then \
+       cd /tmp && curl --silent https://s3.amazonaws.com/RTI/Bundles/5.3.1/Evaluation/rti_connext_dds_secure-5.3.1-eval-x64Linux3gcc5.4.0.tar.gz | tar -xz \
+       && tar -xvf /tmp/openssl-1.0.2n-target-x64Linux3gcc5.4.0.tar.gz; \
```
nuclearsandwich, line 126: It made more sense to me to make this one long multiline RUN than it did to perform the same check on two consecutive RUN lines.

linux_docker_resources/Dockerfile
clalancette: Agreed, I think we should keep them consistent.  If we also go forth with the plan to make these Dockerfiles install "everything needed to build", then I think we'll need this anyway.

nuclearsandwich:
CI

Rolling [Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/1/)
Galactic [Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/2/)
Foxy [Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/3/)

linux_docker_resources/Dockerfile-RHEL
clalancette: Agreed with that, but isn't this missing a `fi` somewhere?

nuclearsandwich: sigh. lost in the rebases.

nuclearsandwich: [9bc507e](https://github.com/ros2/ci/pull/652/commits/9bc507e1ebfdd2b78b79599b880b9f067779d274)

nuclearsandwich:
CI up to test_communication and test_security

Rolling [Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/4/)
Galactic [Build Status](https://ci.ros5.org/job/test_ci_linux-rhel/5/)
Foxy [Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/6/)

cottsay (approved):
This seems like the right approach.

We should probably align the Ubuntu Dockerfile with this approach. Right now, if we ran a Rolling job on Focal, we'd get the wrong Connext version. Probably save that for a follow-up PR though.

nuclearsandwich:
> We should probably align the Ubuntu Dockerfile with this approach. Right now, if we ran a Rolling job on Focal, we'd get the wrong Connext version. Probably save that for a follow-up PR though.

That's a good point. Done in #653 for Monday.

PR #653, opened 2022-04-09
Use ROS_DISTRO rather than UBUNTU_DISTRO to decide connext version.

nuclearsandwich:
This both better matches what we recommend and matches the support which was just added for RHEL.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -90,7 +90,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y python3-pexpect
 
 # Get and install the RTI web binaries.
 # Connext 5.3.1 for Galactic and earlier.
-RUN if [ "$ROS_DISTRO" != "galactic" -o "$ROS_DISTRO" != "foxy" ]; then \
+RUN if [ "$ROS_DISTRO" = "galactic" -o "$ROS_DISTRO" = "foxy" ]; then \
       cd /tmp && curl --silent https://s3.amazonaws.com/RTI/Bundles/5.3.1/Evaluation/rti_connext_dds_secure-5.3.1-eval-x64Linux3gcc5.4.0.tar.gz | tar -xz && \
       tar -xvf /tmp/openssl-1.0.2n-target-x64Linux3gcc5.4.0.tar.gz; \
     fi
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -89,18 +89,23 @@ RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-insta
 RUN apt-get update && apt-get install --no-install-recommends -y python3-pexpect
 
 # Get and install the RTI web binaries.
-# Connext 5.3.1 for focal and earlier.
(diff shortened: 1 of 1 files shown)


linux_docker_resources/Dockerfile
```diff
@@ -89,18 +89,23 @@ RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-insta
 RUN apt-get update && apt-get install --no-install-recommends -y python3-pexpect
 
 # Get and install the RTI web binaries.
-# Connext 5.3.1 for focal and earlier.
-RUN if test ${UBUNTU_DISTRO} != jammy; then cd /tmp && curl --silent https://s3.amazonaws.com/RTI/Bundles/5.3.1/Evaluation/rti_connext_dds_secure-5.3.1-eval-x64Linux3gcc5.4.0.tar.gz | tar -xz; fi
-RUN if test ${UBUNTU_DISTRO} != jammy; then cd /tmp && tar -xvf /tmp/openssl-1.0.2n-target-x64Linux3gcc5.4.0.tar.gz; fi
+# Connext 5.3.1 for Galactic and earlier.
+RUN if [ "$ROS_DISTRO" != "galactic" -o "$ROS_DISTRO" != "foxy" ]; then \
```
clalancette, line 93: Isn't this logic inverted?  Shouldn't this be `$ROS_DISTRO = "foxy" -o "$ROS_DISTRO = "galactic"` ?

linux_docker_resources/Dockerfile
```diff
@@ -89,18 +89,23 @@ RUN if test ${PLATFORM} = x86; then apt-get update && apt-get install --no-insta
...
 # Connext 6.0.1 for jammy, the evaluation bundles don't contain security extensions so we need to distribute the pro binaries to ourselves.
 COPY rticonnextdds-src/ /tmp/rticonnextdds-src
-RUN for splitpkg in \
-  /tmp/rticonnextdds-src/rti_connext_dds-6.0.1-pro-host-x64Linux.run \
-  /tmp/rticonnextdds-src/rti_connext_dds-6.0.1.25-pro-host-x64Linux.rtipkg \
-  /tmp/rticonnextdds-src/rti_connext_dds-6.0.1.25-pro-target-x64Linux4gcc7.3.0.rtipkg; do \
-    cat $(echo ${splitpkg}.0?? | sort) > $splitpkg; \
-  done
-RUN chmod 755 /tmp/rticonnextdds-src/rti_connext_dds-6.0.1-pro-host-x64Linux.run
+RUN if [ "$ROS_DISTRO" != "galactic" -o "$ROS_DISTRO" != "foxy" ]; then \
```
clalancette, line 99: Also, shouldn't this be `$ROS_DISTRO != "foxy" -a $ROS_DISTRO != "galactic"`?

clalancette (changes requested):

linux_docker_resources/Dockerfile
nuclearsandwich: You are correct. These are what happens when you write conditionals late at night.
```suggestion
RUN if [ "$ROS_DISTRO" = "galactic" -o "$ROS_DISTRO" = "foxy" ]; then \
```

linux_docker_resources/Dockerfile
nuclearsandwich:
```suggestion
RUN if [ "$ROS_DISTRO" != "galactic" -a "$ROS_DISTRO" != "foxy" ]; then \
```

clalancette (approved):
Looks good to me with green CI on Rolling and at least one of Galactic or Foxy.

clalancette:
We removed Connext 5 support when we removed Foxy support, so closing this one out.

PR #667, opened 2022-06-08
Opt-out of Phased Updates in apt.

nuclearsandwich:
Docker containers are built using a succession of images, each of of which is created by running an intermediate container and saving the result.

These individual containers can end up with different information that puts them in different "phases" of the apt phased update rollout.

When combined with a phased update for a set of binary packages that all rely on matching versions, it can create an unsatisfiable condition where an earlier binary from the source package was installed in the "ahead" phase and a later package is installed "behind".

Phased updates only became supported in apt 2.1.16 released in Ubuntu 21.04 so this was not an issue for us prior to Ubuntu 22.04.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -11,6 +11,10 @@ ARG COMPILE_WITH_CLANG=false
 # See: http://askubuntu.com/questions/506158/unable-to-initialize-frontend-dialog-when-using-ssh
 ENV DEBIAN_FRONTEND noninteractive
 
+# Opt-out of phased updates, which can create inconsistencies between installed package versions as different containers end up on different phases.
+# https://wiki.ubuntu.com/PhasedUpdates
+RUN echo 'APT::Get::Never-Include-Phased-Updates: True;' > /etc/apt/apt.conf.d/90-phased-updates
+
 RUN apt-get update && apt-get install --no-install-recommends -y locales
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
:penguin: [Build Status](https://ci.ros2.org/job/test_ci_linux/119/)
:penguin: :mechanical_arm:  [Build Status](https://ci.ros2.org/job/test_ci_linux-aarch64/15/)

nuclearsandwich:
I haven't gone back to find the record yet but were apt installation issues on ARM hosts earlier which we thought at the time were problems with apt mirror synchronization but may actually have been an earlier less egregious version of this problem.

nuclearsandwich:
> penguin [Build Status](https://ci.ros2.org/job/test_ci_linux/119/)
> penguin mechanical_arm [Build Status](https://ci.ros2.org/job/test_ci_linux-aarch64/15/)

These results (50:50) are not encouraging since any time we bust the docker cache we have a chance of landing all "behind" or all "ahead" and being able to build.

So something about this change is insufficient and I kept "getting lucky" when testing it locally.

nuclearsandwich:
:penguin: [Build Status](https://ci.ros2.org/job/test_ci_linux/120/)
:penguin: :mechanical_arm:  [Build Status](https://ci.ros2.org/job/test_ci_linux-aarch64/16/)

Blast545 (approved):

Blast545:
LGTM after reading a bit on this:

https://wiki.ubuntu.com/StableReleaseUpdates#Phasing
https://askubuntu.com/questions/1246962/what-apt-configuration-file-should-the-never-include-phased-updates-flag-be-pl

PR #671, opened 2022-07-12
Start nightly jobs at 4:00 UTC (21:00 PDT).

nuclearsandwich:
Currently nightly jobs start at 7:30 UTC (00:30 PDT) but as the job
counts and runtimes grow, this has started to spill over into the early
morning.

Additionally, the build farmer report currently runs at 17:00 UTC (10:00
PDT) and the last builds are frequently still in progress when the
report generates, meaning that feedback for a job can be delayed up to
24 hours since it won't be scanned until the next build report.

Moving the jobs 3.5 hours earlier should enable them to complete before
the build report is generated, improving its utility.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -41,7 +41,7 @@ except ImportError:
 
 DEFAULT_REPOS_URL = 'https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos'
 DEFAULT_MAIL_RECIPIENTS = 'ros2-buildfarm@googlegroups.com'
-PERIODIC_JOB_SPEC = '30 7 * * *'
+PERIODIC_JOB_SPEC = '0 3 * * *'
 
 template_prefix_path[:] = \
     [os.path.join(os.path.abspath(os.path.dirname(__file__)), 'job_templates')]


clalancette (commented):
This sounds completely reasonable to me.  But I will note that this may interfere more with the workday in Asia, so pinging @ros2/ci-builders and @ros2/team for comment.

audrow (approved):

Blast545 (approved):
LGTM, happy to hear from the Asia team before this is merged JIC.

gbiggs:
21:00 PDT is OK with me, it's midday for east Asia and I've usually finished my package maintenance work by then.

clalancette (approved):

tfoote:
Historically there was a router reset on the HQ network that we made sure all the nightlies started after. I think that has been resolved with our improved networking hardware. But we should double check with @nuclearsandwich before he is away.

nuclearsandwich:
> Historically there was a router reset on the HQ network that we made sure all the nightlies started after. I think that has been resolved with our improved networking hardware.

Yeah we discussed that in the ROS 2 meeting this week as one of the reasons but the infrastructure we've had since this past winter no longer requires a nightly reset.

nuclearsandwich:
In an extremely Friday turn of events, although I updated the PR description to use 4am UTC I did not push the updated commit so the change that I just merged, and subsequently deployed, was for 3am UTC.

nuclearsandwich:
Fixed with https://github.com/ros2/ci/pull/673

PR #673, opened 2022-07-15
Correct disagreement between commit message and reality.

nuclearsandwich:
The previous commit message was updated to reflect the intended change to 4am UTC but the commit remained at the previously intended target of 3am.


nuclearsandwich:
Merging this since it was what everyone meant to approve anyway.

PR #674, opened 2022-07-27
Install python3-pytest-benchmark

sloretz:
For this PR https://github.com/ros2/rclpy/pull/973


clalancette:
What's the story with Windows here?  Do we need to pip install it there?

sloretz:
> What's the story with Windows here? Do we need to pip install it there?

We could, though as far as I know the [only job running benchmarks is on Linux](https://build.ros2.org/view/Rci/job/Rci__benchmark_ubuntu_jammy_amd64/). I haven't tried them on Windows.

PR #675, opened 2022-08-01
Pin the versions of flake8.

clalancette:
This also caused us to refactor how we are doing the discovery
of pip packages.  The new version is actually shorter.


clalancette:
Here's some starting CI, just up to rcutils, to make sure things are working as expected.  We'll greatly expand what we test from there:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17046/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11607/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17471/)

clalancette:
OK, here is a similar test on Foxy:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17053/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11614/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17478/)

ros2_batch_job/__main__.py
```diff
@@ -585,11 +587,18 @@ def need_package_from_pipy(pkg_name):
...
             # On Windows since we switch between the debug and non-debug
             # interpreter all packages need to be reinstalled too
             pip_cmd.append('--force-reinstall')
+        if args.do_venv:
+            # If we are in a virtualenv, we currently expect that the vcs tool
+            # is as well.  Force install it here, even if we would have inherited
+            # it from the environment.
+            job.run(
+                ['"%s"' % job.python, '-m', 'pip', 'install', '-U', '--force-reinstall', 'vcstool'],
+                shell=True)
```
clalancette, line 600: This is seemingly a weird change to include here, but there is a reason.

During testing, I found that anytime we ran `pip install -U --force-reinstall flake8-builtins`, it would *always* upgrade `flake8` from pip, even if it could have gotten it from the operating system.  I'm honestly not sure why that is the case, but this works around it.  Since on Linux we use a new virtualenv every time, this should be safe to do.

clalancette: Ah, never mind all that.  I just found out that `pip` has a `--upgrade-strategy only-if-needed` option, which is probably what we want.

clalancette:
Here's another set of complete runs, which should hopefully fix the RHEL issue:


Rolling (all packages):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17064/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17489/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/17490/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/246/)

Foxy (all packages):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17065/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11625/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17491/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/17492/)

clalancette:
This did not end up working well at all, as you can see by all of the yellow and red above.  Closing this in favor of #677 , which I think will work better.

PR #678, opened 2022-09-20
Add manually triggered asan/tsan jobs

jacobperron:
Useful for triggering on PRs targeting asan/tsan issues and for running on older support platforms (e.g. Ubuntu 20.04).

I'm proposing this versus manually triggering the nightly jobs, thus polluting the job history.

Let me know if there's something else that needs to be done to make these new jobs available on ci.ros2.org.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -291,7 +291,6 @@ def main(argv=None):
                 'test_args_default': data['test_args_default'] + ' --packages-up-to rcpputils',
             })
 
-
         # configure jobs for compiling with clang+libcxx on linux
         if os_name == 'linux':
             # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
@@ -339,7 +338,6 @@ def main(argv=None):
                 'build_args_default': tsan_build_args,
                 'test_args_default': data['test_args_default'] + ' --packages-select rcpputils rcutils',
             })
-
 
         # configure a manually triggered version of the coverage job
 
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> I'm proposing this versus manually triggering the nightly jobs, thus polluting the job history.

If this is just for convenience I have reservations.

I appreciate not polluting job history. If there was a straightforward way to _block_ manually triggering the nightly jobs I'd put that on. There really isn't a circumstance where it is required.

All of the asan/tsan jobs are reproducible with the current manual CI jobs by setting specific parameters.

If we add another manual job then we have to store yet another set of job histories and add to the growing list of jobs that CI builders need to avoid in most cases. It would be _extra_ weird if a manual CI job defaults to a non-default platform for specific job types just because.

If we want to make triggering these manually easier then I can draft you a python script that will trigger the standard manual CI job with a given set of parameters as long as you have a Jenkins token in your environment.

---

If instead there's a desire to both trigger these manually _and_ have them periodically then we would need periodic jobs rather than just additional manual CI jobs and unless there are blockers to running such jobs on build.ros2.org I'd like to try and push new CI jobs there unless they specifically need Windows.

jacobperron:
That's fair. If all we need to do is copy the relevant build/test arguments into the existing manual CI job, then I think that's a reasonable approach.

jacobperron:
Giving it a try on Ubuntu Focal: [Build Status](https://ci.ros2.org/job/ci_linux/17336/)

jacobperron:
Seems to have produced a similar result as the nightly. I'm satisfied :slightly_smiling_face:

PR #679, opened 2022-09-26
Update RTI Connext DDS license.

nuclearsandwich:
A current license file is required for running RTI Connext DDS Pro in the Security tests.


clalancette (approved):
Looks reasonable to me with a green CI run on Linux and Windows.  Should we also run Foxy and Galactic CI to check that they work too?

nuclearsandwich:
Building up to rmw_connextdds and test_security and running just test_security. @clalancette do you think it is worth a full CI run as well? I'm not so sure but I don't want to be cavalier.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11918/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17888/)

clalancette:
> @clalancette do you think it is worth a full CI run as well? I'm not so sure but I don't want to be cavalier.

For cases like this, I usually opt to be conservative and test everything.  That way there should be no surprises on the nightly CI.  The downside, of course, is that we are probably blocking @jacobperron in the meantime, but hopefully only for 24 hours.

nuclearsandwich:
Full CI

## Rolling
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17396/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11928/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17901/)

jacobperron (approved):

PR #680, opened 2022-09-30
Retry RTI Connext installation on Windows

nuclearsandwich:
We're seeing 7-13% of windows builds failing while connext is being installed. As part of our investigation into these issues we're trying to determine how intermittent they are and whether retrying the installation on failure will be at all effective in resolving the issue.

Accompanying ros2-cookbooks PR: https://github.com/ros-infrastructure/ros2-cookbooks/pull/48


clalancette (approved):

nuclearsandwich:
Superceded by #681 which includes these changes along with additional changes fixing Qt installation.

PR #681, opened 2022-10-07
Update recipe for qt download path.

nuclearsandwich:
Qt has moved the 5.12.10 offline installer that we rely on for Windows.


clalancette (commented):
The change looks good to me, but I think we need a PR to ros2-cookbooks, as well, correct?

nuclearsandwich:
https://github.com/ros-infrastructure/ros2-cookbooks/pull/49

nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_windows/18003/) is a rebuild of a job which failed previously and is now past the infra part of the build.

clalancette (approved):

nuclearsandwich:
This PR now includes #680 since the changes here are required for those builds to succeed.

PR #687, opened 2023-01-02
Remove all references to Galactic.

clalancette:

I'm not sure if we want to merge this right away, or wait a little bit just in case we need to take Galactic out of cold storage.  @cottsay @nuclearsandwich Thoughts?


cottsay (approved):

nuclearsandwich (approved):
We no longer have a need to keep Galactic jobs running.

This looks like a comprehensive scan of Galactic mentions and I think that any remaining Galactic focused behavior will apply equally to Foxy.

clalancette:
Thanks for the reviews.  I've done a test deploy, and everything looks as expected.  So I'm going to merge this in and do a real deploy.

PR #690, opened 2023-01-27
Install LTTng and related packages on Linux

christophebedard:
Part of https://github.com/ros2/ros2/issues/1177

See https://github.com/ros2/ros2_tracing/pull/31

Signed-off-by: Christophe Bedard <christophe.bedard@apex.ai>

### DIFF
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -113,7 +113,7 @@ RUN dnf install \
     libyaml-devel \
     libzstd-devel \
     $(if [ "$ROS_DISTRO" = "foxy" ]; then echo log4cxx-devel; fi) \
-    $(if test ${EL_RELEASE} != 8; then echo lttng-ust; fi) \
+    $(if test ${EL_RELEASE} != 8; then echo lttng-ust-devel; fi) \
     mesa-libGL-devel \
     mesa-libGLU-devel \
     opencv-devel \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -113,6 +113,7 @@ RUN dnf install \
     libyaml-devel \
     libzstd-devel \
     $(if [ "$ROS_DISTRO" = "foxy" ]; then echo log4cxx-devel; fi) \
+    lttng-ust \
     mesa-libGL-devel \
     mesa-libGLU-devel \
     opencv-devel \
@@ -133,6 +134,7 @@ RUN dnf install \
     python3-importlib-metadata \
     $(if test ${EL_RELEASE} = 8; then echo python3-importlib-resources; fi) \
     python3-lark-parser \
+    $(if test ${EL_RELEASE} != 8; then echo python3-lttng; fi) \
     python3-lxml \
     $(if test ${EL_RELEASE} != 9; then echo python3-matplotlib; fi) \
     $(if test ${EL_RELEASE} = 8; then echo python3-mock; fi) \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -113,6 +113,7 @@ RUN dnf install \
     libyaml-devel \
     libzstd-devel \
     $(if [ "$ROS_DISTRO" = "foxy" ]; then echo log4cxx-devel; fi) \
+    $(if test ${EL_RELEASE} != 8; then echo lttng-tools; fi) \
     $(if test ${EL_RELEASE} != 8; then echo lttng-ust-devel; fi) \
     mesa-libGL-devel \
     mesa-libGLU-devel \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -113,7 +113,7 @@ RUN dnf install \
     libyaml-devel \
     libzstd-devel \
     $(if [ "$ROS_DISTRO" = "foxy" ]; then echo log4cxx-devel; fi) \
-    lttng-ust \
+    $(if test ${EL_RELEASE} != 8; then echo lttng-ust; fi) \
     mesa-libGL-devel \
     mesa-libGLU-devel \
     opencv-devel \
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -54,6 +54,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   libignition-cmake2-dev \
   libignition-math6-dev \
   liblog4cxx-dev \
+  liblttng-ust-dev \
   liborocos-kdl-dev \
   libspdlog-dev \
   libsqlite3-dev \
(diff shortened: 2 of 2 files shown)


linux_docker_resources/Dockerfile-RHEL
```diff
@@ -133,6 +134,7 @@ RUN dnf install \
     python3-importlib-metadata \
     $(if test ${EL_RELEASE} = 8; then echo python3-importlib-resources; fi) \
     python3-lark-parser \
+    $(if test ${EL_RELEASE} != 8; then echo python3-lttng; fi) \
```
christophebedard, line 137: I think `lttng-ust` should work, but I'm not completely sure about `python3-lttng`.

cottsay: RHEL 8 has `python3-lttng`, but the problem is that it's too old and tracetools breaks if it's present. I removed it in #657. Until tracetools handles the old version gracefully (which I'd still very much like to see), this is the best workaround.

clalancette:
> I can do `${UBUNTU_DISTRO} = jammy` and `${ROS_DISTRO} != humble` for Ubuntu and `${EL_RELEASE} != 8` for RHEL for all these new packages.

Yeah, that is probably a good idea.  I'll also suggest putting this in a separate block in the Ubuntu Dockerfile, similar to https://github.com/ros2/ci/blob/7cbb9f6d57c41cac15fa8fd2b8020d2f7fe388de/linux_docker_resources/Dockerfile#L97

christophebedard:
> I'll also suggest putting this in a separate block in the Ubuntu Dockerfile

Done!

clalancette (approved):
This looks good to me, though I'll want to see Linux CI for all of Foxy, Humble, and Rolling on both Ubuntu and RHEL.  I'd also like to get one more approval from @cottsay .

linux_docker_resources/Dockerfile-RHEL
```diff
@@ -113,6 +113,7 @@ RUN dnf install \
     libyaml-devel \
     libzstd-devel \
     $(if [ "$ROS_DISTRO" = "foxy" ]; then echo log4cxx-devel; fi) \
+    $(if test ${EL_RELEASE} != 8; then echo lttng-ust; fi) \
```
cottsay, line 116: As the Debian side is getting the development files for `lttng-ust`, the RHEL side should as well.
```suggestion
    $(if test ${EL_RELEASE} != 8; then echo lttng-ust-devel; fi) \
```

The rosdep DB already reflects this properly: https://github.com/ros/rosdistro/blob/64bff7eb4ad6e18e0374219abb252a37462a94ae/rosdep/base.yaml#L4161

cottsay (changes requested):
One quick change

cottsay (approved):
Thanks!

clalancette:
OK, I've pushed an exact copy of this branch to the origin.  Here are some CI runs with that in place:

Foxy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18352/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12881/)
* RHEL-8 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/353/)

Humble:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18353/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12882/)
* RHEL-8 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/354/)

Rolling:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18354/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12883/)
* RHEL-9 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/355/)

christophebedard:
I thought we didn't need `lttng-tools` (which provides the CLI), but I forgot that I added it as an `<exec_depend>` of `tracetools` and that we need it in `tracetools_trace`: https://github.com/ros2/ros2_tracing/blob/e664e5bbd28b42da6430623fc1bcf644d6b6a449/tracetools_trace/tracetools_trace/tools/lttng.py#L90-L94. I'll add it as exec and test dependency of `tracetools_trace` in https://github.com/ros2/ros2_tracing/pull/31 and add it here.

cottsay (approved):

christophebedard:
I also noticed that there was no rosdep rule for `lttng-tools` for RHEL, so I added it: https://github.com/ros/rosdistro/pull/36641

clalancette:
All right.  I'm only going to run the latest change against Rolling again, given that it is protected under the same checks as the previous code.

clalancette:
Rolling:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18357/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12887/)
* RHEL-9 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/356/)

christophebedard:
The test failures on RHEL don't seem to be related, but I can't find the same failures in recent jobs.

clalancette:
> The test failures on RHEL don't seem to be related, but I can't find the same failures in recent jobs.

That one is a flaky test; you can see a failure of it in https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_repeated/1448/#showFailuresLink .  I really don't think it is involved here at all.

So I think we are good with this change.

Once we merge this in, that means that we'll be building tracepoints in by default for Rolling in our nightly CI jobs and our nightly packaging (tarball) jobs.  What needs to happen to have it enabled in the debians?  Do we just need to do a release of `ros2_tracing` for that to happen?

christophebedard:
> Once we merge this in, that means that we'll be building tracepoints in by default for Rolling in our nightly CI jobs and our nightly packaging (tarball) jobs. What needs to happen to have it enabled in the debians? Do we just need to do a release of `ros2_tracing` for that to happen?

Specifically after merging https://github.com/ros2/ros2_tracing/pull/31 (which I think we should merge right after this), yes! LTTng will be picked up as a dependency, and the tracepoints will be compiled in.

Well, that's what I think will happen, anyway. I can check the testing repo and validate once it's re-built.

christophebedard:
Next step before merging would be to run CI using this branch and https://github.com/ros2/ros2_tracing/pull/31.

clalancette:
> Next step before merging would be to run CI using this branch and [ros2/ros2_tracing#31](https://github.com/ros2/ros2_tracing/pull/31). See [ros2/ros2_tracing#31 (comment)](https://github.com/ros2/ros2_tracing/pull/31#issuecomment-1489374096)

Awesome, the jobs there look good.

So what is the next step here?  Do you want to merge the ros2_tracing PR, then I can merge this one?

christophebedard:
> So what is the next step here? Do you want to merge the ros2_tracing PR, then I can merge this one?

Since packages will now depend on LTTng, and since CI doesn't seem to run `rosdep install` (except for RHEL, it seems), we should merge this first. Then I'll merge https://github.com/ros2/ros2_tracing/pull/31 right after.

clalancette:
> Since packages will now depend on LTTng, and since CI doesn't seem to run `rosdep install` (except for RHEL, it seems), we should merge this first. Then I'll merge [ros2/ros2_tracing#31](https://github.com/ros2/ros2_tracing/pull/31) right after.

OK, got it.

So, the only thing is that I'm on PTO, not checking email, starting about now until Monday (April 3).  Despite the fact that I think we've done sufficient testing here, I'm somewhat reluctant to merge something like this and disappear.  So either we should have @cottsay merge it and keep an eye on it tomorrow, or I can do it first thing Monday morning when I'm back.  @cottsay what do you think?

christophebedard:
Waiting until Monday as a precaution is fine with me.

PR #695, opened 2023-03-02
Install python3-jsonschema as Linux build requirement

emersonknapp:
Requirement from https://github.com/ros2/rosidl/pull/722

See https://github.com/ros2/rosidl/pull/722#discussion_r1120569055

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -69,6 +69,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-coverage \
   python3-cryptography \
   python3-flake8 \
+  python3-jsonschema \
   python3-lxml \
   python3-mock \
   python3-mypy \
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):
We'll need an additional PR for ROS 2 CI once the cookbook is updated. But this is good for Linux support.

We do need to double check that this package is available in RHEL 9, which we're getting ready to move to for Rolling and Iron.

clalancette:
> We do need to double check that this package is available in RHEL 9, which we're getting ready to move to for Rolling and Iron.

It seems to be available in the default `appstream` repositories for AlmaLinux 9, at least.  @cottsay will that translate to being available in RHEL-9 as well?

cottsay (approved):
> will that translate to being available in RHEL-9 as well?

It is provided by AppStream in both RHEL 8 and RHEL 9, so this looks good.

clalancette:
Given the latest comments, I'm going to run a quick CI for this on both Ubuntu and RHEL, then merge.

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18171/)
* RHEL 8 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/338/)
* RHEL 9 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/339/)

PR #697, opened 2023-04-04
Make the RHEL packaging job respect EL_RELEASE.

clalancette:
It wasn't doing that before, so was always defaulting to RHEL-8.

@Crola1702 FYI, this should in theory fix https://github.com/osrf/buildfarmer/issues/401


clalancette:
CI:
* RHEL [Build Status](http://ci.ros2.org/job/ci_packaging_linux-rhel/41/)

clalancette:
Note: This is a change to the packaging job template, so a straight CI wouldn't confirm whether this fixed the problem or not.  I manually modified the `ci_packaging_linux-rhel` job to add in this piece of the template; assuming CI is successful there, I will revert that and redeploy this for real once we merge it in.

clalancette:
And I've now deployed the change.

Issue #702, opened 2023-05-15
`error waiting for container: unexpected EOF`  failing builds 


claraberendsen:
### Description
We have been encountering the error of `error waiting for container: unexpected EOF` frequently on ci.ros2.org. Seems to be an issue with memory availability for the container.
### Example build
https://ci.ros2.org/view/nightly/job/nightly_linux_debug/2664/console 
### Statistics on the issue for the first two weeks of May: 

nightly_linux_debug: 20% of the times
nightly_linux-aarch64_release: 13.33% of the times
nightly_linux-aarch64_repeated: 13.33% of the times
nightly_linux_release: 7.14% of the times

Crola1702:
### Container EOF error statistics

|job_name|last_fail|first_fail|build_count|failure_count|failure_percentage|
| -- | -- | -- | -- | -- | -- |
|nightly_linux_debug|2023-05-19|2023-05-10|16|5|31.25|
|nightly_linux_foxy_coverage|2023-05-18|2023-05-07|16|2|12.5|
|nightly_linux-rhel_repeated|2023-05-18|2023-05-06|21|5|23.81|
|nightly_linux-rhel_release|2023-05-18|2023-05-12|15|4|26.67|
|nightly_linux-rhel_debug|2023-05-18|2023-05-06|15|3|20.0|
|Humble cyclonedds|2023-05-18|2023-05-09|6|2|33.33|
|Humble connext|2023-05-18|2023-05-06|5|2|40.0|
|Foxy connext|2023-05-14|2023-05-14|5|1|20.0|
|Rolling cyclonedds|2023-05-13|2023-05-13|17|1|5.88|
|Rolling connext|2023-05-13|2023-05-13|16|1|6.25|
|nightly_linux_release|2023-05-12|2023-05-12|24|1|4.17|
|nightly_linux-aarch64_release|2023-05-11|2023-05-10|15|2|13.33|
|nightly_linux-aarch64_repeated|2023-05-10|2023-05-10|15|1|6.67|

claraberendsen:
Today two aarch64 nightlies failed with this issue: [nightly_linux-aarch64_debug #2384](https://ci.ros2.org/job/nightly_linux-aarch64_debug/2384/) and [nightly_linux-aarch64_repeated #2371](https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/2371/). Since the agent didn't get destroyed afterwards the logs were preserved and a detailed investigation was possible. 
### Investigation 
[...]

PR #703, opened 2023-05-18
Add recovery strategy for ` error waiting for container: unexpected EOF` 

claraberendsen:
### Description
This PR adds the functionality for recovering the agent by restarting it when an specific error occurs on the log.  In particular the case that was applied is [`error waiting for container: unexpected EOF`](https://github.com/ros2/ci/issues/702).

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -533,8 +533,7 @@ def main(argv=None):
         create_job(os_name, job_name, 'ci_job.xml.em', {
             'cmake_build_type': 'Release',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
-            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,              
-
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
         })
 
         # configure nightly triggered job with repeated testing
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -217,6 +217,11 @@ def main(argv=None):
             'cmake_build_type': 'None',
         })
 
+        create_job(os_name, 'test_ci_recovery_' + os_name, 'ci_job.xml.em', {
+            'cmake_build_type': 'None',
+            'with_eof_recovery': True
+        })
+
         if os_name == 'windows-metal':
             # Don't create nightlies or packaging jobs for bare-metal Windows
             continue
@@ -528,7 +533,8 @@ def main(argv=None):
         create_job(os_name, job_name, 'ci_job.xml.em', {
             'cmake_build_type': 'Release',
             'time_trigger_spec': PERIODIC_JOB_SPEC,
-            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,
+            'mailer_recipients': DEFAULT_MAIL_RECIPIENTS,              
+
         })
 
         # configure nightly triggered job with repeated testing
--- a/job_templates/scripts/add_recovery_label_on_regex.groovy
+++ b/job_templates/scripts/add_recovery_label_on_regex.groovy
@@ -1,0 +1,28 @@
+  if(!regex) {
+    return """println('WARNING! Incorrect recovery configuration : No regex provided as search key for log')"""
+  }
+  return """\
+    import hudson.model.*
+    def node = build.getBuiltOn()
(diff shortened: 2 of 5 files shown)


claraberendsen:
@nuclearsandwich I have a draft of the process here. I need some insight on how to test this... I have a separate PR that I think should go first to this one, that adds the necessary plugins for this to work. 
At the moment I created a test job in the create_jobs.py file to test that the configuration is correct. However I'm not certain on how to build that and run it, since I don't know when and where is this file being run. Would appreciate your insight.

nuclearsandwich:
> I have a separate PR that I think should go first to this one, that adds the necessary plugins for this to work.

Can you link / reference that here. As discussed in the Infrastructure meeting we're blocked on Jenkins plugin upgrades generally but that's an important item to address in the near future.

> I need some insight on how to test this...

1. Get this draft ready for a test deployment from the current branch
2. Deploy this PR scoped just to test_ci_linux
3. Manually reconfigure test_ci_linux to echo the expected error string
4. Run test_ci_linux and validate the behavior
5. Re-deploy to remove the manual reconfiguration and run an (expected) successful job to confirm that nothing breaks when the job doesn't exhibit the failure.

I'm in favor of just hand-hacking the test_ci_linux job to display the error string and exit, rather than adding a separate test job just for this.
The jobs get re-configured on every deploy and I don't think there's currently enough of a justification to create job for doing those kinds of tests.

> At the moment I created a test job in the create_jobs.py file to test that the configuration is correct. However I'm not certain on how to build that and run it, since I don't know when and where is this file being run. Would appreciate your insight.

This script is generally run by ROS 2 devs locally after merging a PR and making a change. Running it requires Jenkins admin configuration and a configured python environment. There are details and links to further details in the [readme](https://github.com/ros2/ci#setting-up-on-a-jenkins-build-farm).

create_jenkins_job.py
```diff
@@ -217,6 +217,11 @@ def create_job(os_name, job_name, template_file, additional_dict):
             'cmake_build_type': 'None',
         })
 
+        create_job(os_name, 'test_ci_recovery_' + os_name, 'ci_job.xml.em', {
+            'cmake_build_type': 'None',
+            'with_eof_recovery': True
+        })
```
nuclearsandwich, line 223: As mentioned in https://github.com/ros2/ci/pull/703#issuecomment-1558027258 I'm in favor of hand-hacking `test_ci_linux` rather than creating a one-time job for testing.

job_templates/scripts/add_recovery_label_on_regex.groovy
```diff
@@ -0,0 +1,25 @@
...
+         println(' PROBLEM: ${error} was detected in the log. Try to automatically resolve it:')
+         println("Removing labels and adding 'recovery-process' label to node")
+         node.setLabelString('recovery-process')
+         } catch (Exception ex) {
+         println('ERROR - CANNOT PERFORM RECOVERY ACTIONS FOR NVIDIA ERROR')
+         println('Restoring to previous state') node.setLabelString(oldLabels)
+         throw ex
+     }
+    }
+    println('# END SECTION: NVIDIA MISMATCH RECOVERY')
```
nuclearsandwich, line 24: This looks like it was copy-pasted from the build.osrfoundation.org recovery without updating the section output.

I'm also not sure that this script needs to be wrapped in a script since the Groovy DSL isn't used.

job_templates/snippet/system_groovy_script_block.xml.em
```diff
@@ -0,0 +1,15 @@
...
+          <script>@ESCAPE(command)</script>
+          <sandbox>false</sandbox>
+        </script>
+      </source>
+@[end if]@
+@[if script_file]@
+      <source class="hudson.plugins.groovy.FileSystemScriptSource">
+        <scriptFile>@ESCAPE(script_file)</scriptFile>
+      </source>
+@[end if]@
```
nuclearsandwich, line 14: Was this snippet copied from ros_buildfarm? Is there an anticipated use case for this branch of the snippet?

nuclearsandwich (commented):
I did a first pass review in addition to the general feedback and attempts to answer the raised questions.

claraberendsen:
Closing this since we have tracked the actual error and this strategy is not the appropriate solution

PR #706, opened 2023-05-30
Update cookbook version to fix Qt download path.

nuclearsandwich:
Pulls in https://github.com/ros-infrastructure/ros2-cookbooks/pull/55


claraberendsen (approved):

nuclearsandwich:
This isn't a change that requires deployment since it does not change the Jenkins xml configuration. So the next Windows jobs should run with this cookbook.

PR #27, opened 2017-02-15
add the --version option to 'brew list'

wjwwood:
This will print the installed versions along with the packages which are installed.

PR #708, opened 2023-06-01
ARM agents fail because of mismatch on phased updates. 

claraberendsen:
### Description
ci.ros2.org arm agents have had to be put offline because of issues with phased updates even though they were disabled since #667. It seems given this post by the ubuntu team that the name for the file is not `90-phased-updates` but rather `99-phased-updates` in ubuntu jammy. 

### Testing
[Build Status](https://ci.ros2.org/job/test_ci_linux-aarch64/21/)
In this build with the change in the configuration using this branch the issue seems to be resolved.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -16,7 +16,7 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN echo 'APT::Get::Never-Include-Phased-Updates "true";' > /etc/apt/apt.conf.d/99-phased-updates
 
 RUN apt-get update && apt-get install --no-install-recommends -y locales
-RUN locale-gen en_US.UTF-89
+RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 
 # net-tools is for ifconfig
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -13,10 +13,10 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Opt-out of phased updates, which can create inconsistencies between installed package versions as different containers end up on different phases.
 # https://wiki.ubuntu.com/PhasedUpdates
-RUN echo 'APT::Get::Never-Include-Phased-Updates "true";' > /etc/apt/apt.conf.d/90-phased-updates
+RUN echo 'APT::Get::Never-Include-Phased-Updates "true";' > /etc/apt/apt.conf.d/99-phased-updates
 
 RUN apt-get update && apt-get install --no-install-recommends -y locales
-RUN locale-gen en_US.UTF-8
+RUN locale-gen en_US.UTF-89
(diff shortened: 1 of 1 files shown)


Crola1702:
ProTip: Run a build with `--packages-up-to ament_lint` so the time it takes to build is minimal

linux_docker_resources/Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Opt-out of phased updates, which can create inconsistencies between installed package versions as different containers end up on different phases.
 # https://wiki.ubuntu.com/PhasedUpdates
-RUN echo 'APT::Get::Never-Include-Phased-Updates "true";' > /etc/apt/apt.conf.d/90-phased-updates
+RUN echo 'APT::Get::Never-Include-Phased-Updates "true";' > /etc/apt/apt.conf.d/99-phased-updates
```
nuclearsandwich, line 16: Have you tested that this reliably makes a difference somehow? the numeric prefix in these files indicates only precedence and is not significant. Unless there are contents in a present `99-phased-updates` file which are overwriting our declaration in `90-phased-updates` then this change is likely spurious.

claraberendsen: @nuclearsandwich Here are the tests I have done so far: 
- Without this change: [#24](https://ci.ros2.org/job/test_ci_linux-aarch64/24/) - [#25](https://ci.ros2.org/job/test_ci_linux-aarch64/25/)
- With this change: [#27](https://ci.ros2.org/job/test_ci_linux-aarch64/27/console) - [#26](https://ci.ros2.org/job/test_ci_linux-aarch64/26/)

All the documentation I could see from ubuntu always referenced naming the file `99-phased-updates` there is a long discussion [here](https://discourse.ubuntu.com/t/phased-updates-in-apt-in-21-04/20345)  about phased updates in general since 21.04 moving into jammy. 
I can dig a bit further to see why it's working with 99 in front instead of 90 if you need more confidence in this fix.

claraberendsen: ### Further testing for reliability
All of the following tests are passing with the new config
Builds: 
- [#30](https://ci.ros2.org/job/test_ci_linux-aarch64/30/) on same agent as before
- [#31](https://ci.ros2.org/job/test_ci_linux-aarch64/31) on a different agent 
- [#32](https://ci.ros2.org/job/test_ci_linux-aarch64/32) on a different agent
- [#34](https://ci.ros2.org/job/test_ci_linux-aarch64/34/) on a different agent

Blast545 (approved):

claraberendsen:
Closing after discussion that this PR doesn't solve anything in particular. The reason it "fixed" the problem when we had it is because it was invalidating the docker cache and the problem with the library was resolved.  The linked PR is preferred over this one to track remaining tasks.

PR #711, opened 2023-06-15
Change configuration for clang-libcxx jobs.

clalancette:
There are 2 changes:
1. Make sure to force building vendor packages.  That way all packages are built by clang and can be successfully linked.
2. Skip the RQT packages.  They are failing to build for reasons I don't quite understand, but skipping them is better than failing the build completely.

Leaving as a draft until I can get a yellow CI run with this in place.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -299,8 +299,11 @@ def main(argv=None):
 
         # configure jobs for compiling with clang+libcxx on linux
         if os_name == 'linux':
-            # Set the logging implementation to noop because log4cxx will not link properly when using libcxx.
-            clang_libcxx_build_args = data['build_args_default'] + ' --mixin clang-libcxx --packages-skip intra_process_demo'
+            # Make sure to force building vendor packages, since we need these to have been built by
+            # clang as well to link successfully.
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/17/)

clalancette:
All right, with one other fix (https://github.com/ros2/libyaml_vendor/pull/59), I was able to get the job to go from red to yellow.  So I'll change this to a real PR, since we'll need both to get this fully fixed.

Blast545 (approved):
Looks good to me, do we want to start tracking this job? I don't know how often it's run and/or if we want to include them as part of the developer PRs process.

clalancette:
> Looks good to me, do we want to start tracking this job? I don't know how often it's run and/or if we want to include them as part of the developer PRs process. cc: @Crola1702 @claraberendsen

It's currently run nightly, and I think we can just leave it at that.  For now, I don't think it is useful to track this job as it has so many warnings.  Maybe if we get that cleaned up we can add it to the list to be tracked, but I don't have a timeline or that.

PR #712, opened 2023-06-21
Remove all references to the ROS 1 bridge from CI.

clalancette:
Foxy is now End-of-Life, which means we have no active ROS 2 distributions on Focal.  In turn, this means that we have no distributions where we can build the ros1_bridge anymore.  Remove all of the code dealing with the bridge, which includes:

1.  References to "bridge".
2.  References to "ros1".
3.  References to "mixed overlay".

Note that I'm not entirely sure how to test this.  I think if I fire up jobs using this branch, it will probably fail because the templates will be passing variables that the scripts don't understand.  I'm open to ideas on what to do for testing here.

@quarkytale FYI.  If we merge this, then you'll no longer have any way to run CI for `ros1_bridge` maintenance.  But that problem is coming anyway, because we are also going to drop support for Focal.

### DIFF
--- a/CI_BUILDERS.md
+++ b/CI_BUILDERS.md
@@ -16,7 +16,7 @@ Many of the other arguments should be reasonably self-explanatory and its likely
 `CI_BUILD_ARGS` and `CI_TEST_ARGS` are basically passed directly to colcon build and test respectively so you should be familiar with those.
 
 Our CI has limited capacity, and the default `ros2.repos` file is constantly growing.
-To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes.
+To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes. 
 
 ## Nightly and packaging jobs
 
--- a/CI_BUILDERS.md
+++ b/CI_BUILDERS.md
@@ -16,7 +16,7 @@ Many of the other arguments should be reasonably self-explanatory and its likely
 `CI_BUILD_ARGS` and `CI_TEST_ARGS` are basically passed directly to colcon build and test respectively so you should be familiar with those.
 
 Our CI has limited capacity, and the default `ros2.repos` file is constantly growing.
-To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes. 
+To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes.
 
 ## Nightly and packaging jobs
 
--- a/ros2_batch_job/util.py
+++ b/ros2_batch_job/util.py
@@ -192,7 +192,7 @@ class PipProtocol(AsyncSubprocessProtocol):
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
-        # The 'pip install' progress bar output can be pretty noisy in CI output. Print only last line 
+        # The 'pip install' progress bar output can be pretty noisy in CI output. Print only last line
         # every 1000 characters to reduce the noisiness.  We detect the pip progress bar by the beginning sequence of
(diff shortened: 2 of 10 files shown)


nuclearsandwich:
> Note that I'm not entirely sure how to test this. I think if I fire up jobs using this branch, it will probably fail because the templates will be passing variables that the scripts don't understand. I'm open to ideas on what to do for testing here.

A common practice would be to run create_jenkins_jobs.py but scoped only to the test_* jobs. Then run test jobs from this branch.

quarkytale:
Yeah doesn't make sense to add any hacks here to still support `ros1_bridge`, but since there's community interest in the bridge we can find alternatives (eg, Github actions) to help review incoming dev. No point in blocking this, will start another thread.

CI_BUILDERS.md
```diff
@@ -16,7 +16,7 @@ Many of the other arguments should be reasonably self-explanatory and its likely
 `CI_BUILD_ARGS` and `CI_TEST_ARGS` are basically passed directly to colcon build and test respectively so you should be familiar with those.
 
 Our CI has limited capacity, and the default `ros2.repos` file is constantly growing.
-To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes. 
+To save build time limit the scope CI builds to only packages that are changed by the pull request being tested or are likely to be affected by the changes.
```
nuclearsandwich, line 19: This change is spurious and I'd be very tempted to split it out into its own PR. Especially when squashing the PR overall since the commit can neither be marked as solely cosmetic nor specific to a single logical change.

ros2_batch_job/util.py
```diff
@@ -192,7 +192,7 @@ def __init__(self, cmd, exit_on_error, *args, **kwargs):
         AsyncSubprocessProtocol.__init__(self, *args, **kwargs)
 
     def on_stdout_received(self, data):
-        # The 'pip install' progress bar output can be pretty noisy in CI output. Print only last line 
+        # The 'pip install' progress bar output can be pretty noisy in CI output. Print only last line
```
nuclearsandwich, line 195: Same comment as https://github.com/ros2/ci/pull/712/files#diff-225efb36aa7fb8d319b852a0a9d39bdb4bc18b4d3cc8a29a46c576cb6b204fc6R19

nuclearsandwich (approved):
There are two whitespace changes which I think should be split out. Other than that I think this just needs a few smoke tests of the affected job templates before merging.

CI_BUILDERS.md
clalancette: For this particular repository, I actually think these kinds of changes are fine to do inline with other ones (as long as they are small).  That is, it is easy to determine that this is a cosmetic change, and we are never going to be backporting these changes to any other branches.  A PR just to do these small cleanups seems like overkill to me.

That said, this is not a hill I'm going to die on, so I'll go ahead and remove these for now.

clalancette:
All right, I've deployed this PR to all of the `test_*` jobs on https://ci.ros2.org.  Here are some CI runs (only up to ament_cmake_core, since we are testing the infrastructure not the payload here):

CI:
* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/139/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/50/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/312/)
* RHEL [Build Status](http://ci.ros2.org/job/test_ci_linux-rhel/23/)

Packaging:
* Linux [Build Status](http://ci.ros2.org/job/test_packaging_linux/3/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_packaging_linux-aarch64/1/)
* Windows [Build Status](http://ci.ros2.org/job/test_packaging_windows/4/)
* RHEL [Build Status](http://ci.ros2.or/job/test_packaging_linux-rhel/1/)

PR #713, opened 2023-06-22
Remove all references to 'windows-metal'.

clalancette:
It is no longer a build type we have available.

This is kind of a side issue to the main issue of deprecating Foxy, but I figured we should do it while we are here.

PR #714, opened 2023-06-22
Remove support for rmw_connext_cpp from the scripts.

clalancette:
The last distribution that supported this (Foxy) is now EOL.

Unfortunately, removing this is more complicated than I would have liked.  In particular, the job templates were *always* setting '--ignore-rmw rmw_connext_dynamic_cpp', and then adding to that list if other RMWs should be ignored. But we no longer have an RMW that we always want to ignore, so we have to pass --ignore-rmw for each of the ones that we want to ignore (because there may not be any).  To handle that, we had to change the argparse handling of --ignore-rmw to use the 'extend' action, and then we had to modify the entrypoint.sh to deal with the fact that there might be multiple --ignore-rmw flags, any of which could contain 'rmw_connextdds'.

Note that if we merge in https://github.com/ros2/ci/pull/713 first, this one will need to be rebased.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -132,18 +132,8 @@ fi
 if [ "$CI_USE_WHITESPACE_IN_PATHS" = "true" ]; then
   export CI_ARGS="$CI_ARGS --white-space-in sourcespace buildspace installspace workspace"
 fi
-export CI_ARGS="$CI_ARGS --ignore-rmw rmw_connext_dynamic_cpp"
-# TODO(asorbini) `rmw_connext_cpp` is still the default for foxy.
-if [ -n "$CI_ROS_DISTRO" -a \( "$CI_ROS_DISTRO" = foxy \) ]; then
-  if [ "$CI_USE_CONNEXTDDS" = "false" ]; then
-    export CI_ARGS="$CI_ARGS --ignore-rmw rmw_connext_cpp"
-  fi
-else
-  # Always ignore `rmw_connext_cpp` in favor of `rmw_connextdds` for newer releases.
-  export CI_ARGS="$CI_ARGS rmw_connext_cpp"
-  if [ "$CI_USE_CONNEXTDDS" = "false" ]; then
-    export CI_ARGS="$CI_ARGS --ignore-rmw rmw_connextdds"
-  fi
+if [ "$CI_USE_CONNEXTDDS" = "false" ]; then
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_connextdds"
 fi
 if [ "$CI_USE_CYCLONEDDS" = "false" ]; then
   export CI_ARGS="$CI_ARGS --ignore-rmw rmw_cyclonedds_cpp"
@@ -261,24 +251,8 @@ if "!CI_COLCON_BRANCH!" NEQ "" (
 if "!CI_USE_WHITESPACE_IN_PATHS!" == "true" (
   set "CI_ARGS=!CI_ARGS! --white-space-in sourcespace buildspace installspace workspace"
 )
-set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connext_dynamic_cpp"
-:: TODO(asorbini) `rmw_connext_cpp` is still the default for foxy.
-set "CI_CONNEXTDDS_RMW="
-if "!CI_ROS_DISTRO!" NEQ "" (
-  if "!CI_ROS_DISTRO!" == "foxy" (
-    set CI_CONNEXTDDS_RMW=rmw_connext_cpp
-  )
-)
-if "!CI_CONNEXTDDS_RMW!" == "rmw_connext_cpp" (
-  if "!CI_USE_CONNEXTDDS!" == "false" (
-    set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connext_cpp"
-  )
-) else (
-  :: Always ignore `rmw_connext_cpp` in favor of `rmw_connextdds` for newer releases.
-  set "CI_ARGS=!CI_ARGS! rmw_connext_cpp"
-  if "!CI_USE_CONNEXTDDS!" == "false" (
-    set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connextdds"
-  )
+if "!CI_USE_CONNEXTDDS!" == "false" (
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connextdds"
 )
 if "!CI_USE_CYCLONEDDS!" == "false" (
   set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_cyclonedds_cpp"
@@ -367,24 +341,8 @@ if "!CI_ROS_DISTRO!" NEQ "" (
 if "!CI_USE_WHITESPACE_IN_PATHS!" == "true" (
   set "CI_ARGS=!CI_ARGS! --white-space-in sourcespace buildspace installspace workspace"
 )
-set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connext_dynamic_cpp"
-:: TODO(asorbini) `rmw_connext_cpp` is still the default for foxy.
-set "CI_CONNEXTDDS_RMW="
-if "!CI_ROS_DISTRO!" NEQ "" (
-  if "!CI_ROS_DISTRO!" == "foxy" (
-    set CI_CONNEXTDDS_RMW=rmw_connext_cpp
-  )
-)
-if "!CI_CONNEXTDDS_RMW!" == "rmw_connext_cpp" (
-  if "!CI_USE_CONNEXTDDS!" == "false" (
-    set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connext_cpp"
-  )
-) else (
-  :: Always ignore `rmw_connext_cpp` in favor of `rmw_connextdds` for newer releases.
-  set "CI_ARGS=!CI_ARGS! rmw_connext_cpp"
-  if "!CI_USE_CONNEXTDDS!" == "false" (
-    set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connextdds"
-  )
+if "!CI_USE_CONNEXTDDS!" == "false" (
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_connextdds"
 )
 if "!CI_USE_CYCLONEDDS!" == "false" (
   set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_cyclonedds_cpp"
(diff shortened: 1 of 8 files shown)


job_templates/ci_job.xml.em
```diff
@@ -132,27 +132,17 @@ fi
...
-  export CI_ARGS="$CI_ARGS rmw_cyclonedds_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_cyclonedds_cpp"
 fi
 if [ "$CI_USE_FASTRTPS_STATIC" = "false" ]; then
-  export CI_ARGS="$CI_ARGS rmw_fastrtps_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_fastrtps_cpp"
 fi
 if [ "$CI_USE_FASTRTPS_DYNAMIC" = "false" ]; then
-  export CI_ARGS="$CI_ARGS rmw_fastrtps_dynamic_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_fastrtps_dynamic_cpp"
```
nuclearsandwich, line 145: These changes aren't connext related. This does look like a bug fix, since these rmw would only be properly ignored whhen connext is also ignored to provide the initial `--ignore-rmw` argument but I think that this change should be split out into a separate PR.

job_templates/ci_job.xml.em
```diff
@@ -261,33 +251,17 @@ if "!CI_COLCON_BRANCH!" NEQ "" (
...
-  set "CI_ARGS=!CI_ARGS! rmw_cyclonedds_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_cyclonedds_cpp"
 )
 if "!CI_USE_FASTRTPS_STATIC!" == "false" (
-  set "CI_ARGS=!CI_ARGS! rmw_fastrtps_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_fastrtps_cpp"
 )
 if "!CI_USE_FASTRTPS_DYNAMIC!" == "false" (
-  set "CI_ARGS=!CI_ARGS! rmw_fastrtps_dynamic_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_fastrtps_dynamic_cpp"
```
nuclearsandwich, line 264: Same situation as https://github.com/ros2/ci/pull/714/files#diff-9045fc5b0c304d11673ab1a87e5f8395a26511121cb8ea4f23d4362b291f2d0fL149-R145

job_templates/ci_job.xml.em
```diff
@@ -367,33 +341,17 @@ if "!CI_ROS_DISTRO!" NEQ "" (
...
```
nuclearsandwich, line 354: Same situation as https://github.com/ros2/ci/pull/714/files#diff-9045fc5b0c304d11673ab1a87e5f8395a26511121cb8ea4f23d4362b291f2d0fL149-R145

job_templates/packaging_job.xml.em
```diff
@@ -136,27 +136,17 @@ fi
...
-  export CI_ARGS="$CI_ARGS rmw_cyclonedds_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_cyclonedds_cpp"
 fi
 if [ "$CI_USE_FASTRTPS_STATIC" = "false" ]; then
-  export CI_ARGS="$CI_ARGS rmw_fastrtps_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_fastrtps_cpp"
 fi
 if [ "$CI_USE_FASTRTPS_DYNAMIC" = "false" ]; then
-  export CI_ARGS="$CI_ARGS rmw_fastrtps_dynamic_cpp"
+  export CI_ARGS="$CI_ARGS --ignore-rmw rmw_fastrtps_dynamic_cpp"
```
nuclearsandwich, line 149: Same situation as https://github.com/ros2/ci/pull/714/files#diff-9045fc5b0c304d11673ab1a87e5f8395a26511121cb8ea4f23d4362b291f2d0fL149-R145

job_templates/packaging_job.xml.em
```diff
@@ -255,33 +245,17 @@ if "!CI_BRANCH_TO_TEST!" NEQ "" (
...
-  set "CI_ARGS=!CI_ARGS! rmw_cyclonedds_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_cyclonedds_cpp"
 )
 if "!CI_USE_FASTRTPS_STATIC!" == "false" (
-  set "CI_ARGS=!CI_ARGS! rmw_fastrtps_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_fastrtps_cpp"
 )
 if "!CI_USE_FASTRTPS_DYNAMIC!" == "false" (
-  set "CI_ARGS=!CI_ARGS! rmw_fastrtps_dynamic_cpp"
+  set "CI_ARGS=!CI_ARGS! --ignore-rmw rmw_fastrtps_dynamic_cpp"
```
nuclearsandwich, line 258: Same situation as https://github.com/ros2/ci/pull/714/files#diff-9045fc5b0c304d11673ab1a87e5f8395a26511121cb8ea4f23d4362b291f2d0fL149-R145

job_templates/packaging_job.xml.em
```diff
@@ -348,33 +322,17 @@ if "!CI_COLCON_BRANCH!" NEQ "" (
...
```
nuclearsandwich, line 335: Same situation as https://github.com/ros2/ci/pull/714/files#diff-9045fc5b0c304d11673ab1a87e5f8395a26511121cb8ea4f23d4362b291f2d0fL149-R145

linux_docker_resources/entry_point.sh
```diff
@@ -22,13 +22,18 @@ echo "done."
...
-    IGNORE_CONNEXTCPP=`echo ${CI_ARGS} | sed -e 's/.*ignore-rmw \([^-]*\).*/\1/' | sed -e 's/-.*//' | grep rmw_connext_cpp`
-    # Install RTI Connext DDS if we didn't find both `rmw_connextdds` and `rmw_connext_cpp`
-    # within the "ignored RMWs" option string.
-    if [ -z "${IGNORE_CONNEXTDDS}" -o -z "${IGNORE_CONNEXTCPP}" ]; then
+    IGNORE_CONNEXTDDS=""
+    ignore_rwm_seen="false"
+    for arg in ${CI_ARGS} ; do
+        case $arg in
+            ("--ignore-rmw") ignore_rmw_seen="true" ;;
+            ("-"*) ignore_rmw_seen="false" ;;
```
nuclearsandwich, line 30: The refactor here is worth some attention and impacts my request to split the `--ignore-rmw` changes to other rmw into a separate PR. I still think I prefer it, even though it means including this refactor before dropping rmw_connext.

I haven't seen where in the above logic the closing `-` is added. Also, why a single `-` rather than the typical `--` to close an argument list? At first I thought perhaps an option parser was stripping the initial `-` but since `--ignore-rmw` is intact in the case above that doesn't seem the case.

ros2_batch_job/__main__.py
```diff
@@ -195,7 +195,7 @@ def get_args(sysargv=None):
     parser.add_argument(
         '--os', default=None, choices=['linux', 'osx', 'windows'])
     parser.add_argument(
-        '--ignore-rmw', nargs='*', default=[],
+        '--ignore-rmw', nargs='*', default=[], action='extend',
```
nuclearsandwich, line 198: This change is part of the `--ignore-rmw` refactor as well.

nuclearsandwich (changes requested):
The changes here include a refactor of how the `--ignore-rmw` CI args are handled and I'd prefer it if that change were made in an independent pull request both to preserve the refactor independent from dropping connext and in order to test those changes independently.

linux_docker_resources/entry_point.sh
clalancette: > I haven't seen where in the above logic the closing `-` is added. Also, why a single `-` rather than the typical `--` to close an argument list? At first I thought perhaps an option parser was stripping the initial `-` but since `--ignore-rmw` is intact in the case above that doesn't seem the case.

So to be clear, I'm not actually looking for a `--` to quit an option list here.  What I'm looking for is *any* other option that would stop us from processing this list (and options start with a single or double `-`).

That is, I'm trying to handle all of the following:

* `--option foo --ignore-rmw rmw_foo --ignore-rmw rmw_bar`
* `--option foo --ignore rmw rmw_foo rmw_bar`
* `--option foo --ignore-rmw rmw_foo --ignore-rmw rmw_bar --option1`
* `--option foo --ignore-rmw rmw_foo rmw_bar -- blah` (in this case, the `--` is required as otherwise it would be considered as part of the `--ignore-rmw` list)

All of that said, I'm going to actually recreate this comment in the new PR I'm about to open so we can continue to discuss it there.

clalancette:
All right, with #715 now merged and deployed, I've gone ahead and rebased this.  This is now ready for another review.

nuclearsandwich (approved):
One question arose during review but I don't think the answer is blocking.

Farewell to a real one `rwm_connext_cpp`.

linux_docker_resources/entry_point.sh
```diff
@@ -23,18 +23,17 @@ echo "done."
...
 if [ "${ARCH}" != "aarch64" ]; then
     IGNORE_CONNEXTDDS=""
-    IGNORE_CONNEXTCPP=""
     ignore_rwm_seen="false"
     for arg in ${CI_ARGS} ; do
         case $arg in
             ("--ignore-rmw") ignore_rmw_seen="true" ;;
             ("-"*) ignore_rmw_seen="false" ;;
-            (*) if [ $ignore_rmw_seen = "true" ] ; then [ $arg = "rmw_connextdds" ] && IGNORE_CONNEXTDDS="true" ; [ $arg = "rmw_connext_cpp" ] && IGNORE_CONNEXTCPP="true" ; fi
+            (*) if [ $ignore_rmw_seen = "true" ] ; then [ $arg = "rmw_connextdds" ] && IGNORE_CONNEXTDDS="true" && break ; fi
```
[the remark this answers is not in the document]
clalancette: > Why add a break here where there wasn't one before? Is it just because once you've confirmed this condition there is no longer a need to parse arguments?

Yes, that's exactly it.  We are looking for one and only one thing, so we can get out early.

clalancette:
All right, I've gone ahead and deployed this change to the `test_` jobs only.  Here are some CI jobs with this in place:

* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/144/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/57/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/317/)
* Linux with ConnextDDS disabled [Build Status](http://ci.ros2.org/job/test_ci_linux/145/)

clalancette:
All right, this is looking good.  The normal jobs look OK, and the one where I specifically disabled connext did what I expected and didn't install it.  I'm going to go ahead and merge and deploy this one, then move onto the next one.

PR #715, opened 2023-06-22
Revamp --ignore-rmw handling.

clalancette:
The job templates are currently always setting
'--ignore-rmw rmw_connext_dynamic_cpp', and then adding to that list if other RMWs should be ignored.
But it is not necessarily the case that we will always have an RMW to ignore.  Therefore, we switch to passing --ignore-rmw for *each* of the ones that we want to ignore (because there may not be any). To handle that, change the argparse handling of --ignore-rmw to use the 'extend' action, and then modify the entrypoint.sh to deal with the fact that there might be multiple --ignore-rmw flags, any of which could contain 'connext'.

### DIFF
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -23,17 +23,18 @@ echo "done."
 # We only attempt to install Connext on amd64
 if [ "${ARCH}" != "aarch64" ]; then
     IGNORE_CONNEXTDDS=""
+    IGNORE_CONNEXTCPP=""
     ignore_rwm_seen="false"
     for arg in ${CI_ARGS} ; do
         case $arg in
             ("--ignore-rmw") ignore_rmw_seen="true" ;;
             ("-"*) ignore_rmw_seen="false" ;;
-            (*) if [ $ignore_rmw_seen = "true" -a $arg = "rmw_connextdds" ] ; then IGNORE_CONNEXTDDS="true" ; break ; fi
+            (*) if [ $ignore_rmw_seen = "true" ] ; then [ $arg = "rmw_connextdds" ] && IGNORE_CONNEXTDDS="true" ; [ $arg = "rmw_connext_cpp" ] && IGNORE_CONNEXTCPP="true" ; fi
         esac
     done
 
-    # Install RTI Connext DDS if we didn't find `rmw_connextdds` within the "ignored RMWs" option string.
-    if [ -z "${IGNORE_CONNEXTDDS}" ]; then
+    # Install RTI Connext DDS if we didn't find 'rmw_connextdds' and 'rmw_connext_cpp' within the "ignore-rmw" option strings.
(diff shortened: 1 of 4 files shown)


linux_docker_resources/entry_point.sh
```diff
@@ -22,13 +22,18 @@ echo "done."
...
-    # within the "ignored RMWs" option string.
-    if [ -z "${IGNORE_CONNEXTDDS}" -o -z "${IGNORE_CONNEXTCPP}" ]; then
+    IGNORE_CONNEXTDDS=""
+    ignore_rwm_seen="false"
+    for arg in ${CI_ARGS} ; do
+        case $arg in
+            ("--ignore-rmw") ignore_rmw_seen="true" ;;
+            ("-"*) ignore_rmw_seen="false" ;;
+            (*) if [ $ignore_rmw_seen = "true" -a $arg = "rmw_connextdds" ] ; then IGNORE_CONNEXTDDS="true" ; break ; fi
+        esac
```
clalancette, line 32: Recreating the conversation from the other PR:

> I haven't seen where in the above logic the closing - is added. Also, why a single - rather than the typical -- to close an argument list? At first I thought perhaps an option parser was stripping the initial - but since --ignore-rmw is intact in the case above that doesn't seem the case.

So to be clear, I'm not actually looking for a `--` to quit an option list here. What I'm looking for is any other option that would stop us from processing this list (and options start with a single or double `-`).

That is, I'm trying to handle all of the following:

* `--option foo --ignore-rmw rmw_foo --ignore-rmw rmw_bar`
* `--option foo --ignore rmw rmw_foo rmw_bar`
* `--option foo --ignore-rmw rmw_foo --ignore-rmw rmw_bar --option1`
* `--option foo --ignore-rmw rmw_foo rmw_bar -- blah` (in this case, the `--` is required to terminate the list as otherwise it would be considered as part of the `--ignore-rmw` list)

nuclearsandwich: Thanks for elaborating. In review, the change in argument handling should cover the outlined cases.

nuclearsandwich (approved):
Thanks for splitting this out. It was a much cleaner review!

clalancette:
Here's some CI runs with this in place (I deployed this just to the `test_*` jobs).

Regular CI, to test that normal things work:
* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/140/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/51/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/313/)

CI, testing disabling everything but Fast-DDS:
* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/141/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/52/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/314/)

clalancette:
All right, CI looks good.  Going ahead and merging this one, then deploying.

PR #716, opened 2023-06-26
Remove support for Connext 5.3.1.

clalancette:
There are no current distributions that currently require it.

One thing in this commit that warrants some explanation is the complete removal of the sourcing of Connext in the linux_batch/__init__.py and windows_batch/__init__.py files. They are completely removed because with Connext 6, they are no longer used.  On Linux it is sourced from the entry_point.sh, and on Windows I'm not entirely sure where it is sourced from.

I'll also point out that there is one remaining reference to Connext 5.3.1 in the `windows_docker_resources/install_ros2_foxy.json` file.  But I think it makes more sense to remove that when we completely remove foxy support in an upcoming change.

### DIFF
--- a/windows_docker_resources/Dockerfile
+++ b/windows_docker_resources/Dockerfile
@@ -39,8 +39,6 @@ RUN choco install -y git chef-workstation
 RUN IF NOT EXIST "C:\TEMP" mkdir C:\TEMP
 COPY rticonnextdds-license\ C:\TEMP\rticonnextdds-license
 COPY rticonnextdds-src\ C:\TEMP\rticonnextdds-src
-RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe
-RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg
 RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-host-x64Win64.exe.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-host-x64Win64.exe
 RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64VS2017.rtipkg.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64VS2017.rtipkg
 
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -104,7 +104,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y lcov
 RUN pip3 install -U lcov_cobertura_fix
 
 # Install the Connext binary from the OSRF repositories.
-RUN if test \( ${PLATFORM} = x86 -a ${INSTALL_CONNEXT_DEBS} = true -a ${UBUNTU_DISTRO} != jammy \); then apt-get update && RTI_NC_LICENSE_ACCEPTED=yes apt-get install -y rti-connext-dds-5.3.1; fi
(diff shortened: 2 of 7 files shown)


nuclearsandwich:
> on Windows I'm not entirely sure where it is sourced from.

In the interest of [Chesterton's Fence](https://fs.blog/chestertons-fence/) ideally we'd document why those scripts are no longer needed.

windows_docker_resources/Dockerfile
```diff
@@ -39,8 +39,6 @@ RUN choco install -y git chef-workstation
 RUN IF NOT EXIST "C:\TEMP" mkdir C:\TEMP
 COPY rticonnextdds-license\ C:\TEMP\rticonnextdds-license
 COPY rticonnextdds-src\ C:\TEMP\rticonnextdds-src
-RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-host-x64Win64.exe
-RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg.??? C:\TEMP\rticonnextdds-src\rti_connext_dds-5.3.1-pro-target-x64Win64VS2017.rtipkg
```
nuclearsandwich, line 43: We'll have to decide if it makes sense to truncate these files from the branch they're fetched from in order to make CI setup times a bit smaller. But that's a fine thing for a followup PR.

nuclearsandwich (approved):

nuclearsandwich:
> > on Windows I'm not entirely sure where it is sourced from.
> 
> In the interest of [Chesterton's Fence](https://fs.blog/chestertons-fence/) ideally we'd document why those scripts are no longer needed.

We dug into this offline a little and the abbreviated explanation is that on Windows our chef setup configures the [CONNEXTDDS_DIR environment variable system-wide](https://github.com/ros-infrastructure/ros2-cookbooks/blob/6e23e56ddb058b29b41bc65b602a27f8b50d10de/cookbooks/ros2_windows/recipes/rti_connext.rb#L157-L161) so that 
[rti_load_connextddsdir](https://github.com/ros2/rmw_connextdds/blob/rolling/rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake#L367-L398) in `rti_connext_dds_cmake_module` can find and use it.

clalancette:
> On Linux it is sourced from the entry_point.sh, and on Windows I'm not entirely sure where it is sourced from.

OK, after some good help from @nuclearsandwich , we figured out how Connext is sourced.

On Linux, we end up sourcing `/opt/rti.com/rti_connext_dds-6.0.1/scripts/rtisetenv_x64Linux4gcc7.3.0.sh` during the [entrypoint.sh script](https://github.com/ros2/ci/blob/fcb040764562e04fbdc7361fc40c1300f5806197/linux_docker_resources/entry_point.sh#L43).  In turn, this script ends up setting two environment variables: `NDDSHOME` (the top-level of where Connext is installed), and `LD_LIBRARY_PATH` (so the linkers can find the shared libraries).

On Windows, we end up manually creating an environment variable called `CONNEXTDDS_DIR` during the [chef scripts](https://github.com/ros-infrastructure/ros2-cookbooks/blob/6e23e56ddb058b29b41bc65b602a27f8b50d10de/cookbooks/ros2_windows/recipes/rti_connext.rb#L157-L161).

The last piece of the puzzle, then, is how these environment variables get turned into the libraries that are needed for linking/running.  That happens when the [rti_load_connextddsdir](https://github.com/ros2/rmw_connextdds/blob/f167c5bbccc330ae10da778fe50190dd3ad43cfb/rti_connext_dds_cmake_module/cmake/rti_build_helper.cmake#L367-L398) CMake function is called when trying to build `rmw_connextdds`.  It is important to realize there that it first looks for the `CONNEXTDDS_DIR` environment variable (which is what we set on Windows), and if that doesn't exist, falls back to looking for `NDDSHOME` (which is what we set on Linux).  So both work.

clalancette:
With this approved, what I'm going to do here is to run some test jobs.  I shouldn't need a test deploy since there are no changes to the job templates.  Assuming those come back green I'll go ahead and merge.  Thanks!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19055/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13577/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19805/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/436/)

clalancette:
All right, looks like all CI was green, and I manually verified that they all found Connext as expected.  I'll go ahead and merge this one.

PR #717, opened 2023-07-02
Completely remove Foxy support.

clalancette:
I'll note that after this change, there is still a single reference to Foxy in the code here.  But that reference more has to do with Focal support, so I've left it in place and it will be removed when we remove Focal support in an upcoming patch.


cottsay (approved):
I didn't do any grepping or run the script, but the changes look right-on.

clalancette:
> I didn't do any grepping or run the script, but the changes look right-on.

Thanks!  Like other PRs, I'm going to do a deployment of this to the `test_*` jobs first, and run some CI with those.  Assuming those come back like I expect, I'll then merge and do a full deploy.

clalancette:
Test CI:
* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/58/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/318/)
* RHEL [Build Status](http://ci.ros2.org/job/test_ci_linux-rhel/24/)

clalancette:
All right, looks like the infrastructure worked out.  Going ahead and merging this, then doing the full deploy.

PR #718, opened 2023-07-05
Remove focal support.

clalancette:


linux_docker_resources/Dockerfile
```diff
@@ -33,7 +33,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   apt-transport-https
 
 # Add the ROS repositories to the apt sources list.
-RUN if test ${UBUNTU_DISTRO} != jammy; then echo "deb http://repositories.ros.org/ubuntu/testing/ `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list; fi
 RUN if test ${UBUNTU_DISTRO} = jammy; then echo "deb http://repo.ros2.org/ubuntu/testing/ `lsb_release -cs` main" > /etc/apt/sources.list.d/ros2-latest.list; fi
```
nuclearsandwich, line 36:
```suggestion
RUN echo "deb http://repo.ros2.org/ubuntu/testing/ `lsb_release -cs` main" > /etc/apt/sources.list.d/ros2-latest.list
```

Since any future distros would also be using the ROS 2 repositories I think we can drop the conditional here, which was introduced in #598 specifically because there was no ROS 1 repository for Jammy.

linux_docker_resources/Dockerfile
```diff
@@ -44,8 +43,6 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | apt-key add -
 # Install some development tools.
 RUN apt-get update && apt-get install --no-install-recommends -y build-essential ccache cmake pkg-config python3-empy python3-pip python3-setuptools python3-vcstool python3-venv
 RUN apt-get update && apt-get install --no-install-recommends -y python3-lark python3-opencv
-# Install virtualenv 16.7.9 needed for Foxy builds on Focal. https://github.com/ros2/ci/issues/400
-RUN if test ${UBUNTU_DISTRO} = focal; then python3 -m pip install virtualenv==16.7.9; fi
```
nuclearsandwich, line 48: Glad to see the back of this kludge!

nuclearsandwich (approved):
I have a non-blocking suggestion but otherwise good to go!

linux_docker_resources/Dockerfile
clalancette: Yeah, that's fair.  I kind of waffled on what to do here, and ended up leaving it, but I'll go back and remove it.

linux_docker_resources/Dockerfile
clalancette: We still have it in place in https://github.com/ros2/ci/blob/2c8cddea04fd79c1e5d8805a133e695d6f724c52/ros2_batch_job/__main__.py#L472 .  Do you think I should attempt to remove it from there as well?

linux_docker_resources/Dockerfile
clalancette: And fixed in e4bb2b7

linux_docker_resources/Dockerfile
clalancette: I'm going to go forward without this for now.  We can always revisit it in a separate PR.

clalancette:
As with other PRs in this series, I'm going to go ahead and deploy just to the `test_` jobs and then run CI.  Assuming that looks good, I'll then merge this and deploy fully.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/test_ci_linux/147/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/test_ci_linux-aarch64/59/)
* Windows [Build Status](http://ci.ros2.org/job/test_ci_windows/319/)

clalancette:
CI looks good, so merging and deploying.

PR #726, opened 2023-11-14
Pin to RHEL 9.1 waiting pyside2 rebuild on EPEL

claraberendsen:
## Description
Since a couple of days ago nightlies on RHEL have been failing with: 
```
00:01:05.746 Error: 
00:01:05.746  Problem: package python3-pyside2-devel-1:5.15.7-2.el9.x86_64 requires libpyside2.cpython-39-x86_64-linux-gnu.so.5.15()(64bit), but none of the providers can be installed
00:01:05.746   - conflicting requests
00:01:05.746   - nothing provides libQt5Charts.so.5(Qt_5.15.3_PRIVATE_API)(64bit) needed by python3-pyside2-1:5.15.7-2.el9.x86_64
```
It seems that there has been a RHEL 9 release of `qt 5.15.9` and that the `pyside2` package has since been broken due to a dependency on the `5.15.3` private API. 
The release likely solves the  [security issue on version `5.15.3` of qt](https://errata.almalinux.org/9/ALSA-2022-8022.html). 

To resolve this issue this PR ping the version to 9.1 of the docker image that has not been updated.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9',
+        'el_release': '9.1',
         'ros_distro': 'rolling',
     }
 
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9.1',
+        'el_release': '9',
         'ros_distro': 'rolling',
     }
 
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -14,7 +14,7 @@ RUN dnf install epel-release 'dnf-command(config-manager)' --refresh -y && \
 RUN dnf -v repolist --enabled
 
 # Picking qtcharts from EPEL 8 url instead of EPEL 9 due to error https://github.com/ros2/ci/issues/727
-RUN dnf install https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/q/qt5-qtcharts-5.15.3-1.el8.x86_64.rpm
+RUN dnf install -y https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/q/qt5-qtcharts-5.15.3-1.el8.x86_64.rpm
 
 # Install foundation packages
 RUN dnf install \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -20,7 +20,10 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Pac
 
 # add missing clang deps
 RUN dnf install -y https://aws.repo.almalinux.org/8.8/AppStream/x86_64/os/Packages/llvm-libs-15.0.7-1.module_el8.8.0%2B3466%2Bdfcbc058.x86_64.rpm \
-                   https://aws.repo.almalinux.org/8.8/AppStream/x86_64/os/Packages/mesa-dri-drivers-22.3.0-2.el8.x86_64.rpm 
+                   https://aws.repo.almalinux.org/8.8/AppStream/x86_64/os/Packages/mesa-dri-drivers-22.3.0-2.el8.x86_64.rpm \
+                   https://aws.repo.almalinux.org/8.8/AppStream/x86_64/os/Packages/mesa-filesystem-22.3.0-2.el8.x86_64.rpm 
+
+
 
 
 
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -14,7 +14,7 @@ RUN dnf install epel-release 'dnf-command(config-manager)' --refresh -y && \
 RUN dnf -v repolist --enabled
 
 # Picking qtcharts and qtwebengine from EPEL 8 url instead of EPEL 9 due to error https://github.com/ros2/ci/issues/727
-RUN dnf install -y https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/q/qt5-qtcharts-5.15.3-1.el8.x86_64.rpm \
+RUN dnf install -y https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/q/qt5-qtcharts-5.15.3-1.el8.x86_64.rpm 
                   #  https://aws.repo.almalinux.org/8.8/BaseOS/x86_64/os/Packages/libevent-2.1.8-5.el8.x86_64.rpm \
                   #  https://d2lzkl7pfhq30w.cloudfront.net/pub/epel/8/Everything/x86_64/Packages/q/qt5-qtwebengine-5.15.8-5.el8.1.x86_64.rpm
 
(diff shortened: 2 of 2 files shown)


create_jenkins_job.py
```diff
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9',
+        'el_release': '9.1',
```
clalancette, line 124: I'll point out that this is not a long-term solution; people won't be able to install ROS 2 on RHEL 9.2 or later.  However, I think for now it is better to be running CI on RHEL (even on an old version), than not to be running at all.  So I'm OK with the change.

Because this is a change to `create_jenkins_job.py`, this won't have an effect until it is deployed.  I'll suggest deploying this to only the `test_` jobs (something like `create_jenkins_job.py --select-jobs-regexp 'test_*'`), and then running a new CI job using https://ci.ros2.org/view/All/job/test_ci_linux-rhel/.  Once we've determined that this works fine for the test job, we can then merge it and deploy it to everything.

claraberendsen: Agree that this is not a long term solution. 
I will do the deployment on the test jobs and we can check if this solves it.
Long term in these scenarios we would either need to not depend on EPEL and upstream rebuild to keep supporting newer versions of RHEL, otherwise we are bound to the cadence of rebuilts on EPEL.

claraberendsen:
@clalancette from my understanding this change even though is to `create_jenkins_job.py` what it does is change the default args  to be `9.1` instead of `9`. Therefore I can test it by running CI changing those values. I believe this was put in place by Scott in #701 
 
**Almalinux 8**: [Build Status](https://ci.ros2.org/job/ci_linux-rhel/479/)
**Almalinux 9.1**:[Build Status](https://ci.ros2.org/job/ci_linux-rhel/480/)

claraberendsen:
The `9.1` build seems to have failed to find `libclang.so.15` it seems that in almalinux 9 is in the `Devel` package repository instead of `AppStream` ( See https://pkgs.org/search/?q=libclang.so.15).

claraberendsen:
> Is there a reason you're targeting 9.1 instead of 9.2? The current release is 9.3.

I understood the current release was 9.2 :laughing: . I can target 9.2 to see if it produces any other issue.

cottsay:
> it seems that in almalinux 9 is in the `Devel` package repository instead of `AppStream` ( See https://pkgs.org/search/?q=libclang.so.15).

I think this might be an AlmaLinux-specific offering that isn't available on RHEL, so I'm not sure we should use it here. That should be confirmed, though.

claraberendsen:
**Almalinux 9.2** [Build Status](https://ci.ros2.org/job/ci_linux-rhel/483/)

claraberendsen:
> I think this might be an AlmaLinux-specific offering that isn't available on RHEL, so I'm not sure we should use it here. That should be confirmed, though.

You might be right what I can see [here](https://wiki.almalinux.org/repos/AlmaLinux.html) is: 

> Content in the Devel repo includes packages that are not normally provided in the base nor extra repositories, but needed for build-time dependencies of other packages. Devel is NOT meant to satisfy runtime dependencies or for long term use on general purpose machines.

Here is a summary of the results so far: 
So basically on 9.2 the error is the same as if pointing to 9. 
On `9.1` we are missing `libclang.so.15` for `shiboken2`.
`8` failed due to a missing package but on tracetools not on docker build phase. 

@cottsay since I have you in this thread, thought on how to move forward?

claraberendsen:
Looking at the [release date of Almalinux  9.3](https://wiki.almalinux.org/release-notes/) of Nov 13th, this smells like changes to changes on the package upstream in preparation for the new release since the nightlies started to fail on Nov 12th.

cottsay:
> thought on how to move forward?

This is tricky. The packaging issues we've seen in the past have been rooted in SOVERSION bumps during RHEL minor releases breaking EPEL packages, which is a pretty straightforward and predictable problem. What I think we're seeing here is a rather innocuous patch release of Qt that was pushed to the stable (9.2) release of RHEL that shouldn't have broken anything, but pyside2 has a tight coupling to the Qt version that seems to necessitate tandem builds.

Looking at the release history of pyside2 in Fedora, there are a lot of release bumps and rebuilds caused by updates to Qt, so I think this is probably a well known problem with Pyside2. Super bummer.

The only repository hand waving I can think of is to try blocking installation of the newer Qt binaries. If the older ones aren't available in the repository anymore (likely) we may need to install them from the 9.1 repositories, but get the rest of the packages from 9.2/9.3 as usual. Since the Qt releases have the same SOVERSION, it should be safe to do unless there are other packages which also depend on Qt private API.

claraberendsen:
Manually linking the `qt5charts` package seems to have resulted in another error 
[Build Status](https://ci.ros2.org/job/ci_linux-rhel/485/) 
```
Error: 
17:32:03  Problem: package python3-pyside2-devel-1:5.15.7-2.el9.x86_64 requires shiboken2, but none of the providers can be installed
17:32:03   - conflicting requests
17:32:03   - nothing provides libclang.so.15()(64bit) needed by shiboken2-1:5.15.7-2.el9.x86_64
17:32:03   - nothing provides libclang.so.15(LLVM_13)(64bit) needed by shiboken2-1:5.15.7-2.el9.x86_64
```

shiboken2 is also provided by EPEL tracking down provider for libclang.so.15

cottsay:
I don't think you'll be able to install any RHEL 8 (or EPEL 8) packages on any version of EPEL 9.

Qt 5.15.3 is available in the RHEL 9.1 repos, but I no longer think this is a viable route forward. The [EPEL update](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-9e7b2b8d0f) that brought newer versions of qtcharts and qtwebengine is also a problem, and that rabbit hole may go deeper. Trying to downgrade all of the qt packages across the RHEL repositories and EPEL may be unmaintainable or may not be possible at all.

---

The issue related to shiboken2 and clang is separate, and I think is related to the RHEL 9.3 release moving to clang 16. As it happens, shiboken2 is a subpackage of python-pyside2, so whatever solution we arrive at may mitigate both problems.

It's a mess. Part of it is EPEL's fault for not tracking RHEL minor releases very well, but this whole pyside2-private-API thing is pretty awful and unpredictable.

claraberendsen:
Yes, sadly this is a hot mess. 
Trying to manually plumb `qtcharts` by itself might have worked if that was the only divergence but it's not. Most of qt dependencies have been updated on EPEL 9 to work with AlmaLinux 9.3, and since EPEL doesn't manage minor version or and archive the compatible packages for 9.2 that were working before are no longer available. 

One solution in the future could be to maintain our own archive of EPEL packages to be able to revert on breaking changes and not depend on the bleeding edge EPEL major version repo. However, this means hosting them and establishing the mechanism for updating them. 

I'm giving it a last try now to see if I see the end of the dependencies rabbit hole, otherwise if we want to restore service we might have to revert all the way back to RHEL 8 while we wait for `pyside2` rebuild. (@clalancette)

clalancette:
How about a different tactic here.  Is there any chance we could take a cached Docker image off of one of the workers and pin to that Docker image?  It isn't ideal, but hopefully it is a temporary measure.

claraberendsen:
> How about a different tactic here. Is there any chance we could take a cached Docker image off of one of the workers and pin to that Docker image? It isn't ideal, but hopefully it is a temporary measure.

I think I have found one of the agents with old docker images. Copying locally to my computer now and will check were can we put them to fetch them from. 
It isn't ideal but I think it's the only workaround to at least have RHEL 9.2 testing and builds. This will effectually ping us to RHEL 9.2 until we can go back to our normal workaround.

claraberendsen:
Closing this in favour of #728 since strategy has changed. That PR however won't solve the underlying issue linked to this PR but should get us back working at least on ci.ros2.org.

Issue #727, opened 2023-11-14
Tight coupling of `pyside2` to `Qt` version can cause build regressions


claraberendsen:
## Description 
RHEL nightlies have started to fail since Nov 12th with an issue with unprovided dependencies for `Qt 5.15.3` private APIs for the `python3-pyside2-devel` package. 

```
00:01:05.746 Error: 
00:01:05.746  Problem: package python3-pyside2-devel-1:5.15.7-2.el9.x86_64 requires libpyside2.cpython-39-x86_64-linux-gnu.so.5.15()(64bit), but none of the providers can be installed
00:01:05.746   - conflicting requests
00:01:05.746   - nothing provides libQt5Charts.so.5(Qt_5.15.3_PRIVATE_API)(64bit) needed by python3-pyside2-1:5.15.7-2.el9.x86_64
```
OS version on ci is not pinned to a specific minor version of an OS but rather pulls the latest docker image for that given major version to build upon. On Nov 13th a version `9.3` of AlmaLinux was released with updates to `Qt` moving it from `5.15.3` to `5.15.9` likely prompted by security issue [ALSA-2022:8022](https://errata.almalinux.org/9/ALSA-2022-8022.html)

EPEL which provides the `python3-pyside2-devel` package does not manage minor and major version of the updates therefore upon breaking changes to almalinux it need to be rebuilt pointing to the newer package.  Previous errors derived from this were resolved by pinning to a previous version of almalinux with the necessary packages.  ( See #701 )

However pinning to 9.2 or 9.1 didn't solve the issue (details in #726) which seemed odd.

Looking up which package provides the missing file (https://pkgs.org/search/?q=libQt5Charts.so.5) it seems that in most OS is the `qt5-qtcharts` package which is not available in AlmaLinux but it was being provided by EPEL as well. 

#### Details 
OS: AlmaLinux 9.3 / 9.2
EPEL: 9

Further comments and debugging can be found on #726

PR #728, opened 2023-11-16
Use stale container image with AlmaLinux 9.2 and working dependencies

claraberendsen:
### Description 
With the new release of AlmaLinux 9.3 EPEL has made breaking changes to the upstream dependencies that due to
-  #727 

has rendered impossible to build on-demand new RHEL images. 

To get ci back  working this PR fetches prebuilt container images built before of the release that contain the correct dependencies. 
This is not a preferred state for running ci and these changes will affect us more deeply on build.ros2.org however until the drift of dependencies gets resolved upstream it's an acceptable compromise  to keep running ci for RHEL. 

#### Affected jobs on ci.ros2.org
- RHEL CI jobs
- RHEL nightlies 
- RHEL packaging jobs

### Testing 
**Strategy**
 - Deploy changes only to test jobs
 - Build with RHEL 9.2-pinned build args
 
**Test new default config return green ci**

 test_packaging_linux-rhel [Build Status](https://ci.ros2.org/view/All/job/test_packaging_linux-rhel/6/)
test_ci_linux-rhel [Build Status](https://ci.ros2.org/view/All/job/test_ci_linux-rhel/38/)

**Test if statement shifts to other el releases when not in pinned version**
> [!NOTE]  
> The failure is expected here it shows that it's not using the default image

 test_packaging_linux-rhel [Build Status](https://ci.ros2.org/view/All/job/test_packaging_linux-rhel/7/)
test_ci_linux-rhel [Build Status](https://ci.ros2.org/view/All/job/test_ci_linux-rhel/32/)

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9',
+        'el_release': '9.2',
         'ros_distro': 'rolling',
     }
 
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9.2',
+        'el_release': '9.2-pinned',
         'ros_distro': 'rolling',
     }
 
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -211,7 +211,7 @@ docker build --pull ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_ba
 # This ought to be temporary until pyside2 can be rebuilt to work with AlmaLinux 9.3. See https://github.com/ros2/ci/issues/727. 
 # An if-else scenario is added to be able to quickly test if it has been resolved when the rebuilt happens.
 
-if [ "$CI_EL_RELEASE" == "9.2-pinned" ]; then
+if [ "$CI_EL_RELEASE" = "9.2-pinned" ]; then
   echo "Pulling stale image for docker with AlmaLinux 9.2"
   docker pull ghcr.io/ros-infrastructure/ros2_batch_ci_rhel:latest
   docker tag ghcr.io/ros-infrastructure/ros2_batch_ci_rhel ros2_batch_ci_rhel
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -207,14 +207,18 @@ echo "# BEGIN SECTION: Build Dockerfile"
 @[    if os_name == 'linux-aarch64']@
 docker build --pull ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_batch_ci_aarch64 linux_docker_resources
 @[    elif os_name == 'linux-rhel']@
-echo "Pulling stale image for docker with AlmaLinux 9.2"
-docker pull ghcr.io/ros-infrastructure/ros2_batch_ci_rhel:latest
-docker tag ghcr.io/ros-infrastructure/ros2_batch_ci_rhel ros2_batch_ci_rhel
-
 # FIX ME !!
 # Due to issues with pyside2, EPEL and AlmaLinux 9.3 it's impossible to build docker images; to get ci working we reverted to a previous built stale image.
 # This ought to be temporary until pyside2 can be rebuilt to work with AlmaLinux 9.3. See https://github.com/ros2/ci/issues/727. 
-# docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+# An if-else scenario is added to be able to quickly test if it has been resolved when the rebuilt happens.
+if [[ $CI_EL_RELEASE == "9.2-pinned" ]]; then
+  echo "Pulling stale image for docker with AlmaLinux 9.2"
+  docker pull ghcr.io/ros-infrastructure/ros2_batch_ci_rhel:latest
+  docker tag ghcr.io/ros-infrastructure/ros2_batch_ci_rhel ros2_batch_ci_rhel
+; else 
+  docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+; fi
+
 @[    elif os_name == 'linux']@
 docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci linux_docker_resources
 @[    else]@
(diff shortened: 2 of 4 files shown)


clalancette (commented):
I'm wondering if we should make a new "el_release" choice, called something like "9.2-pinned".  Then in `ci_job.xml.em`, if the el_release is 9.2-pinned, we fetch the pinned image.  Otherwise, we go ahead with trying to build the docker image from the Dockerfile-RHEL.  And we make 9.2-pinned the default.

If we do this, then on the nightlies and CI jobs, we'll succeed in using the pinned image.  But if we want to try out a "regular" EL 9 image to see if things are fixed, we can do that without a PR.  We just need to choose "9" as the choice.

What do you think?

claraberendsen:
> What do you think?
I like the idea since it would give us more flexibility to test whenever the dependencies can be updated and gives developers the chance of changing according to need. It's definitely less breaking than the current implementation. 
The only thing that concerns me is that whenever things are fixed we should clean-up that option, but that is payload on our side to keep things in order.

job_templates/ci_job.xml.em
```diff
@@ -207,7 +207,19 @@ echo "# BEGIN SECTION: Build Dockerfile"
 @[    if os_name == 'linux-aarch64']@
 docker build --pull ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_batch_ci_aarch64 linux_docker_resources
 @[    elif os_name == 'linux-rhel']@
-docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_batch_ci_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+# FIX ME !!
```
nuclearsandwich, line 210:
```suggestion
```

This isn't a hack that we can "fix" and take out. The scare comment doesn't provide any value here. At best, this could be a TODO to revert once EPEL 9 has stabilized.

create_jenkins_job.py
```diff
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
-        'el_release': '9',
+        'el_release': '9.2-pinned',
```
nuclearsandwich, line 124: I don't think it makes sense to use a custom value here instead of the true release we target. We didn't do this when we pinned a safe Jammy image and unless there's prior art or a strong motivation to deviate here I'd prefer not to.

Testing the upstream RHEL 9 build can be done by reverting this PR entirely and testing that branch.

job_templates/ci_job.xml.em
```diff
@@ -207,7 +207,19 @@ echo "# BEGIN SECTION: Build Dockerfile"
...
+# Due to issues with pyside2, EPEL and AlmaLinux 9.3 it's impossible to build docker images; to get ci working we reverted to a previous built stale image.
+# This ought to be temporary until pyside2 can be rebuilt to work with AlmaLinux 9.3. See https://github.com/ros2/ci/issues/727. 
+# An if-else scenario is added to be able to quickly test if it has been resolved when the rebuilt happens.
```
nuclearsandwich, line 213: I'm not sure the conditional is justifiable as the real test will be reverting this change once ready. I could be talked out of it but in general I prefer not to add conditions to these scripts and keep them as sigh-readable as possible.

job_templates/packaging_job.xml.em
```diff
@@ -204,7 +204,19 @@ echo "# BEGIN SECTION: Build Dockerfile"
...
+# An if-else scenario is added to be able to quickly test if it has been resolved when the rebuilt happens.
+
+if [ "$CI_EL_RELEASE" == "9.2-pinned" ]; then
+  echo "Pulling stale image for docker with AlmaLinux 9.2"
+  docker pull ghcr.io/ros-infrastructure/ros2_packaging_rhel:latest
+  docker tag ghcr.io/ros-infrastructure/ros2_packaging_rhel ros2_packaging_rhel
+else 
+  docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_packaging_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+fi
+
```
nuclearsandwich, line 219: Once iterated on in the ci_job template, some of the same comments will apply here as well but let's wait until those reviews settle to update here in order to reduce churn.

nuclearsandwich (changes requested):

job_templates/ci_job.xml.em
claraberendsen: @nuclearsandwich I see this more in terms of the effort necessary to test that the issue with AlmaLinux 9.3 and EPEL has been resolved. If I don't leave the mechanism to test easily in the form of the conditional  while reverting this PR, I would need to redeploy the new changes to the test jobs and do the testing vs simply running ci with a different parameter.  This strategy was what allowed me to test the build regressions across different versions of AlmaLinux without needed to redeploy to test each in the closed PR linked. 
Furthermore since both changes are contained in this PR  the new addition should be removed in the same process of reverting the stale images changes and won't need extra PRs of cleanup. 
I prefer to test that the change has worked before reverting to a different state on prod.

create_jenkins_job.py
clalancette: At the end of the day, I don't care too much.  I think it would be slightly nicer to be able to test whether RHEL-9/EPEL has rebuilt everything without a PR, but if we think that is too complicated I'm fine with just unconditionally pinning it.

claraberendsen: Left my thoughts on the comment below. If it's a blocker to merge it for you @nuclearsandwich I'm happy to change it to how it was before though I do see some value in adding an option here.

nuclearsandwich: I'm outnumbered and don't strongly object. So I'll not block on preference.

job_templates/ci_job.xml.em
nuclearsandwich: When there isn't an extraordinary concern, implementor's preference has precedence.

job_templates/packaging_job.xml.em
```diff
@@ -204,7 +204,19 @@ echo "# BEGIN SECTION: Build Dockerfile"
 @[    if os_name == 'linux-aarch64']@
 docker build --pull ${DOCKER_BUILD_ARGS} --build-arg PLATFORM=aarch64 -t ros2_packaging_aarch64 linux_docker_resources
 @[    elif os_name == 'linux-rhel']@
-docker build --pull ${DOCKER_BUILD_ARGS} -t ros2_packaging_rhel linux_docker_resources -f linux_docker_resources/Dockerfile-RHEL
+# FIX ME !!
```
nuclearsandwich, line 207:
```suggestion
```

Same comment as above, FIXME isn't an appropriate symbol for this comment.

nuclearsandwich (approved):
There's one outstanding comment which was already addressed on the ci_linux template but this is ready otherwise.

claraberendsen:
> There's one outstanding comment which was already addressed on the ci_linux template but this is ready otherwise.

Addressed in https://github.com/ros2/ci/pull/728/commits/dfab04bf911fa9e74b2b8b82a57bd799cc18c6ab

claraberendsen:
Now that CI has returned :green_circle:  I will deploy on the RHEL jobs on ci.ros2.org and merge.

claraberendsen:
Deployed with regexp  `'.*-rhel'` to modify only rhel jobs.

PR #732, opened 2023-12-07
Revert "Use stale container image with AlmaLinux 9.2 and working depe…

clalancette:
…ndencies (#728)"

This reverts commit 759a3bcb49f674a535e6b6089d61d2f24f5a09d3.

With the recent work in https://github.com/ros2/ci/pull/730 and https://github.com/ros-visualization/python_qt_binding/pull/129 , we should theoretically no longer have to pin our RHEL images.  This reverts the change that added the pin, and I'll run CI on it next to see what happens.

FYI @claraberendsen @cottsay

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'jammy',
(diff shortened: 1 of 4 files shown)


clalancette:
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/540/)

cottsay (approved):

clalancette:
Merged and now deployed.

PR #738, opened 2024-01-15
Add noble as an Ubuntu distro of choice.

marcoag:
Starting this PR to gather feedback.

I believe with should hold the merge until the bootstrap repository step is done.

Note that:

- No ROS1 has been set.
- The dockerfile will have noble for ROS Rolling.

PR #739, opened 2024-01-16
Add Noble as choice for Ubuntu distro 

claraberendsen:
### Description
This PR adds Ubuntu Noble 24.04 as option for ci.ros2.org.

### Testing
- Deploy to test_ci_linux
- Do a test build.


clalancette:
We already have #738 that does this same thing.  I'd rather continue on in that PR, if that is OK with you @claraberendsen

claraberendsen:
> We already have #738 that does this same thing. I'd rather continue on in that PR, if that is OK with you @claraberendsen

Did not see it! Yes let's continue there.

PR #744, opened 2024-01-28
Pin pytest to an older version so tests succeed.

clalancette:
launch_pytest fails with pytest 8.0.0.  We need to investigate that, but this is a temporary, emergency fix so we don't lose all of the nightlies again.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14599/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20875/)

clalancette:
@nuclearsandwich @claraberendsen @Crola1702 I'm going to emergency merge this so that we get the nightlies back overnight.  But we should discuss this in more detail tomorrow.

PR #747, opened 2024-02-02
Remove the installation of typed_ast from Windows Debug.

clalancette:
This was only needed for older versions of mypy, but with the upgrade to mypy 0.942, we don't need it anymore.

While we are here, also update to a newer version of mypy on Windows, which matches the version in Ubuntu 22.04.

Requires https://github.com/ros-infrastructure/ros2-cookbooks/pull/61 to be merged first.


clalancette:
CI (windows only, since this only affects Windows):

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20921/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20920/)

clalancette:
I'll note that Windows Debug is failing, but it has been failing for 2 months and it is unrelated to this change.

mjcarroll (approved):

PR #750, opened 2024-02-12
Install as many packages in the Dockerfile as possible.

clalancette:
This is in preparation for removing the installation of almost all packages from the batch job script.  Also, this just seems more correct and is much more like what we do when we deliver pre-built packages (.deb and .rpm) to users.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20273/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/579/)


linux_docker_resources/Dockerfile
```diff
@@ -65,9 +65,22 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   pybind11-dev \
   pydocstyle \
   python3-argcomplete \
+  python3-colcon-argcomplete \
+  python3-colcon-common-extensions \
```
cottsay, line 69: Nowhere in ci.ros2.org do we use `colcon-common-extensions`. I'd feel much better if we continued to call out exactly the extensions we want to use in this context.

clalancette: All right, b7cc5de switches to call them out individually.

clalancette:
Rolling CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20341/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/588/)

Humble CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20342/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/589/)

clalancette:
One more round of CI to make sure everything is still happy:

Rolling CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20346/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/591/)

Humble CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20347/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/592/)

PR #754, opened 2024-02-15
Remove unused --python-interpreter argument.

clalancette:
We haven't honored this since we removed OSX support, so just get rid of the code.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -299,9 +299,6 @@ if "!CI_COLCON_MIXIN_URL!" NEQ "" (
 if "!CI_CMAKE_BUILD_TYPE!" NEQ "None" (
   set "CI_ARGS=!CI_ARGS! --cmake-build-type !CI_CMAKE_BUILD_TYPE!"
 )
-if "!CI_CMAKE_BUILD_TYPE!" == "Debug" (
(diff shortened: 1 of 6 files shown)


clalancette:
As mentioned about, currently we don't use this at all.  As of this writing, the way we are finding the Windows Debug interpreter is via some heuristics in [python_cmake_module](https://github.com/ros2/python_cmake_module/blob/a3db97e7ef31d94bd248ddd43fcae8f26b1a3e46/cmake/Modules/FindPythonExtra.cmake#L65-L88).

In https://github.com/ros2/ros2/pull/1524, we are proposing to remove python_cmake_module, which will of course remove those heuristics.  Instead, we'll need to do something different to find the Windows Debug interpreter, which is prototyped in https://github.com/ros2/ci/pull/755 .

cottsay (approved):

clalancette:
After poking at this for a bit, I was mistaken.  We do use this, just not in the way I thought.

In particular, we use this during initial setup to pick the correct Python interpreter to use when installing dependencies.  That is, this is *not* related to the python interpreter used during the colcon/cmake build, but is instead the interpreter used when doing setup.

So with that in mind, we actually need to keep this.  So I'll close out this PR.

PR #755, opened 2024-02-15
Use the correct interpreter for Windows Debug.

clalancette:
This is connected with https://github.com/ros2/ros2/pull/1524 ; before we merge in most of the PRs connected with that, we'll need to merge in something like this fix.  Note that this is still a draft because I need to test deploy this.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -248,6 +248,7 @@ def main(argv=None):
         }
         if os_name == 'windows':
             job_name = job_name[:15]
+            debug_config['build_args_default'] = re.sub(r'(--cmake-args)', r'\1 -DPython3_EXECUTABLE=C:\\Python38\\python_d.exe', data['build_args_default'])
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

cottsay (approved):
A couple of non-critical thoughts:
1. I'm not blown away by the fact that we're coupling this code to a provisioning detail like the installation location of the python debug interpreter executable, but I don't see an obvious way to avoid it.
2. Because it's just a simple substitution, you man consider using `str.replace()` instead of a regex.

clalancette:
CI for this (along with all of the rest of the changes needed) is in https://github.com/ros2/ros2/pull/1524#issuecomment-2365359588

clalancette:
Thanks for the review.  For now, I'm going to leave it as a regex just to get this in.  I have some larger plans for revamping this whole file, so we can probably address that at that point.

PR #759, opened 2024-03-20
Pin all versions of python dependencies.

clalancette:
This allows us to be in control of our destiny,
and not have random python upstream package updates destroy the entire buildfarm.

We accomplish this through 3 different mechanisms, depending on the platform:

1.  Ubuntu - we install everything from debian packages, and nothing from pip.  Because Ubuntu promises to not to breaking changes within a distribution, this guarantees we won't break.  This also serves as our canonical version of what should work.
2.  RHEL - For RHEL-9, we install everything from RPM packages, and nothing from pip.  For RHEL-8, we install most everything from RPM packages, with the exception of mypy, pytest, and pytest-rerunfailures.  Since RHEL/EPEL typically doesn't update major versions within a release, we typically won't break.
3.  Windows - We install all python packages via pip, and we constrain the versions we install to those in Ubuntu.  This means that we'll always match what works on Ubuntu and we should always work.

Still a draft, as I need to run CI here and see what happens.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,4 +20,4 @@
 [submodule "windows_docker_resources/ros2-cookbooks"]
 	path = windows_docker_resources/ros2-cookbooks
 	url = git@github.com:ros-infrastructure/ros2-cookbooks
-	branch = latest
+	branch = clalancette/windows-pip-install
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -47,66 +47,6 @@ from .util import UnbufferedIO
 sys.stdout = UnbufferedIO(sys.stdout)
 sys.stderr = UnbufferedIO(sys.stderr)
 
-# One of the maintainers of pyparsing suggests pinning to 2.4.7 for now;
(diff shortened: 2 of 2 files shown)


clalancette:
Initial CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20513/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21260/)

clalancette:
Note that this needs https://github.com/ros-infrastructure/ros2-cookbooks/pull/66 as well.

clalancette:
Here's additional CI:
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/608/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21280/)

claraberendsen:
Re-running CI on windows with the latest changes as sanity check: 
[Build Status](https://ci.ros2.org/job/ci_windows/21293/)

clalancette:
We switched all of this to use pixi, so closing this.

PR #761, opened 2024-03-25
Install flake8 from proposed

clalancette:
This is still a draft because things aren't totally working here; will need another update.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -190,7 +190,13 @@ RUN apt-get update && apt-get install --no-install-recommends -y libopencv-dev
 RUN apt-get update && apt-get install --no-install-recommends -y libconsole-bridge-dev
 
 # Install build dependencies for rviz et al.
-RUN apt-get update && apt-get install --no-install-recommends -y libassimp-dev libcurl4-openssl-dev libfreetype6-dev libgles2-mesa-dev libglu1-mesa-dev libqt5core5a libqt5gui5 libqt5opengl5 libqt5widgets5 libxaw7-dev libxrandr-dev qtbase5-dev
+RUN apt-get update && apt-get install --no-install-recommends -y libassimp-dev libcurl4-openssl-dev libfreetype6-dev libgles2-mesa-dev libglu1-mesa-dev libqt5opengl5 libxaw7-dev libxrandr-dev qtbase5-dev
+
+RUN if test \( ${UBUNTU_DISTRO} = jammy \); then \
+  apt-get update && apt-get install --no-install-recommends -y libqt5core5a libqt5gui5 libqt5widgets5 \
+;else \
+  apt-get update && apt-get install --no-install-recommends -y libqt5core5t64 libqt5gui5t64 libqt5widgets5t64 \
+;fi
 
 # Install build dependencies for rqt et al.
 RUN apt-get update && apt-get install --no-install-recommends -y pyqt5-dev python3-pyqt5 python3-pyqt5.qtsvg python3-sip-dev python3-pydot python3-pygraphviz
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -125,11 +125,11 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 # Configure proposed source, and install python3-flake8 from it
 RUN if test \( ${UBUNTU_DISTRO} = noble \); then \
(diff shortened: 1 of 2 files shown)


linux_docker_resources/Dockerfile
```diff
@@ -116,6 +121,19 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   uncrustify \
   yamllint
 
+
+# Configure proposed source
+RUN if test \( ${UBUNTU_DISTRO} = noble \); then\
+    echo "deb http://archive.ubuntu.com/ubuntu/ "$(lsb_release -cs)"-proposed restricted main multiverse universe" >> /etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list \
+;fi
+# Install from proposed conflicting packages
```
nuclearsandwich, line 129: What does `conflicting` mean in this context?

ros2_batch_job/__main__.py
```diff
@@ -505,19 +500,23 @@ def need_package_from_pipy(pkg_name):
...
 
         pip_cmd = ['"%s"' % job.python, '-m', 'pip', 'install', '-c', 'constraints.txt', '-U']
         if sys.platform == 'win32':
             # Force reinstall so all dependencies are in virtual environment
             # On Windows since we switch between the debug and non-debug
             # interpreter all packages need to be reinstalled too
             pip_cmd.append('--force-reinstall')
+        else:
+            # We need to use --user so that certain packages (like flake8-blind-except)
+            # can successfully build on RHEL-8 (Humble).
```
nuclearsandwich, line 516: In that case does it make sense to restrict the addition of this flag to the platform that requires it?

linux_docker_resources/rti_web_binaries_install_script.py
```diff
@@ -26,8 +26,8 @@ def install_connext(installer_path, install_directory):
         while(keep_agreeing):
             result_index = child.expect([
                 'Installation Directory.*?:',
-                'Press \[Enter\] to continue:',
-                'Do you accept this license\? \[y/n\]: '])
+                r'Press \[Enter\] to continue:',
+                r'Do you accept this license\? \[y/n\]: '])
```
nuclearsandwich, line 30: It's unclear from any commit message why this change was made alongside the break system packages environment variable usage.

nuclearsandwich (commented):
There appear to be at least three distinct principal changes stacked into this PR and a lot of other smaller commits that are quite difficult to follow.

@clalancette / @claraberendsen would you consider splitting this into
1. The namedtuple refactor of pip dependencies.
2. The refactor of --break-system-packages to use the environment variable
3. The actual changes to install python3-flake8 from noble-proposed

And anything that doesn't fit in those to separate cleanup PRs. Or at least reasonably explained cleanup commits within those PRs.

This is one repository where history spelunking is often necessary when making changes and I would have a very hard time revisiting this PR and re-constructing the rationale behind all changes from the commit log or PR description as they are right now.

For CI, I think it would be alright to continue using this branch as an omnibus for tests rather than requiring each PR go through CI alone.

ros2_batch_job/__main__.py
clalancette: > In that case does it make sense to restrict the addition of this flag to the platform that requires it?

Yeah, but unfortunately here in this script, we don't have that information.  I'm currently using `humble` as a proxy for it, but that is probably a bad idea overall.  I'll think about how to do this better, we may have to add a new `--platform` flag.

linux_docker_resources/Dockerfile
claraberendsen: It meant to mean packages that are broken/outdated on the repository and it needs to install from proposed instead. 
But I can see why `conflicting` can be confusing, does something like `necessary` or `broken` better?

claraberendsen:
@nuclearsandwich I agree on splitting the flake 8 PR, and link it back to this for testing. I think that overall issues are sistematic and getting the solution for noble given the time constraint we have would be easier to achieve if we have a unified testing branch that imports the changes, so we know when we merge all related PRs then noble should work .

claraberendsen:
The split PR for installing from proposed: https://github.com/ros2/ci/pull/765

clalancette:
FYI, I'm still using this PR as a way to test out various changes that we need for Noble.  At this point I think what is in here actually makes Noble build; I still need to split out the various fixes into PRs.  But anyway, here is a Noble build to see where we are:

* Noble [Build Status](http://ci.ros2.org/job/ci_linux/20634/)

clalancette:
Rebased onto the latest, and more fixes for further changes in Noble, here's another try:

* Noble [Build Status](http://ci.ros2.org/job/ci_linux/20665/)

clalancette:
Over the weekend, it looks like the rest of the packages we needed were rebuilt.  I still believe that there is at least one more fix we need here for Humble, but that will be done in separate PRs.  I'm going to close this one out for now, thanks @claraberendsen !

PR #763, opened 2024-03-25
Unconditionally set PIP_BREAK_SYSTEM_PACKAGES

clalancette:
Older versions of pip will ignore it, and newer versions will automatically apply it.


cottsay (approved):

clalancette:
CI (only Ubuntu 22.04, since it only affects Ubuntu and Ubuntu 24.04 fails for other reasons):

* Linux (Ubuntu 22.04) [Build Status](http://ci.ros2.org/job/ci_linux/20593/)

nuclearsandwich (approved):

PR #762, opened 2024-03-25
Switch to raw string for expect string.

clalancette:
Otherwise, modern Python (3.12, at least), complains:

/tmp/rti_web_binaries_install_script.py:29: SyntaxWarning: invalid escape sequence '\['
  'Press \[Enter\] to continue:',
/tmp/rti_web_binaries_install_script.py:30: SyntaxWarning: invalid escape sequence '\?'
  'Do you accept this license\? \[y/n\]: '])

PR #764, opened 2024-03-25
Redo how we install pip packaging.

clalancette:
The current way we install packages in __main__ is really a mess.  Some of the packages have pinned versions, some do not.  Some of the packages are installed only if they are not installed by the underlying OS, some are not.

This commit partially cleans this up.  In particular, for all pip packages that are *not* colcon-related, we have a tuple which contains the package name, the import name, and the (optional) version we want installed.  For every pip package name, we see if we can import it (using the import name); if we can, then we don't need to install it and we skip it.  If we *do* need to install it, then we add it to the list and add the version to the constraints.

This should mean that this code works on all of Ubuntu, RHEL, and Windows.  On Ubuntu and RHEL, almost all packages are installed via the Dockerfiles (the one exception being colcon). For Windows, almost nothing is installed by the chef scripts, so most things will be installed here.

A further cleanup here would to also install colcon from packages rather than pip, but this is large enough as it is. If this is acceptable, we can do that in a followup.

Along with some other fixes, this change allows us to actually launch an Ubuntu 24.04 job (it still won't pass tests, but that is due to code-related reasons which we are fixing separately).

While this can be merged independently, in order for Ubuntu 24.04 to work we also need to merge in https://github.com/ros2/ci/pull/762 and https://github.com/ros2/ci/pull/763

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -94,6 +94,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-flake8-docstrings \
   python3-flake8-import-order \
   python3-flake8-quotes \
+  python3-importlib-metadata \
   python3-jsonschema \
   python3-lxml \
   python3-mock \
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -68,6 +68,8 @@ pip_dependencies = [
     PipPackage('lark', 'lark', '==1.1.1'),
     PipPackage('mypy', 'mypy', '==0.931'),
     PipPackage('nose', 'nose', ''),
+    PipPackage('pathspec', 'pathspec', ''),
+    PipPackage('pep8', 'pep8', ''),
     PipPackage('pydocstyle', 'pydocstyle', ''),
     PipPackage('pyflakes', 'pyflakes', ''),
     PipPackage('pyparsing', 'pyparsing', '==2.4.7'),
@@ -76,10 +78,12 @@ pip_dependencies = [
     PipPackage('pytest-mock', 'pytest_mock', ''),
     PipPackage('pytest-repeat', 'pytest_repeat', ''),
     PipPackage('pytest-rerunfailures', 'pytest_rerunfailures', ''),
+    PipPackage('pytest-runner', 'ptr', ''),
     PipPackage('pytest-timeout', 'pytest_timeout', '==2.1.0'),
     PipPackage('pyyaml', 'yaml', ''),
     PipPackage('setuptools', 'setuptools', '==59.6.0'),
     PipPackage('vcstool', 'vcstool', ''),
+    PipPackage('yamllint', 'yamllint', ''),
 ]
 
 colcon_packages = [
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -465,8 +465,10 @@ def run(args, build_function, blacklisted_package_names=None):
         for pkgname, importname, version in pip_dependencies:
             if need_package_from_pipy(importname):
                 pip_packages.append(pkgname)
-                if version:
-                    constraints.append(f'{pkgname}{version}')
+            # Even if we don't need to install the package, we still add the constraints
+            # (if they exist) so that other package installations will respect these.
+            if version:
+                constraints.append(f'{pkgname}{version}')
 
         if sys.platform == 'win32':
             # Install fork of pyreadline containing fix for deprecation warnings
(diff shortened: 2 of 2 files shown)


clalancette:
Full CI on this one:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20594/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14952/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21315/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/619/)

linux_docker_resources/Dockerfile
```diff
@@ -94,6 +94,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-flake8-docstrings \
   python3-flake8-import-order \
   python3-flake8-quotes \
+  python3-importlib-metadata \
```
nuclearsandwich, line 97: Can you please confirm the origin of this new dependency? This is added because Debian has split this module from setuptools into its own deb package correct?

ros2_batch_job/__main__.py
```diff
@@ -452,22 +452,21 @@ def need_package_from_pipy(pkg_name):
...
-        # We prefer to get lark from the distribution if it exists.  If not we install it via pip.
-        if need_package_from_pipy("lark"):
-            pip_packages += ["lark==1.1.1"]
+        # We prefer to get packages from the distribution if they are already installed.
+        # If not, we add to the list to install from pip.
+        for pkgname, importname, version in pip_dependencies:
+            if need_package_from_pipy(importname):
+                pip_packages.append(pkgname)
+                if version:
+                    constraints.append(f'{pkgname}{version}')
```
nuclearsandwich, line 469: It's pretty cool to have this set up as completely generic!

nuclearsandwich (approved):
Thanks for splitting this out. I've got an inline comment or two but none are blocking.

linux_docker_resources/Dockerfile
clalancette: Yes, it has been split out.

To be perfectly frank, I'm not 100% sure that this is needed.  However, we currently install this in RHEL, and we also install it via pip (conditionally as of this PR).  So I thought it would be best if we were consistent on how we install it.  In the future we may want to consider removing it completely, but I think we would have to test that out separately.

clalancette:
After the latest work which should fix the outstanding issues, here is another try at CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20612/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14965/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21333/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/621/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21334/)

ros2_batch_job/__main__.py
```diff
@@ -505,12 +506,12 @@ def need_package_from_pipy(pkg_name):
...
+                ['cryptography', 'lxml', 'numpy'], shell=True)
+
+        print('Using constraints:')
+        print('\n'.join(constraints))
         with open('constraints.txt', 'w') as outfp:
-            outfp.write('empy < 4\n')
-            outfp.write('flake8 < 5.0.0\n')
-            outfp.write('setuptools==59.6.0\n')
-            outfp.write('pytest==6.2.5\n')
+            outfp.write('\n'.join(constraints))
```
cottsay, line 514:
```suggestion
            outfp.write('\n'.join(constraints) + '\n')
```

clalancette:
Another try at Windows after the addition of the latest set of packages, which should make this exactly the same as the nightlies now:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21346/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21347/)

cottsay (approved):

clalancette:
One more really quick Linux run, just to ensure that the latest change works:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20621/)

clalancette:
OK.  There are more failures on Windows Debug than I would like, but it has also been broken for about a week so I'm not super surprised.  The remaining failures mostly seem to revolve around timeouts in cpplint/cppcheck, which we can probably increase the timeouts for.  None of it seems connected with the change in Python packages here.  So in order to make progress on switching to Noble, I'm going to merge this one in; we'll work on improving Windows Debug in the next few weeks.

Thank you to everyone for the reviews.

PR #765, opened 2024-03-26
Install flake8 on noble from noble-proposed

claraberendsen:
### Description
This PR configures ubuntu distro proposed repository as a source for installing packages that have not been updated in the upstream distribution.
In particular installs `flake8` from noble-proposed instead of noble.


clalancette:
I've rebased this on the latest now that I merged in #764.  I also simplified it; we should only need to do this action for noble.  Here is a quick CI run on Jammy and Noble to see that it installs what we expect:

* Jammy [Build Status](http://ci.ros2.org/job/ci_linux/20626/)
* Noble [Build Status](http://ci.ros2.org/job/ci_linux/20627/)

clalancette:
All right.  I've done some additional work here to ensure that we *only* install `python3-flake8` from proposed.  This is *still* not enough to make this work on noble; we need the rest of the changes from https://github.com/ros2/ci/pull/761 as well.  But this is one step closer, so I still think we should get this in.

clalancette:
Here is simple CI, only on Ubuntu.  Jammy is expected to succeed, Noble is expected to fail, but it should at least install python3-flake8 7.0.0:

* Jammy [Build Status](http://ci.ros2.org/job/ci_linux/20632/)
* Noble [Build Status](http://ci.ros2.org/job/ci_linux/20633/)

clalancette (approved):

nuclearsandwich (commented):
@claraberendsen is the pin-priority for noble-proposed set up correctly by default according to https://wiki.ubuntu.com/Testing/EnableProposed#Selective_upgrading_from_-proposed

I don't see any configuration changes to apt priority in this PR.

linux_docker_resources/Dockerfile
```diff
@@ -122,6 +122,16 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   uncrustify \
   yamllint
 
+# Configure proposed source, and install python3-flake8 from it
+RUN if test \( ${UBUNTU_DISTRO} = noble \); then\
+    printf "deb http://archive.ubuntu.com/ubuntu/ "$(lsb_release -cs)"-proposed restricted main multiverse universe" >> /etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list && \
+    printf "Package: *\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: -10" >> /etc/apt/preferences.d/proposed-updates && \
+    cat /etc/apt/preferences.d/proposed-updates && \
```
nuclearsandwich, line 129: Is this just for visibility / debug purposes or does it serve some other intent?

linux_docker_resources/Dockerfile
```diff
@@ -122,6 +122,16 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   uncrustify \
   yamllint
 
+# Configure proposed source, and install python3-flake8 from it
+RUN if test \( ${UBUNTU_DISTRO} = noble \); then\
+    printf "deb http://archive.ubuntu.com/ubuntu/ "$(lsb_release -cs)"-proposed restricted main multiverse universe" >> /etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list && \
+    printf "Package: *\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: -10" >> /etc/apt/preferences.d/proposed-updates && \
```
nuclearsandwich, line 128: Are negative values valid here? Most of the docs I have seen have `1` as the lowest value.

Do we need an additional line upping the priority of python3-flake8 to prevent it from being downgraded downstream?
```suggestion
    printf "Package: *\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: -10" >> /etc/apt/preferences.d/proposed-updates && \
        printf "Package: python3-flake8\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: 100" >> /etc/apt/preferences.d/proposed-updates && \
```

nuclearsandwich (approved):
A couple of open questions but however they're answered to the author's satisfaction is sufficient for me.

claraberendsen:
@nuclearsandwich Here is the outcome of the research. It seems that the default value for the priority of proposed was 100, which didn't serve our use case.
I've decided to add a priority negative to the proposed source, that way according to [how apt interprets priority](https://manpages.ubuntu.com/manpages/focal/man5/apt_preferences.5.html) 
```
  P < 0
           prevents the version from being installed
```
would make it so that proposed has to be explicitly asked for in order to install from that source.
Here is a run with the latest changes.

[Build Status](https://ci.ros2.org/job/ci_linux/20641/)

linux_docker_resources/Dockerfile
claraberendsen: Yes they are see comment below. As is now, we are installing python-flake8 first from noble and then updating it to noble-proposed so it shouldn't be downgraded afterwards.

linux_docker_resources/Dockerfile
claraberendsen: Debugging going to remove it. Had some issues with `echo -e` adding the `-e` character because that's a specific attribute from bash.

linux_docker_resources/Dockerfile
claraberendsen: * It shouldn't be downgraded because the version we have is more recent. And the default for noble sources is 500, which won't replace if version is newer.

linux_docker_resources/Dockerfile
```diff
@@ -122,6 +122,15 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
```
nuclearsandwich, line 128:
```suggestion
    printf "deb http://archive.ubuntu.com/ubuntu/ "$(lsb_release -cs)"-proposed restricted main multiverse universe\n" >> /etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list && \
    printf "Package: *\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: -10\n" >> /etc/apt/preferences.d/proposed-updates && \
```

I noticed this when reading job logs. Please add trailing newlines when using printf.

linux_docker_resources/Dockerfile
nuclearsandwich: Thanks. I interpreted

> 
>       P < 0
>           prevents the version from being installed

As preventing the installation even upon request but that's demonstrably not the case bsaed on the job log.

nuclearsandwich (approved):
One other nitpick I noticed while reading the job log.

linux_docker_resources/Dockerfile
```diff
@@ -122,6 +122,15 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   uncrustify \
   yamllint
 
+# Configure proposed source, and install python3-flake8 from it
+RUN if test \( ${UBUNTU_DISTRO} = noble \); then\
+    printf "deb http://archive.ubuntu.com/ubuntu/ "$(lsb_release -cs)"-proposed restricted main multiverse universe\n" >> /etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list && \
+    printf "Package: *\nPin: release a="$(lsb_release -cs)"-proposed\nPin-Priority: -10\n" >> /etc/apt/preferences.d/proposed-updates && \
+    apt-get update &&  \
+    apt-get install --no-install-recommends -y python3-flake8/$(lsb_release -cs)-proposed \
```
clalancette, line 130: I'm still of the opinion that we should remove both the `/etc/apt/sources.list.d/ubuntu-noble-proposed.list` and `/etc/apt/preferences.d/proposed-updates` files once we are done installing.  That way, we won't be pulling metadata updates for the `noble-proposed` for all other `apt` calls below.

That said, I'm not going to block on this.  If you think we don't need to remove them, then we can leave it as-is.

clalancette: I went ahead and made this change.  I'm going to run quick CI on this, then go ahead and merge.

clalancette:
CI:
* Jammy [Build Status](http://ci.ros2.org/job/ci_linux/20644/)
* Noble [Build Status](http://ci.ros2.org/job/ci_linux/20645/)

clalancette:
I'm going ahead and merging this one, so we can make further progress on other changes needed here.

PR #773, opened 2024-04-24
Fix a couple more problems found while testing CI.

clalancette:
1.  Make sure to install the connext debians on both jammy and noble.  This doesn't make a huge difference on the CI jobs (where we'll install Connext via scripts later anyway), but matters for the packaging jobs.
2.  Revamp how we are installing the LTTng-related packages on RHEL.  The logic here is complex enough that it warrants splitting it out to a separate block.  To wit, on Humble we install nothing (ros2_tracing is not installed by default), on Iron we install the dependencies plus python3-lttng (we are using the Python bindings as provided by the platform), and on Jazzy and newer we install the lttng-tools-devel package so we can build our own Python bindings.  This also makes the RHEL logic match the Ubuntu logic.

@nuclearsandwich @marcoag These are the fixes we talked about earlier this morning
@cottsay for review of the RHEL-related parts

### DIFF
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -104,9 +104,6 @@ RUN dnf install \
     libxml2 \
     libyaml-devel \
     libzstd-devel \
-    $(if test ${EL_RELEASE/.*/} != 8; then echo lttng-tools; fi) \
-    $(if test ${EL_RELEASE/.*/} != 8 -a ${ROS_DISTRO} = rolling; then echo lttng-tools-devel; fi) \
-    $(if test ${EL_RELEASE/.*/} != 8; then echo lttng-ust-devel; fi) \
     lz4-devel \
     mesa-libGL-devel \
     mesa-libGLU-devel \
@@ -118,7 +115,6 @@ RUN dnf install \
     $(if test ${EL_RELEASE/.*/} != 8; then echo pybind11-devel; fi) \
     python3-PyYAML \
     python3-argcomplete \
-    $(if test ${EL_RELEASE/.*/} != 8; then echo python3-babeltrace; fi) \
     python3-cairo \
     python3-catkin_pkg \
     python3-cryptography \
@@ -133,7 +129,6 @@ RUN dnf install \
     python3-importlib-metadata \
     $(if test ${EL_RELEASE/.*/} = 8; then echo python3-importlib-resources; fi) \
     python3-lark-parser \
-    $(if test ${EL_RELEASE/.*/} != 8 -a ${ROS_DISTRO} != rolling; then echo python3-lttng; fi) \
     python3-lxml \
     python3-matplotlib \
     $(if test ${EL_RELEASE/.*/} = 8; then echo python3-mock; fi) \
@@ -174,6 +169,15 @@ RUN dnf install \
     yaml-cpp-devel \
     yamllint \
     --refresh -y
+
+# Install dependencies for LTTng
+RUN if test \( ${ROS_DISTRO} = humble \); then \
+      true \
+    ; elif test \( ${ROS_DISTRO} = iron \); then \
+      dnf install lttng-ust-devel lttng-tools python3-babeltrace python3-lttng --refresh -y \
+    ; else \
+      dnf install lttng-ust-devel lttng-tools python3-babeltrace lttng-tools-devel --refresh -y \
+    ; fi
 
 # Install dependencies of Connext and its installer
 RUN dnf install \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -104,6 +104,9 @@ RUN dnf install \
     libxml2 \
     libyaml-devel \
     libzstd-devel \
+    $(if test ${EL_RELEASE/.*/} != 8; then echo lttng-tools; fi) \
+    $(if test ${EL_RELEASE/.*/} != 8; then echo lttng-tools-devel; fi) \
+    $(if test ${EL_RELEASE/.*/} != 8; then echo lttng-ust-devel; fi) \
     lz4-devel \
     mesa-libGL-devel \
     mesa-libGLU-devel \
@@ -115,6 +118,7 @@ RUN dnf install \
     $(if test ${EL_RELEASE/.*/} != 8; then echo pybind11-devel; fi) \
     python3-PyYAML \
     python3-argcomplete \
+    $(if test ${EL_RELEASE/.*/} != 8; then echo python3-babeltrace; fi) \
     python3-cairo \
     python3-catkin_pkg \
     python3-cryptography \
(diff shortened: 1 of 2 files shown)


linux_docker_resources/Dockerfile-RHEL
```diff
@@ -175,6 +170,15 @@ RUN dnf install \
...
 
+# Install dependencies for LTTng
+RUN if test \( ${ROS_DISTRO} = humble \); then \
+      true \
+    ; elif test \( ${ROS_DISTRO} = iron \); then \
+      dnf install lttng-ust-devel lttng-tools python3-babeltrace python3-lttng --refresh -y \
+    ; else \
+      dnf install lttng-ust-devel lttng-tools python3-babeltrace lttng-tools-devel --refresh -y \
+    ; fi
+
```
christophebedard, line 181: So we don't need to check for `${EL_RELEASE/.*/} != 8` anymore?

clalancette: Right, good question.  Only `humble` uses RHEL-8, so the first part of the `if` stanza is a stand-in for what used to be our RHEL-8 checks.  All of Iron, Jazzy, and Rolling use RHEL-9, so they all have the rest of the packages available.

In some sense, I guess the logic here really is:

```
if ($OS == RHEL-8)
   # Do nothing, since the lttng packages are too old *and* humble doesn't support this out-of-the-box
elif ($OS == RHEL-9)
  if ($ROS_DISTRO == iron)
    dnf install lttng-ust-devel lttng-tools python3-babeltrace python3-lttng --refresh -y
  else
    dnf install lttng-ust-devel lttng-tools python3-babeltrace lttng-tools-devel --refresh -y
  fi
fi
```

cottsay: Right now I can trigger a RHEL 8 build of Rolling. This makes that impossible.

I don't really follow why Jazzy warrants these changes. Can't we just change the conditional on line 108 to include Jazzy as well and call it a day?

clalancette: > I'm lukewarm on this. If I understand it correctly, it couples the package logic between the ROS distro and the platform.

Yes and no.  It's complicated.

If we ignore the exact platform for the moment, the situation is this:
1.  Humble doesn't need any lttng packages in CI, because we don't support ros2_tracing out-of-the-box.
2.  Iron needs lttng-ust-devel, lttng-tools, python3-babeltrace, and python3-lttng, because there we use the Python bindings as provided by the platform.
3.  Rolling, Jazzy, and presumably newer needs lttng-ust-devel, lttng-tools, python3-babeltrace, and lttng-tools-devel because we build our own Python bindings.

If we then add in the platform to the mix, the situation is that:
1.  RHEL-8 can't possibly support any of this, because the version of lttng is too old (2.8) and our code doesn't support this old version.
2.  RHEL-9 can support any of the combinations of things that we need for any of the ROS distributions.

I'm not sure where that leaves us in terms of this logic.  Any suggestions on what we might do differently here?

clalancette: > I don't really follow why Jazzy warrants these changes. Can't we just change the conditional on line 108 to include Jazzy as well and call it a day?

We can, but then we'll run into the same problem when we get to K-Turtle.  I was trying to make it so we didn't have to think about this in the future.

> Right now I can trigger a RHEL 8 build of Rolling. This makes that impossible.

True, but I highly doubt that would work for other reasons anyway (python version, etc).

cottsay: >  I was trying to make it so we didn't have to think about this in the future.

Alright, then let's drop the `-a ${UBUNTU_DISTRO} = jammy` entirely. If Iron don't build when `lttng-tools-devel` is installed, we're boxing ourselves out of building multiple ROS distributions on the same host anyway.

clalancette: > Alright, then let's drop the `-a ${UBUNTU_DISTRO} = jammy` entirely. If Iron don't build when `lttng-tools-devel` is installed, we're boxing ourselves out of building multiple ROS distributions on the same host anyway.

All right, fair enough.  I'll change this up (I'll have to do the same thing with python3-lttng, but that should be fine).

cottsay: Actually, let's start a database to hold all of this inter-platform dependency logic. We could even whip up a tool to automatically invoke the system's package manager to install the packages automatically.

/s ...kinda.

clalancette: > Actually, let's start a database to hold all of this inter-platform dependency logic. We could even whip up a tool to automatically invoke the system's package manager to install the packages automatically.

Heh, yeah.  I've actually tried to refactor this code to support `rosdep`, but it would be a big change; we'd have to move all of the logic for dealing with `ros2.repos` into the Dockerfiles so we could then run `rosdep` in there.  And then we'd have to figure out how to do the equivalent in Windows.  It isn't impossible, but it also isn't trivial.

cottsay (approved):

linux_docker_resources/Dockerfile-RHEL
cottsay: FWIW, that's how the CI jobs on build.ros2.org work, and they still have docker caching.

clalancette:
New CI after changes from review:

CI:
Rolling+Noble amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux/20931/)
Rolling+RHEL-9 amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/642/)
Jazzy+Noble amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux/20932/)
Jazzy+RHEL-9 amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/643/)
Iron+Jammy amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux/20933/)
Iron+RHEL-9 amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/644/)
Humble+Jammy amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux/20934/)
Humble+RHEL-8 amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/645/)

clalancette:
And here are packaging jobs:

Rolling+Noble amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/554/)
Rolling+RHEL-9 amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-rhel/55/)
Jazzy+Noble amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/555/)
Jazzy+RHEL-9 amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-rhel/56/)
Iron+Jammy amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/556/)
Iron+RHEL-9 amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-rhel/57/)
Humble+Jammy amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/557/)
Humble+RHEL-8 amd64: [Build Status](http://ci.ros2.org/job/ci_packaging_linux-rhel/58/)

clalancette:
All right.  The Humble+RHEL-8 combination is currently failing, but it is failing on `master` as well: Humble+RHEL-8 amd64 CI: [Build Status](http://ci.ros2.org/job/ci_linux-rhel/654/).  I'll address it separately.

With that, I'm going to go ahead and merge this one in.  Thanks for the reviews, much appreciated.

PR #782, opened 2024-06-11
Install fastjsonschema everywhere.

clalancette:
We are going to need it for an upcoming change to rosidl.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -95,6 +95,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-colcon-zsh \
   python3-coverage \
   python3-cryptography \
+  python3-fastjsonschema \
   python3-flake8 \
   python3-flake8-blind-except \
   python3-flake8-builtins \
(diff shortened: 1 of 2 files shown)


clalancette:
CI for this is over in https://github.com/ros2/rosidl/pull/809#issuecomment-2160775257

wjwwood (approved):
lgtm, I guess this is covered by rosdep install in the manual instructions?

clalancette:
> lgtm, I guess this is covered by rosdep install in the manual instructions?

Yep, exactly.  Though you bring up a good point; for Windows installation instructions, we'll need to add it, since we don't use rosdep there.  See https://github.com/ros2/ros2_documentation/pull/4533

clalancette:
And actually, here is a RHEL-8 canary build, because now I suspect this won't actually work there:

* RHEL-8 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/751/)

clalancette:
OK, yeah.  I'm going to have to do something slightly different here that requires another PR.  Upcoming.

clalancette:
All right, we now need https://github.com/ros-infrastructure/ros2-cookbooks/pull/71 as well, which should fix this for both Windows and RHEL-8.

clalancette:
And here is RHEL-8 again with these changes:

* RHEL-8 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/757/)

clalancette:
Updated with the latest hash from ros2-cookbooks now that https://github.com/ros-infrastructure/ros2-cookbooks/pull/71 was merged.  With that, going ahead and merging.

PR #784, opened 2024-06-18
Select VS release for ci run

claraberendsen:
### Description
This PR restores the functionality of selecting which Visual Studio version you want the CI to run with. 

**Changes**
- Inserts a new key in the json solo that specifies the VS release
- Modifies on docker runtime the value according to CI_VS_VERSION
- Updates options to target 2019 and 2022 releases

### DIFF
--- a/job_templates/snippet/property_parameter-definition_common.xml.em
+++ b/job_templates/snippet/property_parameter-definition_common.xml.em
@@ -124,7 +124,7 @@ choices.remove(cmake_build_type)
           <choices class="java.util.Arrays$ArrayList">
             <a class="string-array">
               <string>2019</string>
-              <string>2017</string>
+              <string>2022</string>
             </a>
           </choices>
         </hudson.model.ChoiceParameterDefinition>
--- a/windows_docker_resources/Dockerfile
+++ b/windows_docker_resources/Dockerfile
@@ -49,7 +49,7 @@ RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64V
 ARG ROS_DISTRO
 
 # Dinamically set VS_VERSION from CI options
-RUN powershell -Command {(Get-Content install_ros2_%ROS_DISTRO%.json) -replace '^(\s*"vs_release"\s*:\s*")[^"]*', '${1}${%CI_VISUAL_STUDIO_VERSION%}' | Set-Content install_ros2_%ROS_DISTRO%.json}
+RUN powershell -Command "(Get-Content install_ros2_%ROS_DISTRO%.json) -replace '^(\s*\"vs_release\"\s*:\s*\")[^\"]*', '${1}${%CI_VISUAL_STUDIO_VERSION%}' | Set-Content install_ros2_%ROS_DISTRO%.json"
 
 COPY install_ros2_${ROS_DISTRO}.json C:\TEMP\
 COPY solo.rb C:\TEMP\
--- a/windows_docker_resources/Dockerfile
+++ b/windows_docker_resources/Dockerfile
@@ -71,4 +71,4 @@ RUN c:\cinc-project\cinc\bin\cinc-solo.bat -c C:\TEMP\solo.rb -Eros2ci -j C:\TEM
 WORKDIR C:\ci
 RUN echo "%VS_VERSION%"
 RUN echo %VS_VERSION%
-CMD '"C:\\Program Files (x86)\\Microsoft Visual Studio\\%VS_VERSION%\\BuildTools\\VC\\Auxiliary\\Build\\vcvarsall.bat" x86_amd64 && python run_ros2_batch.py %CI_ARGS%'
+CMD "C:\\Program Files (x86)\\Microsoft Visual Studio\\%VS_VERSION%\\BuildTools\\VC\\Auxiliary\\Build\\vcvarsall.bat" x86_amd64 && python run_ros2_batch.py "%CI_ARGS%"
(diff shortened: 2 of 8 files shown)


windows_docker_resources/Dockerfile
```diff
@@ -48,6 +48,9 @@ RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64V
 # ROS_DISTRO argument should be set to install dependencies for the target ROS version.
 ARG ROS_DISTRO
 
+# Dinamically set VS_VERSION from CI options
```
clalancette, line 51:
```suggestion
# Dynamically set VS_VERSION from CI options
```

windows_docker_resources/install_ros2_rolling.json
```diff
@@ -2,6 +2,7 @@
   "ros2_windows": {
     "download_sources": "false",
     "vs_version": "buildtools",
+    "vs_release": "2022",
```
nuclearsandwich: If we wanted to merge this change before locking in the VS2022 switch we could merge this leaving the rolling default at 2019.

windows_docker_resources/Dockerfile
```diff
@@ -48,6 +48,9 @@ RUN copy /b C:\TEMP\rticonnextdds-src\rti_connext_dds-6.0.1-pro-target-x64Win64V
...
+RUN sed -i 's/"vs_release": .*"/"vs_release": '\"${CI_VISUAL_STUDIO_VERSION}\"'/g' install_ros2_%ROS_DISTRO%.json
```
clalancette, line 52: Is there any way we can set this without doing a sed?  Maybe by passing a parameter?  This isn't a blocker, but it would be nicer.

job_templates/snippet/property_parameter-definition_common.xml.em
```diff
@@ -124,7 +124,7 @@ choices.remove(cmake_build_type)
           <choices class="java.util.Arrays$ArrayList">
             <a class="string-array">
               <string>2019</string>
-              <string>2017</string>
+              <string>2022</string>
```
nuclearsandwich, line 127: Is this parameter actually used anywhere with the docker/cinc windows agents?

When we had bare metal agents both versions of visual studio would be co-installed and we'd use this parameter to select which version's copy of vcvarsall.bat to invoke.

If this is no longer used then we probably shouldn't present it.

clalancette (changes requested):

nuclearsandwich (commented):
There's an unfortunate regression between this setup and the previous iteration of VS version support. In the past we could switch between VS versions with a build parameter. But unless there's some code I have forgotten about to programatically update the cinc solo json files for each distro based on this parameter I don't think we support that any longer.

If we were running on a *nix system, modifying a file in place before using it would be an easy thing to do. However on Windows we'd likely have to use a powershell or python snippet to make it happen.

Doing so might be worth it though since that way we can flip flop rolling between 2019 and 2022 until 2022 is working.

windows_docker_resources/Dockerfile
claraberendsen: Actually that was a brain fart from me, we cannot use sed on windows :laughing: . But the principle stays, the issue with using a parameter is how do we inject that on the chef run.

clalancette: > Actually that was a brain fart from me, we cannot use sed on windows 😆 . But the principle stays, the issue with using a parameter is how do we inject that on the chef run.

That's true, but we must be doing that somehow.  At least, chef seems to know which ros_distro we are using, so there must be a way to pass that information in.

job_templates/snippet/property_parameter-definition_common.xml.em
claraberendsen: I'm using that parameter to handle the vs invocation in the docker file and being able to override the vs version at runtime. See

https://github.com/ros2/ci/blob/20087f40800a9e5f6f427b4659a2d067f2eec425/windows_docker_resources/Dockerfile#L72

Ideally the goal is to restore the functionality to select VS version to CI.

claraberendsen:
Example run with custom VS installation
**[Build Status](https://ci.ros2.org/job/test_ci_windows/352/)**

windows_docker_resources/Dockerfile
```diff
@@ -66,5 +69,7 @@ RUN echo "@todays_date"
 RUN c:\cinc-project\cinc\bin\cinc-solo.bat -c C:\TEMP\solo.rb -Eros2ci -j C:\TEMP\install_ros2_%ROS_DISTRO%.json
 
 WORKDIR C:\ci
-CMD ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Auxiliary\\Build\\vcvarsall.bat", "x86_amd64", "&&", `
+RUN echo "%VS_VERSION%"
+RUN echo %VS_VERSION%
```
nuclearsandwich, line 73: One of these, probably is the one to keep.

windows_docker_resources/install_ros2_humble.json
```diff
@@ -2,6 +2,7 @@
   "ros2_windows": {
     "download_sources": "false",
     "vs_version": "buildtools",
+    "vs_release": "VS_VERSION",
```
nuclearsandwich, line 5: Alas that we cannot comment in json to remark that this value will be populated during the build. One option could be to make the value the comment:
```suggestion
    "vs_release": "REPLACE_WITH_VS_VERSION",
```

nuclearsandwich (approved):

claraberendsen:
Closing in favor of #786

PR #789, opened 2024-08-20
Don't install Connext on RHEL jobs.

clalancette:
We recently realized that we were installing, building, and testing rmw_connextdds on RHEL.  However, we don't officially support Connext on RHEL, and we don't have RHEL Connext RPMs either.

In the ROS PMC meeting, we agreed that this combination of factors mean we don't need to install or test Connext on RHEL.  This will have two positive outcomes:
1.  The RHEL jobs will finish significantly faster, since they no longer need to download, install, build, and test for Connext/rmw_connextdds.
2.  We'll save some disk space on these jobs, since the Connext binaries are fairly large (about 1GB).

I'll note that while I was in here, I inverted the architecture test since we specifically only want to install Connext on Ubuntu amd64 at present (we also install it on Windows, but that uses a completely separate mechanism).

@claraberendsen FYI.

### DIFF
--- a/linux_docker_resources/entry_point.sh
+++ b/linux_docker_resources/entry_point.sh
@@ -20,8 +20,10 @@ echo "Enabling multicast..."
 ifconfig eth0 multicast
 echo "done."
 
-# We only attempt to install Connext on amd64
-if [ "${ARCH}" != "aarch64" ]; then
+. /etc/os-release
+
+# We only attempt to install Connext on Ubuntu amd64
+if [ "${ARCH}" = "x86_64" -a "${ID}" = "ubuntu" ]; then
     IGNORE_CONNEXTDDS=""
     ignore_rwm_seen="false"
     for arg in ${CI_ARGS} ; do


clalancette:
Here is a full CI run on Ubuntu Linux (which should install the Connext binaries, and thus build/test `rmw_connextdds`) and RHEL (which should not install the Connext binaries, and thus should *not* build/test `rmw_connextdds`):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21683/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15958/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1184/)

ahcorde:
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1200/)

Included https://github.com/ros2/rmw_connextdds/pull/158

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21728/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16003/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22431/)

ahcorde (approved):

cottsay:
We've been shipping binary archives with `rmw_connextdds` for a while now. Should we mention somewhere that we no longer do that and why?

clalancette:
> We've been shipping binary archives with `rmw_connextdds` for a while now. Should we mention somewhere that we no longer do that and why?

Oh, hm.  That's an interesting case.  In our default installation, that never could have worked; while we may have shipped `rmw_connextdds`, we never shipped Connext itself.  I guess it could have worked if someone obtained and installed Connext 6.0.1 themselves, but I'm guessing that's a small crowd.

Further, [REP-2000](https://www.ros.org/reps/rep-2000.html) explicitly says we only support Connext on Ubuntu, Windows, and macOS.

Regardless, you are right.  I'll make an announcement on Discourse that we've done this, because it will affect Humble, Iron, and Jazzy as well.

ros-discourse:

https://discourse.ros.org/t/removal-of-rmw-connextdds-on-rhel/39268/1

PR #794, opened 2024-12-06
Remove colcon-common-extensions from Windows.

clalancette:
It is inadvertently pulling in colcon-notifications, which causes minor problems in CI.  Call out the individual plugins we want, which matches how we do this in Linux.

This depends on https://github.com/ros-infrastructure/ros2-cookbooks/pull/76 .  Once we merge in that PR, I'll go ahead and update the hash here.


clalancette:
Rolling Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22984/)
Humble Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22985/)
Jazzy Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22986/)

cottsay (approved):

PR #795, opened 2024-12-06
Pin the PyQt5.sip dependency.

clalancette:
The latest version released on December 6, 2024 (12.16.0) no longer supports Python 3.8.  Unfortunately that is what we are still using for Windows, so pin this to the versions that are used in the respective Ubuntu versions (this will eventually be solved by switching to Conda, but we aren't there yet).

This depends on https://github.com/ros-infrastructure/ros2-cookbooks/pull/77 .  Once that is approved and merged, I will update the corresponding SHA1 hash here.


clalancette:
CI:
Rolling Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22988/)
Humble Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22989/)
Jazzy Windows CI: [Build Status](http://ci.ros2.org/job/ci_windows/22990/)

cottsay (approved):

PR #803, opened 2025-03-06
Pin the version of pixi.

clalancette:
The main reason to do this is to make sure that we aren't subject to man-in-the-middle attacks, where we are served a malicious version of pixi.  Instead, we have a hard-coded sha256 of the binary, and if it fails to match the build will fail.  This also means that if we want to update the version of pixi, we'll have to manually update the hash.


mjcarroll (approved):

mjcarroll:
CC: @claraberendsen 

It would be nice if we could get a dependabot check on this so we don't forget about incrementing this version.

claraberendsen:
@mjcarroll I don't think there is dependabot support for pixi yet (see https://pixi.sh/v0.40.2/advanced/updates_github_actions/). We could do a workaround as said there, what do you think?

PR #802, opened 2025-02-03
Switch to using Pixi/Conda for Windows.

clalancette:
This allows us to much more easily manage dependencies, and to update them in the future.

We accomplish this by first, removing the chef cookbooks. Next, we rewrite the Windows dockerfile to install MSVC, install Connext, install pixi, fetch the dependency list from https://github.com/ros2/ros2 , install those dependencies, and then run the build.

This depends on https://github.com/ros2/ros2/pull/1642 being merged first, as it will provide the file necessary to make this run.

PR #807, opened 2025-03-22
Restore daily invalidation of Windows Dockerfile.

nuclearsandwich:
This was removed in #802 when removing the chef configuration but we do
this on Linux as well.

It makes sense to do this before configuring the pixi workspace if not
also doing before pixi installation.

In the chef configuration, due to the amount of time spent in
installation we did one chef provision before invalidation and then one
after it in the hope that unchanged things would stay unchanged and the
catchup run would not take as much time each day.

On linux we invalidate after initial configuration but before
dependency installation so we could move this up in the Windows
container if we want.

Also I renamed the sigil string from `todays_date` to `today_str` so
that the sigil we use on Linux and Windows match.

### DIFF
--- a/job_templates/ci_job.xml.em
+++ b/job_templates/ci_job.xml.em
@@ -245,6 +245,8 @@ set DOCKERFILE=windows_docker_resources\Dockerfile
 rem "Finding the Release Version is much easier with powershell than cmd"
 powershell $(Get-ItemProperty -Path 'HKLM:SOFTWARE\Microsoft\Windows NT\CurrentVersion\Update\TargetingInfo\Installed\Server.OS.amd64' -Name Version).Version > release_version.txt
 set /p RELEASE_VERSION=&lt; release_version.txt
+rem "Put current date in Dockerfile to force cache invalidation once per day"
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
[Build Status](https://ci.ros2.org/job/test_ci_windows/647/) 
 [Build Status](https://ci.ros2.org/job/test_packaging_windows/20/)

clalancette:
> On linux we invalidate after initial configuration but before
> dependency installation

This is only half true.  On Ubuntu, we do indeed do the invalidation before we install most of the dependencies.  On RHEL, we do the invalidation after the installation of basically everything; the only thing after this is the setup of the user, colcon, and the entry point.  I think we should probably make all 3 consistent, and I would be for doing it *before* installing dependencies (so we get the latest).  What do you think?

cottsay:
> On RHEL, we do the invalidation after the installation of basically everything; the only thing after this is the setup of the user, colcon, and the entry point.

Eh, this is misleading. The invalidation is immediately followed by an update to all packages. We should never be more than 24hrs out-of-date in our RHEL containers.

cottsay:
@nuclearsandwich, have you deployed this change yet?

PR #808, opened 2025-03-25
Fix coverage reports UI regression on coverage jobs 

claraberendsen:
### Description
Since the migration it seems like the coverage report UI has broken due to a duplication of the cobertura block making the new coverage plugin fail since the default new expected path is `cobertura.xml` not `coverage.xml`. 
This block is removed and the block with the correct path is left in place. 

Job with UI breaks: https://ci.ros2.org/job/ci_linux_coverage/227
Fixed job: https://ci.ros2.org/job/ci_linux_coverage/228


cottsay (approved):
Thank you!

nuclearsandwich (approved):

cottsay (approved):
Update for `sourceDirectories` looks good, thanks!

PR #819, opened 2025-07-18
Pin flake8-import-order==0.18.2 on RHEL 8 to avoid importlib.metadata

christophebedard:
## Description

RHEL 8 (Humble) has Python 3.6, which doesn't have `importlib.metadata` ([>=3.8](https://docs.python.org/3/library/importlib.metadata.html)). We're currently installing `flake8-import-order` from pip for RHEL 8; it was updated in June to use `importlib.metadata` https://github.com/PyCQA/flake8-import-order/pull/203.

Pin to the last version before the `importlib.metadata` change, which is the last version before 0.19.0: https://pypi.org/project/flake8-import-order/#history. See https://github.com/ros2/ros2/pull/1709#issuecomment-3089799668

### DIFF
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -216,7 +216,7 @@ ADD rticonnextdds-license/rti_license.dat /tmp/rti_license.dat
 
 # Install newer versions of some pip packages for RHEL-8
 RUN if test ${EL_RELEASE/.*/} = 8; then \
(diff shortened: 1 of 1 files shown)


christophebedard:
RHEL job using this CI branch and the upcoming Humble patch release repos file (https://github.com/ros2/ros2/pull/1709), and testing `ament_copyright` (root-y package with flake8 test failure related to this):

* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3811/)

christophebedard:
The above job passed, so I think this is good. Merging.

PR #827, opened 2025-11-18
Target RHEL 9.6 until EPEL rebuilds packages

cottsay:
## Description

Several dependencies in EPEL will need to be rebuilt to address SOVERSION bumps in RHEL 9.7. Until that happens, we should specifically target the previous version of RHEL.

Yes, RHEL builds will use 9.6 by default, where they currently target the latest 9.x version.

No

### Additional Information

Canary build, demonstrating that the container now builds successfully:
[Build Status](https://ci.ros2.org/job/test_ci_linux-rhel/42/)

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,7 +121,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 0,
         'ubuntu_distro': 'noble',
-        'el_release': '9',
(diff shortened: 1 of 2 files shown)


christophebedard-apexai (approved):
LGTM since it apparently works based on that test CI job

cottsay:
Deployed.

New jobs will use 9.6 by default. Re-running a previous job will require manually changing the `el_release` version to `9.6`.

PR #834, opened 2026-01-05
Add Resolute as an Ubuntu distro choice

claraberendsen:
## Description

Adds Resolute as an Ubuntu distro choice.

Yes, now the user can choose Resolute from the UBUNTU_DISTRO params of the jobs in ros2ci.

No
### Additional Information

There is also a change in conditionals from `!= noble` to `= jammy`, this is necessary to maintain a similar behavior in resolute and noble which is the starting point from where we are going. 
Further explanation can be found in commit 9c0777be549d5934905751fbab6a204eda633951


claraberendsen:
It won't change the default, it simply adds Resolute as an option to run with. I expect things to break this is just opening the door to start testing with Resolute.

Issue #832, opened 2026-01-05
Windows 11 support


claraberendsen:
- [ ] Do binaries built on Win 11 infrastructure work on Windows 10? [@knmcguire]
    - https://ci.ros2.org/view/packaging/job/packaging_windows-2025/ has Windows 11 binaries.
- [ ] Can we run Windows 10 containers on Windows 11 machines? [@claraberendsen]
- [ ] MSVC 2022 
    - https://ci.ros2.org/job/test_ci_windows/686 works with a lot compiler warnings.
- [ ] Test regressions

knmcguire:
For binaries build on the windows 11 CI, they seem to work fine on my  windows 10 machine going through the standard installation instructions for rolling:

<img width="960" height="600" alt="Image" src="https://github.com/user-attachments/assets/effc1ce1-acc3-4fab-ae2a-4738f9fe22be" />

I do see that the binaries are built 3 months ago, so I'll start a new build and try again, but so far so good!

This is the job that I got the binaries from: https://ci.ros2.org/job/test_packaging_windows-2025/

knmcguire:
I found some information about the MVCS 2022 build!

The biggest bulk of the warnings are from cyclone dds, in particular the enum type mismatch (warning[ C5286 C5287](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c5200-through-c5399)). This has been noted by the zettascale team itself (see the [ifstatement ](https://github.com/eclipse-cyclonedds/cyclonedds/blob/ea9f8f7ba8c17b7585a2d4838ee25cd1cb2648a2/src/core/ddsc/tests/CMakeLists.txt#L158-L161) here), however this change was done may last year, so way after the latest r[elease 0.10.05 that ROS2 rolling uses](https://github.com/ros2/ros2/blob/d47e01d4f480f2da71cdcb7bf6e8145fd5f6dab1/ros2.repos#L42-L45). So either we get the latest development from cyclone dds (master), but that will introduce a whole set of instabilities (and that we forget to set the version for any ros distros). Let's perhaps ask the maintainers when they are planning to have a next release.

knmcguire:
Alright. A list of tests I've run with the mvcs branch last week:

[...]

PR #839, opened 2026-01-18
Add a pixi toml url to the build parameter arguments

knmcguire:
## Description

Currently the location of the pixi.toml file is a bit hardcoded. See

https://github.com/ros2/ci/blob/15f43e066233273b2da6715dac69aeb6f7f864fa/windows_docker_resources/Dockerfile#L53

This PR adds a CI_PIXI_TOML_URL parameter that will add an option to add an custom pixi.toml url (like in a gist). This allows us to test PRs like this one: https://github.com/ros2/ros2/pull/1785 (this adds a version change to pixi.toml), and we would able to debug dependency issues without having to make a branch on the main ros2/CI repository for windows.

This is related to issue #832 

Yes, on ci.ros.org users can now see an new parameter called CI_PIXI_TOML_URL that they can fill in with another url as similar fasion as with CI_ROS2_REPOS_URL. But nothing should happen if they don't fill it in.

Yes, only for one commit: https://github.com/ros2/ci/commit/2ff65348ba5551570ac7e6712d76d85c3f4eec6c

this was done with Claude Code (Opus 4.5)

### Additional Information

I'll leave it in draft as I just want to double check things before merging (or have people look at it at least to see if I'm not breaking the whole CI!)


ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-january-20-2026/52015/1

cottsay:
It looks like the "original" behavior determined the URL dynamically by substituting the ROS distro.

When we trigger a job for a distribution other than Rolling, there's currently a bunch of job parameters that we need to change. Would it be possible to make the default value for this parameter empty, which will produce the existing behavior?

There are certainly other parameters that should behave like this but don't, and it would be great to set a better precedent moving forward.

PR #841, opened 2026-02-08
Add a pixi toml url to the build parameter arguments (ported from #839)

knmcguire:
See the description of https://github.com/ros2/ci/pull/839

Ported to be a branch in ROS/ci so that we could more easily deploy it on a test ci job to test it out according to @claraberendsen

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -108,6 +108,7 @@ def main(argv=None):
         'ci_scripts_default_branch': args.ci_scripts_default_branch,
         'default_repos_url': DEFAULT_REPOS_URL,
         'supplemental_repos_url': '',
+        'pixi_toml_url': '',
         'time_trigger_spec': '',
         'mailer_recipients': '',
         'ignore_rmw_default': {'rmw_fastrtps_dynamic_cpp'},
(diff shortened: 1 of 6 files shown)


knmcguire:
Ah yes, but that is the part for the Linux docker build. For the windows docker `CI_ROS_DISTRO` was part of the `BUILD_ARGS`

https://github.com/ros2/ci/blob/2f86d08865fbe106e8e0fe45cb9fe2b9848a4b14/job_templates/ci_job.xml.em#L250

Which I didn't want to add the pixi_toml_url by standard hence I added it as an conditional [here ](https://github.com/ros2/ci/blob/db24a7c558be42f33ac3008c4e630c2bfca56e01/job_templates/ci_job.xml.em#L264C1-L266C2):

```
if "!CI_PIXI_TOML_URL!" NEQ "" (
  set "BUILD_ARGS=!BUILD_ARGS! --build-arg PIXI_TOML_URL=%CI_PIXI_TOML_URL%"
)
```

I'll try to dig it in it a bit further

knmcguire:
@cottsay or @claraberendsen could you deploy this job again to test_ci_windows? It should work this time

knmcguire:
The force push was necessary to sign all my commits and also apply the same fix the packaging job

Just a question though: Do we want to enable this for the ci_launcher as well?

PR #861, opened 2026-04-10
RHEL: sip6 from rhel9

ahcorde:
Related with https://github.com/ros-visualization/qt_gui_core/pull/293


cottsay:
Does the `package.xml` reference a rosdep rule to install this as well?

ahcorde:
probably not, I'm just trying to fix CI (for now)

PR #862, opened 2026-04-10
Revert qt6 support on RHEL

ahcorde:
## Description

Revert qt6 support on RHEL

Related with this comment https://github.com/ros-visualization/qt_gui_core/pull/293#issuecomment-4226467173

### DIFF
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -165,7 +165,6 @@ RUN dnf install \
     python3-pytest-runner \
     python3-pytest-timeout \
     python3-qt5-devel \
-    $(if test ${EL_RELEASE/.*/} -gt 9; then echo python3-qt6-devel python3-pyqt6-devel python3-pyqt6-sip sip6 PyQt-Builder; fi) \
     python3-rosdistro \
     python3-setuptools \
     $(if test ${EL_RELEASE/.*/} != 10; then echo python3-sip-devel; fi) \
@@ -173,9 +172,6 @@ RUN dnf install \
     qt5-qtbase \
     qt5-qtbase-devel \
     qt5-qtbase-gui \
-    qt6-qtbase \
-    qt6-qtbase-devel \
-    qt6-qtbase-gui \
     redhat-rpm-config \
     spdlog-devel \
     sqlite-devel \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -165,7 +165,7 @@ RUN dnf install \
     python3-pytest-runner \
     python3-pytest-timeout \
     $(if test ${EL_RELEASE/.*/} -le 9; then echo python3-qt5-devel; fi) \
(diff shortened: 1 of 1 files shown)


linux_docker_resources/Dockerfile-RHEL
```diff
@@ -165,17 +165,13 @@ RUN dnf install \
     python3-pytest-runner \
     python3-pytest-timeout \
     python3-qt5-devel \
-    $(if test ${EL_RELEASE/.*/} -gt 9; then echo python3-qt6-devel python3-pyqt6-devel python3-pyqt6-sip sip6 PyQt-Builder; fi) \
```
cottsay, line 168: This is for RHEL 10. Why would we roll this back?

linux_docker_resources/Dockerfile-RHEL
```diff
@@ -165,17 +165,13 @@ RUN dnf install \
...
     python3-rosdistro \
     python3-setuptools \
     $(if test ${EL_RELEASE/.*/} != 10; then echo python3-sip-devel; fi) \
     $(if test ${EL_RELEASE/.*/} != 8; then echo python3-types-pyyaml; fi) \
     qt5-qtbase \
     qt5-qtbase-devel \
     qt5-qtbase-gui \
-    qt6-qtbase \
-    qt6-qtbase-devel \
-    qt6-qtbase-gui \
```
cottsay, line 178: Aren't these needed for non-Python Qt6 packages like Rviz?

cottsay (changes requested):

linux_docker_resources/Dockerfile-RHEL
ahcorde: if qt6 is not present in the system we should be able to compile RViz with qt5.

bindings should be fine with qt5 too.

cottsay: Or can RViz use Qt5 as well?

linux_docker_resources/Dockerfile-RHEL
ahcorde: we can keep this line. but I should probably add `qt6-qtbase qt6-qtbase-devel and qt6-qtbase-gui` for the reasons mentioned below

linux_docker_resources/Dockerfile-RHEL
cottsay: Ah, sounds good then. I think we'll want these for RHEL 10 though:
```
$(if test ${EL_RELEASE/.*/} -gt 9; then echo qt6-qtbase qt6-qtbase-devel qt6-qtbase-gui; fi) \
```

linux_docker_resources/Dockerfile-RHEL
sloretz: I added the RHEL 10 lines back

linux_docker_resources/Dockerfile-RHEL
sloretz: Added these for RHEL 10

sloretz (approved):

PR #866, opened 2026-04-14
Qt6 from rolling and Lyrical

ahcorde:
## Description
Instead of using Ubuntu version, I think we should use here ROS_DISTRO.

Kilted is running on Noble and should run Qt5, but Rolling is running on Noble and should run Qt6. 

@cottsay and @sloretz opinions?


cottsay:
Sure, this makes sense. The ROS distro is driving the requirement, not the platform.

cottsay (approved):

sloretz (approved):

PR #869, opened 2026-04-20
Remove reference to python3-qt6-devel

sloretz:
Looks like the right package is python3-pyqt6-devel, which is already listed

## Description

Fixes # (issue)

no


christophebedard (approved):

christophebedard:
since `python3-qt6-devel` definitely doesn't exist on RHEL >9, can we just skip CI?

sloretz:
Just a quick build of ament_cmake to make sure there are no other issues with installing dependencies on RHEL 10

[Build Status](https://ci.ros2.org/job/ci_linux-rhel/8628/)

christophebedard:
CI passed, merging

PR #868, opened 2026-04-20
Add pytest-asyncio dependency

nadavelkabets:
Required for ros2/rclpy#1620

PR #870, opened 2026-04-20
Change RHEL default to RHEL 10

christophebedard:
## Description

In preparation for Lyrical.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -122,7 +122,7 @@ def main(argv=None):
         'dont_notify_every_unstable_build': 'false',
         'build_timeout_mins': 600,
         'ubuntu_distro': 'noble',
-        'el_release': '9',
+        'el_release': '10',
         'ros_distro': 'rolling',
     }
 


cottsay:
This should really be done in the same commit as Resolute, right?

christophebedard:
I don't really know. We kind of need to bump to RHEL 10 to merge https://github.com/ros2/ci/pull/868. Bumping to Resolute at the same time makes sense, but it feels like such the wrong time to do it, when we're trying to get other things in.

cottsay:
> We kind of need to bump to RHEL 10 to merge https://github.com/ros2/ci/pull/868.

It looks like `python3-pytest-asyncio` is available for RHEL 9 via EPEL. Am I missing something? Why does moving to RHEL 10 block that PR?

christophebedard:
I believe it requires Python >=3.12 (for some language features?). RHEL 9 has 3.9.

cottsay:
> I believe it requires Python >=3.12 (for some language features?). RHEL 9 has 3.9.

ros2/rclpy#1620 requires Python >= 3.12? I'm confused.

christophebedard:
it is using asyncio task groups, which were added in 3.11: https://docs.python.org/3/library/asyncio-task.html#task-groups. And apparently it's also using a Python 3.12 feature.

However, we are currently discussing having a check to disable the feature (or just the tests) if the Python version is too old: [#Client Libraries Working Group > AsyncNode @ 💬](https://openrobotics.zulipchat.com/#narrow/channel/533989-Client-Libraries-Working-Group/topic/AsyncNode/near/588487972).

christophebedard:
OK, as per [#Client Libraries Working Group > AsyncNode @ 💬](https://openrobotics.zulipchat.com/#narrow/channel/533989-Client-Libraries-Working-Group/topic/AsyncNode/near/588493180), Nadav is going to try skipping the tests so that we can stay on RHEL 9 for now. Closing this!

PR #871, opened 2026-04-21
Install python3-pytest-asyncio on RHEL > 9

sloretz:
## Description

RHEL 9 version appears to have a bug that fails unrelated tests https://github.com/pytest-dev/pytest-asyncio/issues/293

no

### Additional Information

In support of: https://github.com/ros2/rclpy/pull/1620


christophebedard:
That seems reasonable, since we exclude the tests because of the old Python version on RHEL 9 anyway

christophebedard (approved):

PR #875, opened 2026-04-29
Add lyrical as option for packaging jobs

claraberendsen:
## Description

Adds Lyrical as option for ROS distro selection. 

Yes, Lyrical will now be able to be selected on the jobs. 

No

### DIFF
--- a/job_templates/snippet/property_parameter-definition_common.xml.em
+++ b/job_templates/snippet/property_parameter-definition_common.xml.em
@@ -88,7 +88,7 @@ Note that this only changes some of the underlying system packages installed; it
             <a class="string-array">
               <string>@ros_distro</string>
 @{
-choices = ['humble', 'jazzy', 'kilted', 'rolling', 'lyrical']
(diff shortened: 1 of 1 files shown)


claraberendsen:
Note that when running the packaging jobs, it will be needed to manually input the correct ros2.repos.

CI is not that smart yet

job_templates/snippet/property_parameter-definition_common.xml.em
```diff
@@ -88,7 +88,7 @@ Note that this only changes some of the underlying system packages installed; it
             <a class="string-array">
               <string>@ros_distro</string>
 @{
-choices = ['humble', 'jazzy', 'kilted', 'rolling']
+choices = ['humble', 'jazzy', 'kilted', 'rolling', 'lyrical']
```
cottsay: I don't know if the order matters here, but we should probably maintain the pattern.
```suggestion
choices = ['humble', 'jazzy', 'kilted', 'lyrical', 'rolling']
```

cottsay (approved):

PR #877, opened 2026-05-21
Enable ROS bootstrap repository in RHEL builds

cottsay:
The Ubuntu CI builds require the ROS public repositories to be enabled. While the RHEL builds don't carry the same requirement, we'll need to leverage the bootstrap repository for some short-term packaging workarounds.

In particular, the bootstrap repository contains a workaround for the `bullet-devel` package, which has regressed in EPEL 10.2. While we wait for upstream (me) to fix the problem there, we can pull the package from our bootstrap repository.


cottsay:
Smoke test build for RHEL 10: [Build Status](https://ci.ros2.org/job/ci_linux-rhel/9024/)

gbiggs (approved):

PR #882, opened 2026-06-26
Don't ignore rmw_connextdds by default in linux-aarch64

fgallegosalido:
## Description

Currently, the Connext RMW is ignored by default in linux-aarch64. This pull request changes that so it is now used by default in the CI.

Now the CI doesn't need to explicitly enable the Connext RMW.

No

### DIFF
--- a/job_templates/ci_launcher_job.xml.em
+++ b/job_templates/ci_launcher_job.xml.em
@@ -3,7 +3,6 @@
   <actions/>
   <description>
     Launches a build of the CI job for each supported platform with the same set of parameters.
-    Use of RTI Connext will always be disabled on aarch64 jobs.
   </description>
   <keepDependencies>false</keepDependencies>
   <properties>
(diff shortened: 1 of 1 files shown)


fgallegosalido:
@cottsay can you launch a build with the default CI parameters, to be sure the Connext RMW is not ignored?

cottsay:
This sort of change has to be deployed to be tested, so I deployed it to the "test" launcher job and triggered it: [Build Status](https://ci.ros2.org/job/test_ci_launcher/27/)

fgallegosalido:
Thank you @cottsay! It seems the Connext RMW is no longer ignored in linux-aarch64. However, we are not using the Connext Debian packages (which are required for linux-aarch64). I can't seem to figure out how to always enable them in linux-aarch64, as I thought the change on https://github.com/ros2/ci/blob/master/create_jenkins_job.py#L144 would suffice. Do you have any idea what I am missing?

fgallegosalido:
@cottsay another question I wanted to ask is what else do we need to do to have ARM as an officially supported architecture in the [REP 2000](https://reps.openrobotics.org/rep-2000/)? I'm not sure if that is still getting updated or if there is any other place where the Connext RMW could be considered to support ARM moving forward.

fgallegosalido:
@cottsay any updates on this? This was still missing that the Debian packages are used by default in linux-aarch64, but I can't figure out what I'm missing for that to happen.

fgallegosalido:
@cottsay I think the trick is to force linux-aarch4 builds to use the Connext Debians by default. Let me know if this change is correct and launch a new CI please.

Issue #881, opened 2026-06-26
CI windows Jazzy/kilted/humble fails due to vs2022


knmcguire:
If someone is starting up a ci job for windows for Jazzy, Kilted and Humble, it will fail as it trying to build with VS2022 but their pixi.toml will only work on VS2019. That is at least my theory.

Currently it seems that we set VS2022 for all the distros in the ci, but should ofcourse set that for VS2019 for the older distros.

See this run: https://ci.ros2.org/job/ci_windows/28449/console

knmcguire:
I started a rebuild of the above job with VS2019 to verify this:

https://ci.ros2.org/job/ci_windows/28497/

knmcguire:
This kind of got lost in my todo list. But I've done a ci run with jazzy and MSVC 2019 here and it fails too:  https://ci.ros2.org/job/ci_windows/28946/

So I'm afraid more things are going on then and a simple switch to MSVC 2019 for older distros won't cut it for ci testing

PR #883, opened 2026-06-26
Prevent testing windows on Windows 10 target ROS distros

claraberendsen:
## Description

Since Windows 10 has gone EOL we should stop testing Windows CI for kilted and older. This change uses an artifact in the [ParametrizedBuildTrigger](https://plugins.jenkins.io/parameterized-trigger/) plugin to prevent the build up based on a file creation. This file is dynamically generated with a bash script as the first build step.

It's possible to use the ConditionalBuildStep plugin here, but this changes the hierarchy of the setup from having downstream projects to a static project for ci_launcher. To have the same hierarchy there is  a secondary plugin we could use [Flexible Plugin](https://plugins.jenkins.io/flexible-publish/), but I'm hesitant of adding another plugin to maintain. 

Truly we should shift at least the ci_launcher over to pipelines and re-architect the way we are tracking/reporting the build numbers to address some of the current limitations (build number changes, dynamic downstream projects,etc). 

For now this is a workaround to prevent wasting resources on platforms we no longer support. 

Yes, there will be no `ci_windows` run on Kilted and older. 

For code analysis, the write up of the code I did it myself.

### DIFF
--- a/job_templates/ci_launcher_job.xml.em
+++ b/job_templates/ci_launcher_job.xml.em
@@ -1,4 +1,4 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version='1.1' encoding='UTF-8'?>
 <project>
   <actions/>
   <description>
@@ -51,6 +51,20 @@
   <triggers/>
   <concurrentBuild>false</concurrentBuild>
   <builders>
+  <hudson.tasks.Shell>
+    <command>
+      #!/bin/bash
+      REGEX="^(jazzy|humble|kilted)$"
+
+      if [[ "$CI_ROS_DISTRO" =~ $REGEX ]]; then
+        echo "$CI_ROS_DISTRO targets an EOL Windows version. Skipping Windows CI"
+        rm -f trigger_win_build.properties 
+      else
+        echo "Trigger Windows build for $CI_ROS_DISTRO" >> trigger_win_build.properties 
+      fi
+    </command>
+    <configuredLocalRules/>
+  </hudson.tasks.Shell>
     <hudson.plugins.groovy.SystemGroovy plugin="groovy@@457.v99900cb_85593">
       <source class="hudson.plugins.groovy.StringSystemScriptSource">
         <script plugin="script-security@@1369.v9b_98a_4e95b_2d">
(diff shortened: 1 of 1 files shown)


knmcguire:
This will close https://github.com/ros2/ci/issues/881. Currently it fails completely on windows 11 even with MSVC 2019 selected

knmcguire (approved):
Looks good to me!

But perhaps we want to test it out on the test job first?

mjcarroll (approved):

skyegalaxy (approved):

claraberendsen:
@knmcguire tested this on `test_ci_launcher` and put in a fix. Now it's working as expected. 
- Skips windows if EOL distro: https://ci.ros2.org/job/test_ci_launcher/31/console
- Builds windows lyrical onwards: https://ci.ros2.org/job/test_ci_launcher/32/console

claraberendsen:
After some coffee and taking a look at it again I modified what was being printed to console to match was is being built. This is important since `ros-ci-for-pr` script parses the console log to generate the comment. 

Here is the latest test runs: 

- On rolling -> builds everything: https://ci.ros2.org/job/test_ci_launcher/42/console
- On Jazzy -> skips windows: https://ci.ros2.org/job/test_ci_launcher/43/console

knmcguire (approved):
Looks good to me!

The only thing I'm wondering is how the PR build farm launch script will handle this now, but perhaps something for the ros pmc to comment on

skyegalaxy:
hmm, what changes are needed for the `ros-ci-for-pr` script? that is my go-to for running CI jobs now

knmcguire:
Ah yes that was the script I was trying to look for: https://github.com/ros-tooling/ros-github-scripts/blob/main/ros_github_scripts/ci_for_pr.py

It seems like it's platform agnostic to me? So it just grabs whatever the launcher starts (but I'm also look at it just before bed time so perhaps maybe I'm missing something). At least I don't see any mention of windows other then just the commented out example

skyegalaxy:
yeah I also took a brief pass and I think the net effect is that for older distros, `ros-ci-for-pr` just won't post a windows job in the comment anymore. seems like what we want

skyegalaxy (approved):

claraberendsen:
Yeah, from looking at the code `ros-ci-for-pr` parses the log from ci_launcher so the last fix I put in would mean it no longer shows windows job there.

PR #885, opened 2026-07-09
Do not install python3-pytest-asyncio on humble

Ryan4253:
## Description
As title. Fixes #884 

### Is this user-facing behavior change?
Yes, CI stops breaking in rclpy

### Did you use Generative AI?
No

### Additional Information
Docker-RHEL is not modified since `python3-pytest-asyncio` is already gated by RHEL>9 and Humble uses RHEL8

PR #886, opened 2026-07-13
Fix Humble CI failures caused by pytest-asyncio 0.18.1

### LINKED ISSUE #884: `python3-pytest-asyncio` breaks all rclpy pytest tests on Humble CI
### Generated by Generative AI

Investigated with Claude Fable

### Operating System:

`ci_linux`, `ci_linux-aarch64`

### ROS version or commit hash:

humble

### RMW implementation (if applicable):

_No response_

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

rclpy

### 'ros2 doctor --report' output

_No response_

### Steps to reproduce issue

1. `docker run -it ubuntu:22.04`
2. `apt-get update && apt-get install -y python3-pytest python3-pytest-asyncio`  (installs pytest-asyncio 0.18.1)
3. `echo "def test_ok(): pass" > /tmp/test_min.py`
4. `python3 -m pytest /tmp/test_min.py -We` → INTERNALERROR before any test is collected

### Expected behavior

Humble rclpy pytest tests should run and pass as they don't use `pytest-asyncio`

### Actual behavior

Every pytest invocation in rclpy on Humble Linux CI dies at configure time. e.g. see https://ci.ros2.org/job/ci_linux/29507/
```
INTERNALERROR> DeprecationWarning: The 'asyncio_mode' default value will change to 'strict' in future,
please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.
```

### Additional information

https://github.com/ros2/ci/pull/868 adds `pytest-asyncio` unconditionally to the CI image. On Ubuntu 22.04 this installs `pytest-asyncio 0.18.1`. 0.18.1 defaults to legacy mode which emits a deprecation warning. This fails under `-We`, causing the CI in humble to fail. This can  be observed in https://github.com/ros2/rclpy/pull/1671 and https://github.com/ros2/rclpy/pull/1653

The distros beyond 22.04 ships versions >0.18 which [removed legacy mode](https://github.com/pytest-dev/pytest-asyncio/pull/414).  The default became strict mode which emits no deprecation warning so no CI failures. RHEL is unaffected because its not installed unless >RHEL9 and humble runs RHEL8 for CI. 

Given that we don't use `pytest-asyncio` in Humble and I see no PR backporting `AsyncNode` we should get rid of it in the Docker image for humble

FarbodDaneshjoo:
## 📝 Description

The Ubuntu 22.04 package version of `python3-pytest-asyncio` (0.18.1) causes pytest initialization failures in Humble CI because warnings are treated as errors.

Since Humble rclpy tests do not require `pytest-asyncio`, this dependency is now excluded from Humble images while remaining available for other distributions.

## 🧪 Testing

Not run (Docker dependency-only change).

Fixes #884

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -53,6 +53,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y build-essential
 RUN apt-get update && apt-get install --no-install-recommends -y python3-lark python3-opencv
 
 # Install build and test dependencies of ROS 2 packages.
+# pytest-asyncio 0.18.1 on Ubuntu 22.04 is incompatible with Humble CI (-Werror).
+# Keep it disabled for Humble to avoid breaking pytest initialization.
 RUN apt-get update && apt-get install --no-install-recommends -y \
   $(if test ${ROS_DISTRO} != humble -a ${ROS_DISTRO} != jazzy; then echo cargo; fi) \
   clang-format \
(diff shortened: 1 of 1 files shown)


cottsay:
Duplicate of #885

PR #892, opened 2026-08-27
Add python3-colorama dependency

penww:
## Description

Fix CI issue for https://github.com/ros2/ros2cli/pull/1248. In that PR, we added color support for ros2cli, `python3-colorama` is used for cross-platform colored terminal output.

No.

### Additional Information

https://pypi.org/project/colorama/

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -98,6 +98,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   python3-colcon-ros-domain-id-coordinator \
   python3-colcon-test-result \
   python3-colcon-zsh \
+  python3-colorama \
   python3-coverage \
   python3-cryptography \
   python3-fastjsonschema \
--- a/linux_docker_resources/Dockerfile-RHEL
+++ b/linux_docker_resources/Dockerfile-RHEL
@@ -126,6 +126,7 @@ RUN dnf install \
     $(if test ${EL_RELEASE/.*/} != 8; then echo python3-babeltrace; fi) \
     python3-cairo \
     python3-catkin_pkg \
+    python3-colorama \
     python3-cryptography \
     python3-devel \
     python3-empy \


knmcguire (commented):
Hi! I don't think it is a good idea to add dependencies of actual ros packages here, and these should only be dependencies to support CI as a whole I believe.

Have you tried adding python3-colorama to rosdep? 


(let me double check this internally as well, as I see in the console that rosdep is not even called indeed)

clalancette:
> Have you tried adding python3-colorama to rosdep?

That's already done: https://github.com/ros/rosdistro/blob/2667056d0320bb77edf1e28fec13d56aa8d325e2/rosdep/python.yaml#L3476

But unfortunately, CI here does not use rosdep at all.  I tried to introduce it a few years ago, but it would basically require completely rewriting how the CI scripts work.  So I abandoned it then.  I think this is the best we can do for now.

Note that this will also require adding in `python3-colorama` to https://github.com/ros2/ros2/blob/rolling/pixi.toml for Windows.

knmcguire:
ahh yes thanks @clalancette for clarifying. I looked at the base.yaml file, that's why I missed it. 

And indeed, pixi.toml addition needs to be done as well ([colorama](https://anaconda.org/channels/conda-forge/packages/colorama/files) is on conda-forge). We should be able to use version 0.4.6 for this to match [Ubuntu 26.04](https://repology.org/project/python:colorama/versions)

penww:
@clalancette @knmcguire Thanks for your review and feedback, have added PR for Windows in pixi.toml: https://github.com/ros2/ros2/pull/1857. 

Please help to review again.

knmcguire:
I had to make a port of your branch for here for me to test it out in CI.

Linux: [Build Status](https://ci.ros2.org/job/ci_linux/30351/)
Linux-rhel: [Build Status](https://ci.ros2.org/job/ci_linux-rhel/10151/)
Linux-Aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/23039/)

Windows CI run with pixi is running in the other PR

knmcguire:
Rebuildling aarch because it failed.

[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/23044/)

Btw... I see the build status here embedded as passed though but when you click on it it says failed... Anyone having the same issue?

Crola1702:
@knmcguire if I paste the protected build status are you able to see the result? 

[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/23039/)

knmcguire:
@Crola1702 this is what I see

image

knmcguire:
I see if now after the plugin update

Screenshot_20260829-191200

knmcguire (approved):
Alright, the issue with the embedded status aside, the ci jobs do pass. The aarch one does have some test failures but the same ones happen on the nightlies as well so your addition did not start a regression there.

I'll approve and merge this!

PR #895, opened 2026-09-07
Add clang-tidy to (non-RHEL) Linux

christophebedard:
## Description

Needed to add simple tests for `ament_clang_tidy`, see https://github.com/ament/ament_lint/pull/563.

The RHEL Dockerfile already includes `clang-tools-extra`, which seems to include `clang-tidy`, see the ci_linux-rhel job here: https://github.com/ament/ament_lint/pull/563#issuecomment-5561579764.

No


christophebedard:
CI is at https://github.com/ament/ament_lint/pull/563#issuecomment-5563496856

cottsay (approved):
This seems fine to me. Make sure you update the `package.xml` of packages which use `clang-tidy` so that the test dependency can be installed by rosdep.

christophebedard:
Thanks! It's strictly for `ament_clang_tidy`, which already declares a dependency on `clang-tidy`. It just never had tests until https://github.com/ament/ament_lint/pull/563

PR #900, opened 2026-09-09
Build Windows CI with Ninja

mjcarroll:
## Description

Switch Windows CI from the Visual Studio generator to Ninja. CMake ignores `CMAKE_<LANG>_COMPILER_LAUNCHER` under Visual Studio, so this is a prerequisite for the compiler cache in #899.

- `CMAKE_BUILD_TYPE` defaults to `Release` when unset, matching the previous `--config Release`. Packaging stays on Visual Studio.
- Ninja's object paths are longer and overrun `MAX_PATH` on some rosidl sources, so the build space is `b` and the workspace is `subst`ed onto `W:`.
- `ninja` comes from the `buildfarm` pixi environment, so the image installs and runs `-e buildfarm`.

Depends on ros2/ros2#1854.

No

Yes, Claude Code (Claude Opus 5).

### Additional Information

An earlier revision passed `test_ci_windows` together with #899; the current head needs a rerun.


mjcarroll:
Unspecified Type (Release): [Build Status](https://ci.ros2.org/job/test_ci_windows/760/)

clalancette:
New CI (looks like it took 5.5 hours): [Build Status](https://ci.ros2.org/job/test_ci_windows/765/)

ros2_batch_job/__main__.py
```diff
@@ -211,6 +211,9 @@ def __call__(self, parser, namespace, values, option_string=None):
             raise Exception('Argument {} and "--white-space-in" cannot both be used'.format(name))
         elif space_directory is None:
             space_directory = colcon_space_defaults[name]
+            if name == 'buildspace' and os.name == 'nt':
```
clalancette: A few things about this:

1.  We are keeping the space_directory short here, but if this is used together with `white_space_in`, will it get too long again?
2.   I'm wondering if we should stop making this a special snowflake, and just change `colcon_space_defaults` to use `b` instead?  It is not strictly required for Linux, but I don't think it will actually hurt anything there either.

windows_docker_resources/Dockerfile
```diff
@@ -55,8 +55,8 @@ WORKDIR C:\pixi_ws
 RUN powershell -Command "if (-not $env:PIXI_TOML_URL.StartsWith('https://')) { Write-Error 'PIXI_TOML_URL must use HTTPS protocol'; exit 1 }"
 ADD ${PIXI_TOML_URL} pixi.toml
 RUN powershell -Command "(Get-Item pixi.toml).LastWriteTime = Get-Date"
-RUN pixi --color never --no-progress -q install
-RUN pixi --color never --no-progress -q list
+RUN pixi --color never --no-progress -q install -e buildfarm
+RUN pixi --color never --no-progress -q list -e buildfarm
```
clalancette: I'm guessing we need https://github.com/ros2/ros2/pull/1854 for this to work, right?

mjcarroll: Yes, we need them at the same time.

mjcarroll: Well, we need that first, actually.

ros2_batch_job/__main__.py
mjcarroll: I think we can actually drop it.  If we did it for Linux, I believe we would also have to edit the regexes that Jenkins uses to read test results.

I think we have enough path margin now to do away with this for the moment.

mjcarroll:
One more without the buildspace change: [Build Status](https://ci.ros2.org/job/test_ci_windows/766/)

PR #901, opened 2026-09-14
Order the Windows Dockerfile by how often each layer changes

mjcarroll:
Reorders the Windows Dockerfile so the expensive layers stop getting rebuilt by inputs below them in the cache chain. Instructions are unchanged apart from position, plus dropping a redundant `mkdir`.

- **`ARG ROS_DISTRO` / `ARG PIXI_TOML_URL` moved down to first use.** Every `RUN` after an `ARG` has its value in its cache key, so any build with a different `CI_PIXI_TOML_URL` or distro reinstalled VS Build Tools.
- **Connext moved above the daily bust and the `pixi.toml` `ADD`.** It depends only on `ROS_DISTRO` and the copied submodules, but was rerunning daily and on every `pixi.toml` change.
- **Daily `@today_str` bust sits directly above pixi**, which it exists for: loose pins mean the solve can change without the file changing.

Resulting order: MSVC → pixi binary → Connext (per distro) → daily bust → pixi environment.


knmcguire:
In terms of the content I agree but still running a windows CI here as sanity check: [Build Status](https://ci.ros2.org/job/ci_windows/29445/)

windows_docker_resources/Dockerfile
```diff
@@ -47,28 +43,30 @@ RUN powershell -noexit Expand-Archive -Path pixi-x86_64-pc-windows-msvc.zip -Des
...
-RUN powershell -Command "(Get-Item pixi.toml).LastWriteTime = Get-Date"
-RUN pixi --color never --no-progress -q install
-RUN pixi --color never --no-progress -q list
+ARG ROS_DISTRO
 
 # Setup environment variables needed for Connext
 ENV ROS_DISTRO=${ROS_DISTRO}
 ENV RTI_LICENSE_FILE C:\connext\rti_license.dat
 
-RUN IF NOT EXIST "C:\TEMP" mkdir C:\TEMP
```
clalancette, line 65: Are you absolutely sure we don't need this?  I don't have evidence one way or the other, but I'm just checking.

clalancette (approved):
The idea is fine with me, I have one comment just double-checking that `C:\TEMP` is always guaranteed to exist.

